PR #54, opened 2010-11-01
reduceRight() doesn't have an initial value either

mcmire:
Similar to reduce(), except if you don't pass an initial value for reduceRight(), it's supposed to use the _last_ value in the collection.


test/collections.js
RAIHERE: ###

Issue #586, opened 2012-05-04
Add _.deepClone() to the object methods


evanplaice:
I have a very simple recursive implementation of deep clone for objects and arrays.

I wrote this a while back. It could use some cleanup and optimization but it should work.

if(!Object.prototype.clone) {
Object.prototype.clone = function() {
var newObj = (this instanceof Array) ? [] : {};
for(i in this) {
if(i == 'clone') continue;
if(this[i] && typeof this[i] == 'object') {
newObj[i] = this[i].clone();
} else { 
newObj[i] = this[i];
}
}
return newObj;
};
}

PR #595, opened 2012-05-07
Deep copying with _.clone(obj, deep)

cederberg:
I sometimes return copies of nested internal data structures, so deep copying is needed to avoid accidental changes by other parts of the program or by loaded plug-ins. See also issue #586 and elsewhere.

Since the _.clone() function didn't support deep copies, I thought it would be fairly straight-forward to add it without breaking the API. Not sure that the implementation is the best though.

Also fixed a few corner cases with Date, Function and RegExp objects being cloned. Not very common scenarios perhaps, but easy enough to handle.
### clone_.clone(object, [deep])

Create a copy of the object. The copy is shallow, unless deep is true. This means that nested objects or arrays are copied by reference, not duplicated. Note that the prototype chain is not preserved by this function, unlike Object.create.

``` javascript
_.clone({name : 'moe'});
=> {name : 'moe'};

_.clone([1, 2, 3]);
=> [1, 2, 3];
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -1057,14 +1057,20 @@ _.defaults(iceCream, {flavor : "vanilla", sprinkles : "lots"});
 </pre>
 
       <p id="clone">
-        <b class="header">clone</b><code>_.clone(object)</code>
+        <b class="header">clone</b><code>_.clone(object, [deep])</code>
         <br />
-        Create a shallow-copied clone of the <b>object</b>. Any nested objects
-        or arrays will be copied by reference, not duplicated.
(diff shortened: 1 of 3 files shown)


gotofritz:
The JSON.parse(JSON.stringify trick doesn't work for everything. Example:

    var homer = { name: "Homer Simpsons", sayCatchprase: () => console.log('doh') };
    var bart = JSON.parse(JSON.stringify(homer));
    homer.sayCatchprase(); 
    // doh
    bart.sayCatchprase(); 
    // Uncaught TypeError: bart.sayCatchprase is not a function

PR #1428, opened 2014-01-31
Restrict _.union and _.difference to accept arrays only

mahdi-shojaee:
According to PR #1317, `_.union` and `_.difference` have been restricted to accept arrays only.
This accomplished by adding a new parameter named `strict` to the internal `flatten` function.

So we have:

``` javascript
_.union([1, 2], [3]) // => [1, 2, 3]
_.union([1, 2], 3) // => [1, 2]

_.difference([1, 2], [1]) // => [2]
_.difference([1, 2], 1) // => [1, 2]
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
   <meta name="viewport" content="target-densitydpi=device-dpi" />
   <meta name="HandheldFriendly" content="true"/>
-  <link rel="canonical" href="http://underscorejs.org" />
+  <link rel="canonical" href="http://underscorejs.org/" />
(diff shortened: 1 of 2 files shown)


philfreo:
Today I was looking for a function like _.difference that actually took 2 objects and returned only the _differences_ between those objects, preferring one. Kind of like the opposite of _.defaults.

e.g.:

```
_.objDifference({ x: 1, y: 2, z: 3}, { x: 1, y: 3, z: 4 })
=> { y: 2, z: 3 }
```

Since I didn't see anything to do this, I did this (because all the attributes I cared about were in `obj1`):

```
_.pick(obj1, _.filter(_.keys(obj1), function(key) { return !_.isEqual(obj1[key], obj2[key]); }))
```

Is there an easier way to do this already? If not, do you think something like this makes sense in underscore?

PR #1542, opened 2014-03-26
New feature : reset the debounce

guiled:
It seems a little crappy to add a function into a function, but it provides a feature that was asked in #952 without breaking backward compatibility.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -745,7 +745,7 @@
       }
     };
 
-    return function() {
+    var func = function() {
       context = this;
       args = arguments;
       timestamp = _.now();
@@ -760,6 +760,12 @@
 
       return result;
     };
+    func.reset = function () {
+      clearTimeout(timeout);
+      timeout = null;
+    };
+    
+    return func;
   };
 
   // Returns a function that will be executed at most one time, no matter how


nilbus:
I don't think this is a crappy approach. :+1:

nilbus:
This missing feature is easy to work around though.

```javascript
var actionToDebounce = function() { ... }
var cancelWrapper = function() {
  if (!cancelCondition) actionToDebounce();
}
_.debounce(cancelWrapper, wait);
```

akre54:
> It seems a little crappy to add a function into a function,

Not at all. It's a very common pattern in JS. Underscore tends not to do it that often, but there are certain legitimate use cases (like this one and [`_.partial.placeholder`](https://github.com/jashkenas/underscore/issues/1688)), where I think we should explore the possibility.

underscore.js
```diff
@@ -760,6 +760,12 @@
 
       return result;
     };
+    func.reset = function () {
+      clearTimeout(timeout);
+      timeout = null;
+    };
```
jridgewell: We should release `context` and `args` as well: `context = args = null`.

akre54:
Any updates on this? If someone wants to add in @jridgewell's change and some tests, I'll merge it in.

underscore.js
jdalton: Lodash calls this method `clear`.

akre54: Fine by me :+1:

jridgewell:
Closing in favor of #2337 (your commit was included).

underscore.js
jdalton: Oops I typoed. ~~`clear`~~ `cancel`.

PR #1545, opened 2014-03-28
Make _.each break when false is returned

melnikov-s:
Now that `_.each` no longer uses the native `forEach` it can break when `false` is returned.

### DIFF
--- a/test/collections.js
+++ b/test/collections.js
@@ -41,6 +41,17 @@
     answers = 0;
     _.each(b, function(){ ++answers; });
     equal(answers, 100, 'enumerates [0, length)');
+
+    var lastVal;
+    var c = [1,2,3,4,5];
+    _.each(c, function(num,i) {
+        lastVal = num;
+        if (i === 2) {
+            return false;
+        }
+    });
+
+    equal(lastVal, 3, 'breaks when false is returned');
   });
 
   test('map', function() {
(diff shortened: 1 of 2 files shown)


Nasicus:
Is the status still the same in here?

As far as I could see in 4fb42e44fb4  the "breaker" object - and therefore the only way to break the `each` loop has been removed. I guess this means that `each` will "never" support break? I find this kind of sad, since for me it's a quite important feature of a loop.

jridgewell:
> As far as I could see in 4fb42e4 the "breaker" object - and therefore the only way to break the each loop has been removed.

Outside devs were never able to access that object, only internal underscore functions.

> Is the status still the same in here?

Yup.

Nasicus:
> Outside devs were never able to access that object, only internal underscore functions.

Ah didn't realize that.

However what speaks against to create (as already mentioned above), two sentinel constants like `_.LOOP_BREAK` and `_.LOOP_CONTINUE` (this variable is not really needed, but it may result in better readable/looking code) ?

One could then write his `_each` like this:

	_.each([0,1,2,3], function(i){
		if (i === 0){
			return _.LOOP_CONTINUE; //or just use 'return;'
		}

		if (i === 2){
			return _.LOOP_BREAK;
		}	

		console.log(i);
	});

Sure it's more "complicated" compared to the `return false`of jQuery however it's also much more clearer if you read the code in my opinion.

jgonggrijp:
I'm opposed to this. Maybe a special sentinel value, but not a common everyday value that lots of iteratees are likely to return by coincidence. Then still, `break` is a procedural thing. `find` is the real functional equivalent.

PR #1585, opened 2014-04-16
Deep extend

MikeBazhenov:
Hello!
This commit added new function, for extending big object as:

``` js

var source = {
   auto: 1,
   moto: 1,
   boat: {
      fly: 0,
      cry: 1,
      subkey: {
      //... and so on
      }
   }
}

var object = {
   auto: 0,
   moto: 1,
   boat: {
      fly: 1,
      cry: 0,
      dry: {
          subkey: {
             //... and so on
          }
      },
      subkey: {
      //... and so on
      }
   }
}

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 raw
 node_modules
+.idea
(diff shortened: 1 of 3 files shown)


pedramphp:
Hi Guys,

there is bug with this code, it overwrites the initial settings

http://jsbin.com/weqoce/1/

PR #1650, opened 2014-05-24
Saperated CSS file from HTML

sandeshdamkondwar:
Going to style all the documentation.
Keeping the CSS rules away from html is good idea now, because CSS rules gonna grow.


akre54:
Let's leave this as-is, makes it a bit easier to see everything when it's in one file. If the CSS grows beyond the current scope we can revisit.

PR #1672, opened 2014-06-10
Update description of _.isEmpty.

scottaj:
The HTML documentation did not make it clear how `_.isEmpty` works for strings and arrays vs. objects. The distinction is clear in the annotated source, so I tried to make it reflect that.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1437,8 +1437,9 @@ _.isEqual(moe, clone);
       <p id="isEmpty">
         <b class="header">isEmpty</b><code>_.isEmpty(object)</code>
         <br />
-        Returns <i>true</i> if an enumerable <b>object</b> contains no values
-        (no enumerable own-properties).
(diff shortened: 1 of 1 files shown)


vbarzana:
Guys, this is a concept problem and will bring thousand of developers semantic problems, _.isEmpty(234) simply can't return true, because that is a number and it's not empty, what are you discussing here. I as a developer expect:
_.isEmpty(function(){})    to return false
_.isEmpty(0)    to return false
_.isEmpty(222)    to return false
_.isEmpty({a: 'x'})  to return false
...
and 
_.isEmpty()    to return true
_.isEmpty(undefined)    to return true
_.isEmpty("")    to return true
_.isEmpty(null)    to return true
_.isEmpty({})  to return true
_.isEmpty([])  to return true
...
Can someone please explain me what is the real functionality of skipping these objects? and creating such confusion among the community?
Thanks,
Victor

PR #1711, opened 2014-06-28
Test shimmed bind + normal bind

megawac:
Just some coverage of the shimmed bind


jgonggrijp:
Only meaningful if #2845 gets merged.

PR #1751, opened 2014-07-16
sortBy comparator handles arrays

megawac:
Fixes #1779, #1880

Another try at getting `sortBy` support for multiple criteria since #1359 was rejected. I'd argue this approach is more robust as its implementation is based on the return of the iterator instead of only handling multi prop lookups. Thus we can implement the same behaviour as #1359 with a `properties` mixin if desired later.

``` javascript
_.mixin({
   properties: function(props) {
     return function(obj) {
         return _.map(props, function(prop) {
             return obj[prop];
         });
     }
  }
});

_.sortBy([{a: 1, b: 2}, {a: 1, b: 3}, {a: 1}, {a: 2, b: -1}, {b: 5}], _.properties(["a", "b"]));
```

---

Post implementation enables:

``` javascript
_.sortBy([{a: 1, b: 2}, {a: 1, b: -1}, {a: 1, b: 3}, {a: 2, b: -5}], function(x) {
    return [x.a, x.b];
});
// => [{"a": 1,"b": -1},{"a": 1,"b": 2},{"a": 1,"b": 3},{"a": 2,"b": -5}]
```

---

Exmple Usage (naive semver sort):

``` javascript
var versions = ["0.1", "0.1.1", "2.0", "10.1", "6.2", "2.0.1", "1.7.3", "1.10.2", "0.2.0", "0.3.0", "0.6.0", "1.0.0"];
_.sortBy(versions, function(version) {
   return _.map(version.split("."), _.partial(parseInt, _, 10));
});
// => ["0.1", "0.1.1", "0.2.0", "0.3.0", "0.6.0", "1.0.0", "1.7.3", "1.10.2", "2.0", "2.0.1", "6.2", "10.1"]
```

---

The thing that tripped me up implementing this is that the comparitor ranks `undefined > anything`. I'd argue that `undefined < anything` but that's just me.

I'd really like to see this behaviour in v2.

### DIFF
--- a/test/collections.js
+++ b/test/collections.js
@@ -496,6 +496,33 @@
 
     list = ['q', 'w', 'e', 'r', 't', 'y'];
     deepEqual(_.sortBy(list), ['e', 'q', 'r', 't', 'w', 'y'], 'uses _.identity if iterator is not specified');
+
+    var smallCollection = _.shuffle(_.map(collection.slice(2, 7), function(pair) {
+      return {x: pair.x, y: pair.y};
+    }));
+    sorted = [{x: 1, y: 3}, {x: 1, y: 4}, {x: 1, y: 5}, {x: 1, y: 6}, {x: 2, y: 1}];
+
+    deepEqual(_.sortBy(smallCollection, function(pair) {
+      return [pair.x, pair.y];
+    }), sorted, 'Returning an array will sort by precedence');
+
+    deepEqual(_.sortBy(smallCollection, function(pair) {
+      return [pair.y, pair.x];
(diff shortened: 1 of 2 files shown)


jgonggrijp:
`_.sortBy` is for when you can reduce your values to a single string or number so that you can meaningfully apply the `<` operator. If you need to do something more fancy, pass your own custom comparator function to [`Array.prototype.sort`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) instead. Note that this function is also available in Underscore OOP notation and in chains as `_(yourArray).sort`.

Related to #2848.

PR #1757, opened 2014-07-21
remove usage of Array.prototype.slice

michaelficarra:


jashkenas:
I don't see why this needs to be removed here...

PR #1758, opened 2014-07-21
Add _.slice function

megawac:
Add a slice function with the python contract `_.slice(collection, start, end, step)`

See https://docs.python.org/2/library/functions.html#slice. Like python slice, it doesn't do any coercing of inputs (besides 0 which is coerced to 1). Python will give an error for invalid inputs, this implementation doesn't bother and carries on. One nice benefit of the step feature is now we can do non destructive reversing `_.slice(arr, null, null, -1)`

Created some jsperf benchmarks the other day http://jsperf.com/cr-52768

I think it may make sense to add a `_.reverse` function after this which will also fix a bug `_(document.childNodes).reverse(); // Exception`

### DIFF
--- a/test/arrays.js
+++ b/test/arrays.js
@@ -2,6 +2,50 @@
 
   module('Arrays');
 
+  test('slice', function() {
+    var array = [1, 2, 3],
+        zeros = [, null, undefined];
+
+    deepEqual(_.slice(array, 1), [2, 3], 'should work with a positive `start`');
+
+    var expected = _.map(zeros, _.constant(array));
+    var actual = _.map(zeros, function(start) {
+      return _.slice(array, start);
+    });
+    deepEqual(actual, expected);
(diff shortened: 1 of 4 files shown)


jashkenas:
Are there any use cases for `step` outside of reversing?

jridgewell:
http://jsperf.com/cr-52768/6

jridgewell:
http://jsperf.com/cr-52768/6

I think this is trying to do too much. Can we make it a simple, dense version of `Array#slice`?

megawac:
@jashkenas from my experience, the step arg is most commonly used for reversing but it can also be useful for picking every 2nd element. I'm not sure if thats useful enough for us to entertain though (was fun to implement though)

jgonggrijp:
Not all review comments were addressed.

PR #1880, opened 2014-09-28
Fixing issue with sorting which fails when different datatypes are given...

eraser88:
... in the sortby list.

Fixing sorting issue mentioned below :

If i have objects like these : 
`[{"1", 100}, {"1", 1}, {"1", 2}, {"1", 200}]`
If i sort the elements using both the columns then the result should be :
`[{"1", 1}, {"1", 2}, {"1", 100}, {"1", 200}]`
But in the current code the result that is displayed is :
`[{"1", 1}, {"1", 100}, {"1", 2}, {"1", 200}]`

### DIFF
--- a/test/collections.js
+++ b/test/collections.js
@@ -504,15 +504,15 @@
     deepEqual(_.sortBy(collection, 'x'), collection, 'sortBy accepts property string');
 
     var collectionWithDifferentDatatypes = [
-      new Pair("1", 100), new Pair("1", 1),
-      new Pair("1", 2), new Pair("1", 200)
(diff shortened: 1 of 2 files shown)


jgonggrijp:
`_.sortBy` is for when you can reduce your values to a single string or number so that you can meaningfully apply the `<` operator. If you need to do something more fancy, pass your own custom comparator function to [`Array.prototype.sort`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) instead. Note that this function is also available in Underscore OOP notation and in chains as `_(yourArray).sort`.

Related to #2848.

PR #1919, opened 2014-11-05
Add _.chunk to Array functions.

### LINKED ISSUE #1891: _.chunk or _.stride function
A common function I'm sure many of us run into (especially regarding pagination) is the need to break an array into chunks of a specified length. I propose a `_.chunk()` (or `_.stride` if "chunk" isn't clear enough). It would also be useful in a number of other underscore features, including my proposal for `_.objects` https://github.com/jashkenas/underscore/issues/1890.

The first value would be the list, the second value would be the chunk-size/stride-length. If the chunk size doesn't divide evenly into the length of the list, the final item should be the remainder.

```javascript
_.chunk([1,2,3,4,5,6,7,8], 3);
```
Would yield
```javascript
[[1,2,3], [4,5,6], [7,8]]
```
It might be worth having an optional "remainder" boolean argument that could omit the remainder if set to false, but should probably be true by default. This would be useful in cases where the chunks were going to be mapped using `_.object` and dangling arguments weren't desirable.
```javascript
_.chunk([1,2,3,4,5,6,7,8], 3, false);
```
Would yield
```javascript
[[1,2,3], [4,5,6]]
```

keinkonzept:
Fixes #1891

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -635,6 +635,19 @@
 
     return range;
   };
+  
+  // Split an **array** into several arrays containing **int** or less elements
+  // of initial array
+  _.chunk = function(array, int) {
+    var result = [];
+    
+    if (int <= 0) return [];
+    if (int >= _.size(array)) return array;
+    
+    while (_.size(array) > 0) result.push(array.splice(0, int));
+    
+    return result;
+  };
 
   // Function (ahem) Functions
   // ------------------
--- a/underscore.js
+++ b/underscore.js
@@ -650,12 +650,15 @@
   // Split an **array** into several arrays containing **count** or less elements
   // of initial array
   _.chunk = function(array, count) {
+    count = count || 1;
+
+    if (count <= 1) return array.slice();
+
     var result = [];
+    var i = 0, length = array.length;
 
-    if (array == null || count <= 0) return [];
-
-    for (var i = 0, length = array.length; i < length;) {
-      result.push(array.slice(i, i += count));
+    while (i < length) {
+      result.push(slice.call(array, i, i += count));
     }
 
     return result;
--- a/underscore.js
+++ b/underscore.js
@@ -647,15 +647,16 @@
     return range;
   };
   
-  // Split an **array** into several arrays containing **int** or less elements
+  // Split an **array** into several arrays containing **count** or less elements
   // of initial array
-  _.chunk = function(array, int) {
+  _.chunk = function(array, count) {
     var result = [];
     
-    if (int <= 0) return [];
-    if (int >= _.size(array)) return array;
+    if (array == null || count <= 0) return [];
     
-    while (_.size(array) > 0) result.push(array.splice(0, int));
+    for (var i = 0, lng = array.length; i < lng; i += count) {
+      result.push(array.slice(i, i + int);
+    }
     
     return result;
   };
--- a/underscore.js
+++ b/underscore.js
@@ -653,9 +653,10 @@
     var result = [];
     
     if (array == null || count <= 0) return [];
+    if (count >= array.length) return array;
     
     for (var i = 0, lng = array.length; i < lng; i += count) {
-      result.push(array.slice(i, i + int);
+      result.push(array.slice(i, i + count));
     }
     
     return result;
--- a/underscore.js
+++ b/underscore.js
@@ -646,18 +646,18 @@
 
     return range;
   };
-  
+
   // Split an **array** into several arrays containing **count** or less elements
   // of initial array
   _.chunk = function(array, count) {
     var result = [];
-    
+
     if (array == null || count <= 0) return [];
-    
-    for (var i = 0, lng = array.length; i < lng;) {
+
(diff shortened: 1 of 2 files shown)


jridgewell:
Mind removing the trailing spaces on lines 649, 654, 656, and 660?

underscore.js
```diff
@@ -646,6 +646,20 @@
...
   };
+  
+  // Split an **array** into several arrays containing **count** or less elements
+  // of initial array
+  _.chunk = function(array, count) {
+    var result = [];
+    
+    if (array == null || count <= 0) return [];
+    
+    for (var i = 0, lng = array.length; i < lng;) {
```
jridgewell: Mind using `length` instead of `lng`?

stephennancekivell:
This would be good.

underscore.js
```diff
@@ -647,6 +647,20 @@
...
   };
 
+  // Split an **array** into several arrays containing **count** or less elements
+  // of initial array
+  _.chunk = function(array, count) {
+    var result = [];
+
+    if (array == null || count <= 0) return [];
+
+    for (var i = 0, length = array.length; i < length;) {
```
jridgewell: Mind using a while-loop? Every other for-loop in the codebase has an increment expression...

underscore.js
```diff
@@ -647,6 +647,20 @@
...
+      result.push(array.slice(i, i += count));
```
jridgewell: We should use `slice.call(array, i, i+= count)` to sidestep overwritten `Array#slice` and open this up to array-likes.

jridgewell:
@keinkonzept: Some final nitpicks, and I think this'll be good to merge.

megawac:
Based on how often this has been requested, I think it should be added

michaelficarra:
I think I agree at this point.

jridgewell:
That's three :+1:, then. Any last comments on the source before merging?

michaelficarra:
I think this implementation is broken for a count of 1. Also, I don't like having a default count. I'll comment more when in front of a computer.

megawac:
Seems fine for count 1 (its a tested case for default => 1 and 1). I've squashed the commits down and fixed style issues here:

https://github.com/jashkenas/underscore/compare/master...megawac:chunk

michaelficarra:
No, definitely broken. It should return an array of singletons.

underscore.js
```diff
@@ -647,6 +647,23 @@
     return range;
   };
 
+  // Split an **array** into several arrays containing **count** or less elements
+  // of initial array
+  _.chunk = function(array, count) {
+    count = count || 1;
+
+    if (count <= 1) return array.slice();
```
michaelficarra: Direct call to slice should instead use slice helper.

jridgewell: > No, definitely broken. It should return an array of singletons.

Should be `count < 1` to fix.

underscore.js
```diff
@@ -647,6 +647,23 @@
     return range;
   };
 
+  // Split an **array** into several arrays containing **count** or less elements
+  // of initial array
+  _.chunk = function(array, count) {
+    count = count || 1;
```
jridgewell: To support the `count = undefined` case and fix the `count = 1` case, we should default to `0` here.

megawac: I think the default = 1 makes sense. However do we want to mirror lodash
which makes anything less than 1 mean 1 chunk size?

E.g. _.chunk([1,2,3], 0) => [[1], [2], [3]]

On Tue, Jul 28, 2015 at 11:32 AM, Justin Ridgewell <notifications@github.com

megawac: I think that behaviour sort of makes sense so that chunk always returns an
array of arrays

megawac: The only other reasonable option I see is to return an empty array for
chunk < 1

jridgewell: That's fine with me, I was just coding the the current tests.

megawac: Hows this look @jridgewell https://github.com/jashkenas/underscore/compare/master...megawac:chunk

I'm still on the fence about `<1` case. I think empty array may be best

jridgewell: I think lodash has already set the prior art and breaking from it just causes compatibility headaches. We can discuss in [underdash](https://github.com/underdash/underdash/issues/4)?

megawac: /cc @jdalton

jdalton: I think the current behavior is correct if you're chunk is less than 1 there is no chunk and a shallow clone is returned.
Think of it like if you had a pie and you cut it into 0 slices you still have a whole pie.

megawac: A shallow clone isn't returned in lodash. It returns the array in chunks of 1. I think returning a shallow clone would be even worse.

jridgewell: I'm actually leaning with @megawac here. We're not cutting the array into 3 slices, we're gathering into groups of 3. And groups of 0 don't have anything.

jdalton: Hmm ok cool.

Issue #96, opened 2011-01-15
_.keys returns incorrect keys for arrays that have gaps


wheresrhys:
Accidentally closed this issue: https://github.com/documentcloud/underscore/issues/closed#issue/95

PR #1924, opened 2014-11-07
making _.unescape be able to unescape ' and ` escaped in Dec format

yuriskliar:
Let's have underscore unescape &amp;#39; and &amp;#96; These are valid HTML entities for ' and ` symbols.
Any browser works both with Dec and Hex representation of these symbols.

The issue is that Django web framework escapes ' and `  using Dec approach. And most probably other frameworks may do the same way.

So in such situation it's necessary to write some additional code which should probably be inside of the library.

### DIFF
--- a/test/utility.js
+++ b/test/utility.js
@@ -138,8 +138,8 @@
     equal(_.unescape(str), str, 'can unescape &amp;');
 
     // test unescape of ' and ` symbols escaped in Dec format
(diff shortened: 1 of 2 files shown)


jeegarshah:
underscore_string escapeHTML and unescapeHTML has more robust solution and unescapes dec or hex.

PR #1939, opened 2014-11-17
Add _.comparator and align sortedIndex to sortBy

### LINKED ISSUE #1768: Comparator edge-case handling
There are several inconsistencies with the comparator functions I feel should be addressed in 2.0. Mentioned this earlier in https://github.com/jashkenas/backbone/pull/3243
- `max`: `NaN` including `undefined` is considered the lowest item in the array. **Even lower then no item**

``` javascript
_.max([{a: {}}], 'a'); // => -Infinity
```
- `min`: same as max but `NaN` values are considered the highest items
- `sortBy`: special cases all `undefined` values to the end of the array. All `NaN` values are moved to the front of the array
- `sortedIndex`: **different from `sortBy`** (see https://github.com/lodash/lodash/commit/b07337f7a23a6607021981ab6ccdfdbff9e329e5#commitcomment-7164123): computed `NaN` is moved to the end of the list instead of the front

``` javascript
var array = [{a: NaN}, {a: NaN}, {a: {}}, {a: 1}, {a: 2}];
_.sortedIndex(array, {a: 3}, 'a'); // => 0
var sorted = _.sortBy(array.concat({a: 3}), 'a');
_.findIndex(sorted, {a: 3}); // => 5
```

---

I propose a `baseComparator(a, b)` used by those 4 functions (special case min?) which handles these cases consistently. I feel strongly about changing `sortBy`'s handling of `NaN` vs `undefined`. I think they should be considered `NaN < -Infinity < x`.

``` javascript
// Reasoning
_.sortBy([{a: 1}, {a: -Infinity}, {a: 2}, {}], 'a')
// Should => [{}, {a: -Infinity}, {a: 1}, {a: 2}]
```

### LINKED ISSUE #1834: _.sortedIndex align with _.sortBy
``` js
_.sortBy([undefined, 1, undefined, 2]);
// => [1, 2, undefined, undefined]
_.sortedIndex([1, 2, undefined, undefined], undefined);
// => 0; though should be 2
```

megawac:
Resolves #1768 and fixes #1834

This could use some review and thoughts

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -20,7 +20,7 @@
     "no-floating-decimal": 2,
     "no-inner-declarations": 2,
     "no-lonely-if": 1,
-    "no-nested-ternary": 2,
+    "no-nested-ternary": 0,
     "no-new-object": 0,
     "no-new-func": 0,
     "no-underscore-dangle": 0,
--- a/test/arrays.js
+++ b/test/arrays.js
@@ -136,6 +136,17 @@
     equal(_.sortedIndex(array, 2147483648), 2147483648, 'should work with large indexes');
   });
 
+  test('sortedIndex conforms to sortBy (#1768)', function() {
+    var sorted = _([{a: NaN}, {a: NaN}, {a: void 0}, {a: 0}, {a: 1}, {a: 2}, {a: 3}, {a: '4'}, {a: 'foo'}])
+                  .sortBy('a');
+    _.each([0, 1, 2, 3, '4', 'foo'], function(val) {
+      equal(_.sortedIndex(sorted, {a: val}, 'a'), _.findIndex(sorted, {a: val}), 'For val: ' + val);
+    });
+
+    sorted = _.sortBy([undefined, 1, undefined, 2]);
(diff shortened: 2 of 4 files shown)


braddunbar:
> Anyway I'm a big fan of exposing comparator as it allows devs easy
customization for oddball data structures and frankly I've needed a solid
comparator in my own code in the past

Can you expand on this? Wouldn't you just return a different value or use `#sort` directly in those cases?

jashkenas:
To repeat -- let's please back this out for the time being. (I don't want to forget about it.)

megawac:
@braddunbar lets say you want to to do add support to something similar to pythons class [`__cmp__`](https://docs.python.org/2.6/reference/datamodel.html#object.__cmp__)

```js
_.mixin({
   comparator: _.wrap(_.compatator, function(baseCompare, a, b) {
     if (_.isFunction(b.__cmp__)) return b.__cmp__(a);
     if (_.isFunction(a.__cmp__) return a.__cmp__(b);
     return baseCompare(a, b);
  })
});
```

PR #1942, opened 2014-11-17
Add _.call (for use when chaining)

almost:
When chaining it would be useful to be able to pass the whole object through a function, kind of like _.tap except the value is replaced with the return of the function.

Implementation is trivial and of course it doesn't make much sense when not chaining.

Anyone think this is a good idea?

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -1413,6 +1413,11 @@
     instance._chain = true;
     return instance;
   };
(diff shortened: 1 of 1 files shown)


jgonggrijp:
None of the review comments was addressed. Closing this until this changes.

Issue #1869, opened 2014-09-24
one method for updating object attributes


tkoomzaaskz:
I miss an underscore method that would take an object and update all it's attributes, within a iteratee and preserver all keys. Say I've got `{a:1, b:2, c:3}` and I want to use a square function to receive: `{a:1, b:4, c:9}`. Current solution is to do something like this:

```js
_.object(_.map(original, function (value, key) {
    return [key, value * value];
}));
```

but I don;t like that. It's not that readable - you need to use another 2 functions and know that _.object constructs another object from pairs. There could be one simple function for this, since updating all object attributes is not a rare task. How about adding it to underscore? An example solution can be found here:
http://stackoverflow.com/questions/26013802/underscore-js-iterate-over-object-attributes-and-modify-them

greenify:
> fully agree with @megawac. This is a common case afterall. (@tkoomzaaskz)

Any chance to get this into 1.8 e.g. with #1953?

megawac:
@jashkenas can you make a decision on this and #1953?

jashkenas:
This is a dupe of the implemented #1953, no? Let's move over there.

Issue #1887, opened 2014-10-11
hi _.map bug when handling object here is a fix 


pascaldls:
function( obj, iteratee, context ) {
            if (obj == null) return [];
            iteratee = _.iteratee(iteratee, context);
            var keys = obj.length !== +obj.length && _.keys(obj),
                length = (keys || obj).length,
                results = {},
                currentKey;
            for (var index = 0; index < length; index++) {
              currentKey = keys ? keys[index] : index;
              results[currentKey] = iteratee(obj[currentKey], currentKey, obj);
            }
            if ( _.isObject( obj ) && _.isArray( results) ) {
                return _.object( results ) ;
            } 
            return results;
        }

Issue #220, opened 2011-05-31
Object-preserving map() function


ThiefMaster:
Would it be possible to add a `map()`-like function (since changing the current one would break things) which works exactly like `map()` but returns an object instead of an array if an object was passed?

Example:

```
_.map2({abc: 123, xyz: '6', foo: '9'}, function(num, key) {
    return parseInt(num, 10);
});
=> {abc: 123, xyz: 6, foo: 9}
```

It might also make sense to have it return `[key, value]` tuples but just modifying the values might be sufficient for many cases.

PR #1953, opened 2014-11-30
Object-preserving map() function: mapValues

greenify:
This adds the `mapValues` methods.

It works exactly like `map`, but preserves the object.

```
_.map2({a: 1, b: 2}, function(val, key) {
    return val * 2;
});
=> {a: 2, b: 4}
```

There are at least three issues about this function (#220 #1867 #1887), and as I also do this frequently it would be awesome to get this into underscore.


megawac:
Closes #1869

jashkenas:
This looks good to me. Go for it.

PR #1964, opened 2014-12-04
Make _.clone clone Date, RegExp, and Number  #1915

### LINKED ISSUE #1915: _.clone doesn't clone Date, Regexp
Built-in types which match typeof==='object' are not cloned by _.clone, but instead are replaced with empty object {}.

    JSON.stringify(_.clone(new Date(0))) === '{}'

Possible solution: put checks in _.clone which uses copy-constructors for known types (e.g. Date, RegExp).

For example (Date):

    var copy = new Date(orig.getTime());

To reproduce:

    _.clone(new Date())
    _.clone(new RegExp())

Both return an empty object.

Obviously it is impossible and impractical to make clone work with *all* objects out there, but supporting the JS built-in types will prevent others from experiencing my recent headache.

oveddan:
Fixes #1915

### DIFF
--- a/test/objects.js
+++ b/test/objects.js
@@ -268,6 +268,40 @@
     equal(_.clone(undefined), void 0, 'non objects should not be changed by clone');
     equal(_.clone(1), 1, 'non objects should not be changed by clone');
     equal(_.clone(null), null, 'non objects should not be changed by clone');
+
+    var stringObject = new String('abc');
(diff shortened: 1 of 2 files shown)


jridgewell:
Closed by https://github.com/jashkenas/underscore/commit/9eb89257ffacae4986d8068f1d1812dde6ee8f07. https://github.com/jashkenas/underscore/issues/1915#issuecomment-66702790

PR #1981, opened 2014-12-18
Add _.findLastIndex and handle NaN in *IndexOf

megawac:
This fixes https://github.com/jashkenas/underscore/issues/1750 and adds a `_.findLastIndex` function

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -666,7 +666,7 @@
       idx = from < 0 ? idx + from + 1 : Math.min(idx, from + 1);
     }
     if (item !== item) {
-      return _.findLastIndex(array.slice(0, idx), _.isNaN);
+      return _.findLastIndex(slice.call(array, 0, idx), _.isNaN);
     }
     while (--idx >= 0) if (array[idx] === item) return idx;
     return -1;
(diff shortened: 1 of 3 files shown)


underscore.js
```diff
@@ -648,9 +654,18 @@
   // Returns the first index on an array-like that passes a predicate test
   _.findIndex = function(array, predicate, context) {
     predicate = cb(predicate, context);
-    var length = array != null ? array.length : 0;
-    for (var i = 0; i < length; i++) {
-      if (predicate(array[i], i, array)) return i;
+    var length = array != null && array.length;
+    for (var index = 0; index < length; index++) {
```
jridgewell: Is there a reason to use `index` over `i`?

megawac: Nope, but we should make up our mind and be consistent.

jridgewell: I think it's already `i`. :stuck_out_tongue:

megawac: Thanks for the :eyes: @jridgewell

megawac:
Any objections to this current implementation?

PR #1988, opened 2014-12-20
add iteration fixes of _.keys to _.isEmpty

### LINKED ISSUE #1983: _.isEmpty should use _.keys
The iteration fixes of _.keys should be used by _.isEmpty.

maxbeatty:
This is an attempt to fix #1983 

I tried reusing parts of _.keys that made sense to detect if an object was truly empty in IE < 9

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -1143,12 +1143,22 @@
     return eq(a, b, [], []);
   };
 
+  /**
+  * Used as the maximum length of an array-like value.
+  * See the [ES6 spec](http://people.mozilla.org/~jorendorff/es6-draft.html#sec-tolength)
+  * for more details.
+  */
+  var MAX_SAFE_INTEGER = Math.pow(2, 53) - 1;
+  function isLength(value) {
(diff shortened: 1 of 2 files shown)


maxbeatty:
Is there anything left to do here?

underscore.js
```diff
@@ -1143,13 +1143,23 @@
...
   };
 
+  /**
+  * Used as the maximum length of an array-like value.
+  * See the [ES6 spec](http://people.mozilla.org/~jorendorff/es6-draft.html#sec-tolength)
+  * for more details.
+  */
+  var MAX_SAFE_INTEGER = Math.pow(2, 53) - 1;
+  function isLength(value) {
+    return typeof value == 'number' && value > -1 && value <= MAX_SAFE_INTEGER;
```
megawac: /cc #1990

underscore.js
```diff
@@ -1143,13 +1143,23 @@
     return eq(a, b, [], []);
   };
 
+  /**
```
akre54: We don't use docblocks anywhere else in the code. Please use inline comments here.

Issue #1590, opened 2014-04-17
Each iterates badly when it contains the `length` field


sgenoud:
Each iterates correctly on most objects:

```.js
_.each({a:2}, function(value, key) { console.log('value: ' + value + ', key:' + key) });
value: 2, key:a 
```

It fails if the object contains a `length` property:

```.js
_.each({a:2, length:4}, function(value, key) { console.log('value: ' + value + ', key:' + key) });
value: undefined, key:0
value: undefined, key:1
value: undefined, key:2
value: undefined, key:3 
```

The source of the bug is that the object is considered an "Array-like": https://github.com/jashkenas/underscore/blob/master/underscore.js#L69

jashkenas:
Closing this ol' ticket down — in the absence of `arrayMap`, `arrayFilter` etc ... just doing an `arrayEach` doesn't seem like a good idea.

YuriGor:
Hi folks, so is there any workaround to fix this?
I have to iterate over echarts3 options object recursively, and there are a LOTS of 'length' keys so I fail often.

PR #1990, opened 2014-12-28
Implement a more rigid is length check

megawac:
Stuck in an airport today so going to get around to a couple of these things we really should get it

As discussed in https://github.com/jashkenas/underscore/issues/1590 and others. [Here's jdd's implementation for a reference](https://github.com/lodash/lodash/blob/master/lodash.js#L3646-3655), also as discussed in that thread it may make sense to check for a `splice`/`slice` function to  consider a method a true array in places like  `_.isEmpty`

### DIFF
--- a/test/.eslintrc
+++ b/test/.eslintrc
@@ -14,7 +14,8 @@
     "notEqual": false,
     "throws": false,
     "asyncTest": false,
-    "start": false
+    "start": false,
+    "expect": false
   },
   "rules": {
     "brace-style": 0,
--- a/test/collections.js
+++ b/test/collections.js
@@ -69,6 +69,50 @@
     });
   });
 
+  test('Iterating objects with sketchy length properties', function() {
+    var functions = [
+        'each', 'map', 'filter', 'find',
+        'some', 'every', 'max', 'min',
+        'groupBy', 'countBy', 'partition', 'indexBy'
+    ];
+    var reducers = ['reduce', 'reduceRight'];
+
+    var tricks = [
+      {length: '5'},
+      {
+        length: {
+          valueOf: _.constant(5)
+        }
+      },
+      {length: Math.pow(2, 53) + 1},
+      {length: Math.pow(2, 53)},
+      {length: null},
+      {length: -2},
+      {length: new Number(15)}
+    ];
+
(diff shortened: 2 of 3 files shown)


megawac:
Any objections to current form (need :eyes:)

michaelficarra:
LGTM.

jridgewell:
:+1:

jashkenas:
Missed the discussion, but this is lame — let's revert this.

It covers a few edge cases, but misses the 95% overlap use case, and is merely a fig leaf gesture in the vague direction of correctness.

megawac:
It's not only a shot at improving edge correctness but aims to improve consistency throughout the codebase. Before this commit the is length check was implemented differently in different places (e.g. `uniq`).

This will also make it easier to handle other edge cases in certain places if someone takes a shot at further refactors. For instance, it may be possible to drop the `null`/`undefined` checks in certain places with this. Another change I would like to propose through `isArrayLike` is fixing #1992

jashkenas:
> For instance, it may be possible to drop the null/undefined checks in certain places with this. Another change I would like to propose through isArrayLike is fixing #1992

Both of those might be fine concepts — but if they are, they should be part of this PR. At the moment, it's merely two steps backward — one for the extra function redirection, and one for the silly MAX_ARRAY_INDEX baloney. Let's revert it.

jdalton:
Having the check moved to a function improves readability. I've seen devs ask multiple times what `obj.length === +obj.length` does but if it was named `isLength` that may give them a better idea of the check.

The function is small and so gets inlined easily when JITed. In this PR the check looks to be based on ES6's [ToLength](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-tolength) helper which indirectly references [MAX_SAFE_INTEGER](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-number.max_safe_integer).

Another nice thing is a helper like `isLength` keeps the check in one place so if later it's decided to expand the check code only needs to be changed in one place instead of every method.

```js
function isLength(value) {
  return typeof value == 'number' && value > -1 && value % 1 == 0 && value <= MAX_SAFE_INTEGER;
}
```

PR #1991, opened 2014-12-29
Wrap redundant reduce and reduceRight code in helper

### LINKED ISSUE #1734: Reduce reduce
Currently `_.reduce` and `_.reduceRight` are really similar. They could be reduced by having a base form of `reduce` that also accepts an iteration helper either `each` or `eachRight`. This may be applicable to other areas too. For reference see [this](https://github.com/lodash/lodash/blob/66fc3c71573cdfcbdd2160b6ae7d6cbd2bfbbc2c/dist/lodash.js#L4945-L4972) & [this](https://github.com/lodash/lodash/blob/66fc3c71573cdfcbdd2160b6ae7d6cbd2bfbbc2c/dist/lodash.js#L2135-L2156).

megawac:
The loops could be further abstracted, but I didn't like how it looked after going that far

Fixes #1734

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -160,24 +160,30 @@
     return results;
   };
 
-  // Wrapping function to construct a reducer function
+  // Create a reducing function iterating left or right.
   function createReduce(dir) {
-    return function(obj, iteratee, memo, context) {
-      if (obj == null) obj = [];
-      iteratee = optimizeCb(iteratee, context, 4);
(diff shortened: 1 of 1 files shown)


jashkenas:
Nice. One tiny style thing -- comments should be full sentences. Periods and all.

bkillenit:
Yes, agreed, comments should be fully comprehensible sentences.

PR #1993, opened 2014-12-29
Implement zip via unzip

megawac:
As @jdalton noted a while ago https://github.com/jashkenas/underscore/pull/1806#discussion_r16799291 (though wasn't possible at the time until #1520 was merged)


jashkenas:
Let's tag 'em as we close 'em.

PR #1994, opened 2014-12-29
Invoke shouldn't throw for null or undefined

megawac:
This has been brought up before

 - https://github.com/jashkenas/underscore/pull/1520#issuecomment-37668674

I often need to do an invoke where some items may or may not have a function defined (should this only check if its defined?). So I do something like

```js
var someCollection = [...];

_.chain(someCollection).filter('update').invoke('update');
```

Anyway, this is consistent with the rest of the codebase if nothing else

### DIFF
--- a/test/collections.js
+++ b/test/collections.js
@@ -406,7 +406,7 @@
     strictEqual(_.includes, _.contains, 'alias for includes');
   });
 
-  test('invoke', 3, function() {
+  test('invoke', 5, function() {
     var list = [[5, 1, 7], [3, 2, 1]];
     var result = _.invoke(list, 'sort');
     deepEqual(result[0], [1, 5, 7], 'first array sorted');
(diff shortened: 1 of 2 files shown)


jashkenas:
I don't really like all of this silent-error-in-your-codebase-propagation that we're doing ... but if that's the way things are going, then I guess we should go there.

michaelficarra:
> I don't really like all of this silent-error-in-your-codebase-propagation that we're doing

Me either. I hate it. I think it's the worst thing about underscore by far. I've been very vocal about this on other issues. Can we *please* get rid of all the `null` propagation so users get timely errors as close to the source as possible?

jdalton:
I think it's a great feature so there :P
It makes things just work & cuts down on guard checks devs have to repeatedly put in place. These checks come out of devs complaining about that and offers a real usability win here.

akre54:
But the downside is that errors are much harder to find and fix when they do happen. I'd take a quick fix for a bad data problem over a hidden problem in production any day of the week and twice on Sunday.

jdalton:
>  I'd take a quick fix for a bad data problem 

Let's redirect discussion to #2000 to avoid repeating ourselves a ton.

braddunbar:
> any day of the week and twice on Sunday.

Ha! I've never heard that particular turn of phrase before. :smile:

PR #1998, opened 2015-01-03
Allow nested checks in _.has

VictorBjelkholm:
Adds some basic support to be able to do something like this:
```
var obj = {foo: {bar: "Hello"}};
console.assert(_.has(obj, 'foo.bar'));
console.assert(!_.has(obj, 'foo.foo'));
```

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -122,6 +122,13 @@
     return result;
   };
 
+	// An internal function to resolve object from string of properties.
+	var resolveObject = function(obj, path) {
+		return [obj || self].concat(path.split('.')).reduce(function(prev, curr) {
+			return prev[curr];
+		});
+	};
+
   // Collection Functions
   // --------------------
 
@@ -1239,6 +1246,12 @@
   // Shortcut function for checking if an object has a given property directly
   // on itself (in other words, not on a prototype).
   _.has = function(obj, key) {
+    if (_.isString(key) && key.indexOf('.') !== -1) {
+      var keys = key.split('.');
+      key = keys.pop();
+      var path = keys.join('.');
+      obj = resolveObject(obj, path);
+    }
     return obj != null && hasOwnProperty.call(obj, key);
   };
 
--- a/test/objects.js
+++ b/test/objects.js
@@ -730,7 +730,7 @@
   });
 
   test('has', function () {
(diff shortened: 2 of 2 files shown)


underscore.js
```diff
@@ -122,6 +122,13 @@
     return result;
   };
 
+	// An internal function to resolve object from string of properties.
+	var resolveObject = function(obj, path) {
+		return [obj || self].concat(path.split('.')).reduce(function(prev, curr) {
+			return prev[curr];
```
spadgos: This would throw on `resolveObject({ }, 'foo.bar')`. Perhaps `return prev && prev[curr]`?

underscore.js
```diff
@@ -122,6 +122,13 @@
     return result;
   };
 
+	// An internal function to resolve object from string of properties.
+	var resolveObject = function(obj, path) {
+		return [obj || self].concat(path.split('.')).reduce(function(prev, curr) {
```
spadgos: Why put the object into the array? How about:

```js
path.split('.').reduce(function(prev, curr) { ... }, obj || self);
```

underscore.js
```diff
@@ -1239,6 +1246,12 @@
   // Shortcut function for checking if an object has a given property directly
   // on itself (in other words, not on a prototype).
   _.has = function(obj, key) {
+    if (_.isString(key) && key.indexOf('.') !== -1) {
+      var keys = key.split('.');
+      key = keys.pop();
+      var path = keys.join('.');
+      obj = resolveObject(obj, path);
+    }
```
spadgos: I guess this would mean that you could never check for the existence of a key which contains a dot:

```js
_.has({ 'my.key': true}, 'my.key') === false
```

You could make it more like:

```js
return obj != null && (hasOwnProperty.call(obj, key) || resolveObject(obj, path));
```

*(There'd be an issue there since resolve object returns the value, not the bool showing existence of the key, but you get the idea.)*

VictorBjelkholm: Ah, thanks for the input Nick! I'll look into when I get a chance. Thanks.

Sincerely,
Victor Bjelkholm

megawac:
You can use underscore contribs rock solid hasPath but were not planning on adding string programming to core. Search the past issues for string programming

PR #1999, opened 2015-01-04
Add function to get the modulus index of array

jisaacks:
I find myself doing this a lot. Its a simple addition but I think it would be a great convenience to have this.

I didn't bother adding tests or building. I figured if you find this acceptable I can always add those.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -494,6 +494,12 @@
     return slice.call(array, n == null || guard ? 1 : n);
   };
 
+  //Returns the element at the modulus index
+  _.mod = function(array, modulo) {
+    var index = modulo % array.length;
+    return array[index];
+  }
+
   // Trim out all falsy values from an array.
   _.compact = function(array) {
     return _.filter(array, _.identity);


akre54:
Thanks for the pull. I think this doesn't fit with the rest of Underscore's general-purpose, composable utility functions and is a bit too special-case for inclusion. You should feel free to `mixin` to your copy of Underscore or propose at Underscore-contrib, however.

jisaacks:
I think it is really useful and general purpose as it allows unlimited iteration over a limited set of things. 

For example, if you had an array of colors like `colors = ['red', 'blue', 'green', 'orange'];` And you wanted to apply the colors over a large set of things, you could do `_.mod(colors, i)` at any index like 27 and get the correct color. I personally do not think thats a very special case. However, I could be bias from my own experiences. 

Underscore-contrib appears to have a `_.mod` but that just finds the modulo value, I think getting an item from an array is more useful. So I may change the name and make a PR there.

Thanks.

akre54:
I've written code like that quite a few times. But I almost always feel it reads better inline than with a helper. Personal preference, really.

```js
color = colors[ i % colors.length ];
```

jisaacks:
Fair enough.

Issue #1996, opened 2014-12-30
can _.bindAll accept an array of function names


bishopZ:
I wrote this Stackoverflow issue for this common (imo) Underscore + Backbone pattern. 
http://stackoverflow.com/questions/27697299/passing-an-array-of-names-to-bindall-arguments-list-has-wrong-type/

The best answer was to do this:
```
_.bindAll.apply(_, [this].concat(_.values(this.events)));
```
Would be much more readable if it could be written like this:
```
_.bindAll(_.values(this.events));
```
Many of the Underscore functions accept both a string or an array of strings, why not this one?

akre54:
I'm a little confused. The values from `this.events` are already auto-bound by `delegateEvents` to your view. In general and with only a few common exceptions, I feel `bindAll` is an anti-pattern. Personally I prefer to bind at the point of passing to a callback, assigning a `context` arg, or where absolutely necessary binding in `initialize`. 

This does add consistency with the rest of these type methods though, so I'm fine with including it.

bishopZ:
```delegateEvents``` re-adds the event functions to the view's DOM elements, but it does not bind the event function to the view. Even after calling ```delegateEvents``` (or initializing the view), the ```this``` keyword still references the event object and not the view object. Calling ```_.bindAll``` causes ```this``` to be the view object rather than the event object.

**Comment by akre54** [comment]
That simply isn't true. The [method is bound](https://github.com/jashkenas/backbone/blob/1.1.2/backbone.js#L1077) to the view instance within delegateEvents.

You can try it for yourself:

```js
var View = Backbone.View.extend({
  events: {
    'click': 'showThis'
  },
  showThis: function() {
    console.log(this); // the view instance
  }
});

var view = new View;
view.$el.trigger('click');
view.showThis();
```

bishopZ:
You are correct. Not sure how I got the wrong impression with that.

We do use _.bindAll to bind view methods that get used as callbacks for ajax callbacks, event bus handlers and animation callbacks. I did not know that the event handlers were already proxied.

akre54:
And all those methods you either know ahead of time (in which case you can call bindAll as normal) or can bind in place (as with the animation callbacks), right?

bishopZ:
We can. ```_.bindAll``` in this case helps us centralize all the proxies in a view. This improves readability and makes troubleshooting easier.

[...]

PR #2005, opened 2015-01-08
PR for issue #1996, enhance _.bindAll

bishopZ:


megawac:
Use `flatten()` as we do in `omit` and `pick`

So something like

```js
function bindAll(obj) {
    var keys = flatten(arguments, false, false, 1);
    var length = keys.length;
    if (length < 1) throw new Error('bindAll must be passed function names');
    while (length--) {
      var key = keys[i];
      obj[key] = _.bind(obj[key], obj);
    }
    return obj;
};
```

Also add some tests svp :smiley_cat:

jridgewell:
Closing in favor of #2141.

PR #2010, opened 2015-01-16
Small but important fix in official documentation/homepage (index.html) - 5.2kb => 15.2kb minified 

trebor74hr:
and Gzipped


jridgewell:
It's 15.3kb without gzip, 5.2kb with `gzip -9`.

trebor74hr:
It was too late ;) ... Anyway, maybe you should note out minified-not-zipped size 15.3K too?

PR #2003, opened 2015-01-06
Add engine proxy methods to toJSON, toString and valueOf

megawac:
Fixes #1562

PR #1965, opened 2014-12-04
Allow customizing `_.iteratee`.

jridgewell:
@jdalton [brought up](https://github.com/jashkenas/underscore/issues/1883#issuecomment-57821014) that unexposing `_.iteratee`'s' `argCount` disallows overriding `_.iteratee` to [customize the way collection functions iterate](https://github.com/jashkenas/underscore/issues/1883#issuecomment-57826279).

PR #2012, opened 2015-01-19
Prepare 1.8 release

### LINKED ISSUE #2000: _.invoke & nullish elements.
Follow up to #1994.

``` js
_.invoke(['a', null, undefined, 'd'], 'toUpperCase')
```

would still error

megawac:
Here we go...
========

Please commit directly to this branch if you feel inclined, I'm going to be away for the rest of the day and travelling tomorrow

To dos:

- [x] Doc `findIndex` #1587 
- [ ] doc ``findKey` with example
- [x] Document `_.findLastIndex` #1981 
- [ ] Doc proxy methods https://github.com/jashkenas/underscore/pull/2003
- [x] doc #1953 
- [ ] doc #2029 

Prereqs (can people take a look at these) before we can release:

- [x] https://github.com/jashkenas/underscore/pull/1990
- [x] **#1981**
- [x] #1983, #1988 (requires #1990)

Perhaps issues

- [x] #1590 (add `forIn`?)
- [ ] **#1965**
- [x] **#1953**
- [ ] #1919 
- [ ] #1758
- [x] #2003 
- [ ] Revert #1994 or fix #2000

Reference:
https://github.com/jashkenas/underscore/compare/1.7.0...master

http://rawgit.com/jashkenas/underscore/1.8/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -1001,6 +1001,8 @@ _.unzip(["moe", 30, true], ["larry", 40, false], ["curly", 50, false]);
 _.object(['moe', 'larry', 'curly'], [30, 40, 50]);
 =&gt; {moe: 30, larry: 40, curly: 50}
 
+_.object([['moe', 30], ['larry', 40], ['curly', 50]]);
+=&gt; {moe: 30, larry: 40, curly: 50}
 </pre>
 
       <p id="indexOf">
--- a/index.html
+++ b/index.html
@@ -245,6 +245,8 @@
       <li>- <a href="#indexOf">indexOf</a></li>
       <li>- <a href="#lastIndexOf">lastIndexOf</a></li>
       <li>- <a href="#sortedIndex">sortedIndex</a></li>
+      <li>- <a href="#findIndex">findIndex</a></li>
+      <li>- <a href="#findLastIndex">findLastIndex</a></li>
       <li>- <a href="#range">range</a></li>
     </ul>
 
@@ -275,9 +277,11 @@
       <li>- <a href="#keys">keys</a></li>
       <li>- <a href="#keysIn">keysIn</a></li>
       <li>- <a href="#values">values</a></li>
+      <li>- <a href="#mapValues">mapValues</a></li>
       <li>- <a href="#pairs">pairs</a></li>
       <li>- <a href="#invert">invert</a></li>
       <li>- <a href="#object-functions">functions</a></li>
+      <li>- <a href="#findKey">findKey</a></li>
       <li>- <a href="#extend">extend</a></li>
       <li>- <a href="#assign">assign</a></li>
       <li>- <a href="#create">create</a></li>
@@ -1050,6 +1054,46 @@ _.sortedIndex(stooges, {name: 'larry', age: 50}, 'age');
 =&gt; 1
 </pre>
 
+      <p id="findIndex">
+        <b class="header">findIndex</b><code>_.findIndex(array, predicate, [context])</code>
+        <br />
+        Similar to <a href="#indexOf"><tt>_.indexOf</tt></a>, returns the first index
+        where the <strong>predicate</strong> truth test passes; otherwise returns <i>-1</i>.
+      </p>
+      <pre>
+function isPrime(element, index, array) {
+  var start = 2;
+  while (start <= Math.sqrt(element)) {
+    if (element % start++ < 1) {
+      return false;
+    }
+  }
+  return element > 1;
+}
+_.findIndex([4, 6, 8, 12], isPrime);
+=&gt; -1 // not found
+_.findIndex([4, 6, 7, 12], isPrime);
+=&gt; 2
+</pre>
+
+      <p id="findLastIndex">
+        <b class="header">findLastIndex</b><code>_.findLastIndex(array, predicate, [context])</code>
+        <br />
+        Like <a href="#findIndex"><tt>_.findIndex</tt></a> but iterates the array in reverse,
+        returning the index closest to the end where the <strong>predicate</strong> truth test
+        passes.
+      </p>
+      <pre>
+var users = [{'id': 1, 'name': 'Bob', 'last': 'Brown'},
+             {'id': 2, 'name': 'Ted', 'last': 'White'},
+             {'id': 3, 'name': 'Frank', 'last': 'James'},
+             {'id': 4, 'name': 'Ted', 'last': 'Jones'}];
+_.findLastIndex(users, {
+  name: 'Ted'
+});
+=&gt; 3
+</pre>
+
       <p id="range">
         <b class="header">range</b><code>_.range([start], stop, [step])</code>
         <br />
@@ -1350,6 +1394,20 @@ _.values({one: 1, two: 2, three: 3});
 =&gt; [1, 2, 3]
 </pre>
 
+<p id="mapValues">
+        <b class="header">mapValues</b><code>_.mapValues(array, iteratee, [context])</code>
+        <br />
+        Iterates an object and returns a new object with the values computed by
+        the given <tt>iteratee</tt>.
+      </p>
+      <pre>
+_.mapValues({a: 1, b: 2}, function(val, key) {
+    return val * 2;
+});
+=> {a: 2, b: 4}
+=&gt; 3
+</pre>
+
       <p id="pairs">
         <b class="header">pairs</b><code>_.pairs(object)</code>
         <br />
@@ -1382,6 +1440,17 @@ _.invert({Moe: "Moses", Larry: "Louis", Curly: "Jerome"});
       <pre>
 _.functions(_);
 =&gt; ["all", "any", "bind", "bindAll", "clone", "compact", "compose" ...
+</pre>
+
+<p id="findKey">
+        <b class="header">findKey</b><code>_.findKey(object, predicate, [context])</code>
+        <br />
+        Similar to <a href="#findIndex"><tt>_.findIndex</tt></a> but iterates objects.
+        Returns the <i>key</i> where the <strong>predicate</strong> truth test
+        passes or <i>undefined</i>.
+      </p>
+      <pre>
+
 </pre>
 
       <p id="extend">
@@ -2186,7 +2255,9 @@ _([1, 2, 3]).value();
             Added <tt>_.keysIn</tt> which returns the enumerable properties of an object.
           </li>
           <li>
-            Reverted change where <tt>_.extend</tt> only iterates own properties, it now uses <tt>_.keysIn</tt>. Added <tt>_.assign</tt> that aligns closely with ECMA6 <tt>Object.assign</tt> and iterates over own properties.
+            Reverted change where <tt>_.extend</tt> only iterates own properties, it now
+            uses <tt>_.keysIn</tt>. Added <tt>_.assign</tt> that aligns closely with ECMA6
+            <tt>Object.assign</tt> and iterates over own properties.
           </li>
 
           <li>
@@ -2194,11 +2265,22 @@ _([1, 2, 3]).value();
           </li>
 
           <li>
-            Added the handy <tt>_.transform</tt> method as an alternative to <tt>_.reduce</tt> in many cases.
-          </li
->
+            Added <tt>_.mapValues</tt>.
+          </li>
+
           <li>
-            Added <tt>_.isError</tt> which determines if an object is derived from the <tt>Error</tt> object.
+            Added <tt>_.findIndex</tt> (similar to <a href="https://developer.mozilla.org/Web/JavaScript/Reference/Global_Objects/Array/findIndex">Array.prototype.findIndex</a>),
+            <tt>_.findLastIndex</tt>, and <tt>_.findKey</tt>.
+          </li>
+
+          <li>
+            Added the handy <tt>_.transform</tt> method as an alternative to <tt>_.reduce</tt>
+            in many cases.
+          </li>
+
+          <li>
+            Added <tt>_.isError</tt> which determines if an object is derived from the <tt>Error</tt>
+            object.
           </li>
 
           <li>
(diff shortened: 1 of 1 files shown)


jridgewell:
If we don't revert #1994, I'd like to see #2000 get addressed.

index.html
```diff
@@ -1275,17 +1290,31 @@ <h2 id="objects">Object Functions</h2>
...
       </p>
       <pre>
 _.keys({one: 1, two: 2, three: 3});
 =&gt; ["one", "two", "three"]
 </pre>
 
+      <p id="keysIn">
+        <b class="header">keysIn</b><code>_.keysIn(object)</code>
+        <br />
+        Retrieve all the inheritted and own properties of an <b>object</b>.
```
jridgewell: If we say `enumerable` to describe `_.keys`, we should do the same here.

index.html
```diff
@@ -1339,6 +1368,49 @@ <h2 id="objects">Object Functions</h2>
 =&gt; {name: 'moe', age: 50}
 </pre>
 
+<p id="assign">
+        <b class="header">assign</b><code>_.assign(destination, *sources)</code>
+        <br />
+        Like <a href="https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object/assign"><tt>Object.assign</tt></a> and similar to <tt>_.extend</tt>, this function copies all of the <strong>own</strong> enumerable properties of the <b>source</b> objects over to the destination object.
```
jridgewell: Mind breaking this line up?

index.html
```diff
@@ -1770,17 +1868,21 @@ <h2 id="utility">Utility Functions</h2>
 =&gt; "Curly, Larry &amp; Moe"</pre>
 
       <p id="result">
-        <b class="header">result</b><code>_.result(object, property)</code>
+        <b class="header">result</b><code>_.result(object, property, [defaultValue])</code>
         <br />
         If the value of the named <b>property</b> is a function then invoke it
-        with the <b>object</b> as context; otherwise, return it.
+        with the <b>object</b> as context; otherwise, return it. If a default value
+        is provided and the property doesn't exist than the default will be returned.
```
jridgewell: We should add that `defaultValue` is invoked if it's a function.

megawac: Have that in change log, could fit it in up here as well

index.html
```diff
@@ -1593,6 +1677,20 @@ <h2 id="objects">Object Functions</h2>
...
 
+      <p id="isError">
+        <b class="header">isError</b><code>_.isError(object)</code>
+        <br />
+        Returns <i>true</i> if <b>object</b> inherrits from an Error.
+      </p>
+      <pre>
+try {
+  throw new TypeError("Example");
+} catch (o_O) {
```
akre54: how about just `e` here instead of `o_0`. No need to get fancy

index.html
```diff
@@ -1770,17 +1868,21 @@ <h2 id="utility">Utility Functions</h2>
 =&gt; "Curly, Larry &amp; Moe"</pre>
 
       <p id="result">
-        <b class="header">result</b><code>_.result(object, property)</code>
+        <b class="header">result</b><code>_.result(object, property, [defaultValue])</code>
```
akre54: /s/defaultValue/default. Fits a bit better with the rest of the signature.

index.html
```diff
@@ -2044,6 +2146,76 @@ <h2 id="links">Links &amp; Suggested Reading</h2>
...
+          <li>
+            Reverted change where <tt>_.extend</tt> only iterates own properties, it now uses <tt>_.keysIn</tt>. Added <tt>_.assign</tt> that aligns closely with ECMA6 <tt>Object.assign</tt> and iterates over own properties.
+          </li>
+
+          <li>
+            Added <tt>_.create</tt> helper for simpler inheritance patterns.
+          </li>
+
+          <li>
+            Added extremely handy <tt>_.transform</tt> method as an alternative to <tt>_.reduce</tt> in many cases.
```
akre54: "extremely handy" is a bit much here. Doesn't really fit with the rest of the changelog

index.html
```diff
@@ -2044,6 +2146,76 @@ <h2 id="links">Links &amp; Suggested Reading</h2>
...
+          </li>
+          <li>
+            <tt>_.contains</tt> can be provided an optional start index.
+          </li>
+          <li>
+            <tt>_.indexOf</tt> and <tt>_.lastIndexOf</tt> now support search for <tt>NaN</tt>.
+          </li>
+
+          <li>
+            <tt>_.iteratee</tt> no longer takes the wrapped functions expected argument length.
```
akre54: `wrapped functions` -> `wrapped functions'` or `wrapped function's`.

This could also be worded a bit better. Maybe "`_.iteratee` no longer exposes the internal `argCount` parameter for [...]"

index.html
```diff
@@ -2044,6 +2146,76 @@ <h2 id="links">Links &amp; Suggested Reading</h2>
...
+          </li>
+          
+          <li>
+            Fixed a bug where <tt>_.keys</tt> and other object and collection methods did not iterate several
```
akre54: be more specific with several. Perhaps "did not iterate over reserved properties"

index.html
```diff
@@ -2044,6 +2146,76 @@ <h2 id="links">Links &amp; Suggested Reading</h2>
...
+            properties in several old IE environments.
```
akre54: remove the second several. It's a bit redundant here to the plural in environments

index.html
```diff
@@ -2044,6 +2146,76 @@ <h2 id="links">Links &amp; Suggested Reading</h2>
...
+          </li>
+          <li>
+            A number of code simplifications, changes for improved cross platform compatibility, and a number of edge case bug fixes.
```
akre54: Use "a number of" only once here. Something a bit more lively for either the first or the second usage

index.html
```diff
@@ -1083,9 +1093,14 @@ <h2 id="functions">Function (uh, ahem) Functions</h2>
         left open to supply at call-time.
       </p>
       <pre>
-var add = function(a, b) { return a + b; };
-add5 = _.partial(add, 5);
-add5(10);
```
akre54: :+1:

index.html
```diff
@@ -959,11 +971,9 @@ <h2 id="arrays">Array Functions</h2>
         If duplicate keys exist, the last value wins.
       </p>
       <pre>
-_.object(['moe', 'larry', 'curly'], [30, 40, 50]);
-=&gt; {moe: 30, larry: 40, curly: 50}
+_.unzip(["moe", 30, true], ["larry", 40, false], ["curly", 50, false]);
```
akre54: Why is the unzip example below the `_.object` documentation? And why did you get rid of the _.object example?

index.html
```diff
@@ -1275,17 +1290,31 @@ <h2 id="objects">Object Functions</h2>
...
 =&gt; ["one", "two", "three"]
 </pre>
 
+      <p id="keysIn">
+        <b class="header">keysIn</b><code>_.keysIn(object)</code>
+        <br />
+        Retrieve all the inheritted and own properties of an <b>object</b>.
+      </p>
+      <pre>
+function Class() {
```
akre54: Let's use something a bit more interesting than `Class`. Maybe Person, Animal, whatever.

index.html
megawac: typos - ill fix it when I get back if noone else wants to get it

index.html
michaelficarra: `default` is a reserved word.

akre54: Right. It's just for the documentation, not the source.

michaelficarra: Still, that would be confusing.

megawac:
Hows this look

index.html
```diff
@@ -513,6 +519,28 @@ <h2 id="collections">Collection Functions (Arrays or Objects)</h2>
 =&gt; [4, 5, 2, 3, 0, 1]
 </pre>
 
+<p id="transform">
+        <b class="header">transform</b><code>_.transform(collection, iteratee, [memo], [context])</code>
+        <br />
+        <tt>_.transform</tt> is an alternative to <tt>_.reduce</tt> that transforms
+        the given collection into a new collection. Unlike <tt>_.reduce</tt> if the
+        memo isn't specified it will default to a new instance of the type of the collection.
```
akre54: This section is a bit confusing to me, I'm not really sure what it's getting at.

index.html
```diff
@@ -513,6 +519,28 @@ <h2 id="collections">Collection Functions (Arrays or Objects)</h2>
...
+        Another distinction is that the memo cannot be changed, which means returning
```
akre54: "can't be changed" -> "is immutable"

"means returning" -> "meaning it's unnecessary to return the memo from the iteratee."

index.html
```diff
@@ -962,8 +1001,6 @@ <h2 id="arrays">Array Functions</h2>
 _.object(['moe', 'larry', 'curly'], [30, 40, 50]);
 =&gt; {moe: 30, larry: 40, curly: 50}
 
-_.object([['moe', 30], ['larry', 40], ['curly', 50]]);
```
akre54: why?

index.html
megawac: I'm having trouble summing up `transform` without a ton of examples (http://stackoverflow.com/a/21536978/1517919). Want to have a go at it ^^

index.html
```diff
@@ -1275,17 +1317,31 @@ <h2 id="objects">Object Functions</h2>
...
       </p>
       <pre>
 _.keys({one: 1, two: 2, three: 3});
 =&gt; ["one", "two", "three"]
 </pre>
 
+      <p id="keysIn">
+        <b class="header">keysIn</b><code>_.keysIn(object)</code>
+        <br />
+        Retrieve all the inheritted and own enumerable properties of an <b>object</b>.
```
akre54: sp: "inheritted" -> "inherited"

"inherited and own enumerable properties" -> "enumerable inherited and own properties"

index.html
```diff
@@ -2044,6 +2177,76 @@ <h2 id="links">Links &amp; Suggested Reading</h2>
...
+          <li>
+            <tt>_.iteratee</tt> no longer exposes setting the <tt>argCount</tt> parameter.
+          </li>
+          
+          <li>
+            Fixed a bug where <tt>_.keys</tt> and other object and collection methods
+            did not iterate several reserved properties in old IE environments.
+          </li>
+          <li>
+            Plenty of simplifications, edge case bug fixes, and changes for improved cross platform compatibility.
```
akre54: "cross platform" -> "cross-platform"

index.html
```diff
@@ -2044,6 +2177,76 @@ <h2 id="links">Links &amp; Suggested Reading</h2>
...
       <h2 id="changelog">Change Log</h2>
 
+      <p id="1.8.0">
+        <b class="header">1.8.0</b> &mdash; <small><i>Jan 23, 2015</i></small> &mdash; <a href="https://github.com/jashkenas/underscore/compare/1.7.0...1.8.0">Diff</a> &mdash; <a href="https://cdn.rawgit.com/jashkenas/underscore/1.8.0/index.html">Docs</a><br />
+        <ul>
+          <li>
+            Added <tt>_.keysIn</tt> which returns the enumerable properties of an object.
+          </li>
+          <li>
+            Reverted change where <tt>_.extend</tt> only iterates own properties, it now uses <tt>_.keysIn</tt>. Added <tt>_.assign</tt> that aligns closely with ECMA6 <tt>Object.assign</tt> and iterates over own properties.
```
akre54: break this line to 80 chars

index.html
akre54: I gotta say I don't really see a whole lot of use for this method. What does it do exactly, and why do we need it? Is it just to duck-type the input and avoid returning the output? What makes it "more flexible" than reduce?

index.html
akre54: What would be confusing?

index.html
megawac: We've discussed why its more flexible than reduce in some cases in the thread it was discussed #1901, it's also usually less verbose. Theres a number of places it will be extremely useful for `Marionette` in cases where there is a contract `{*} => {*} | [*] => [*]` (see my `deepMap` or [`deepPick`](http://codereview.stackexchange.com/a/58279/32103) examples)

index.html
jridgewell: @akre54: Your proposed change to "memo is immutable" is **very** misleading. The memo _is_ mutable, but your return value doesn't affect it.

index.html
jridgewell: > why do we need it

I'm indifferent to the wording, but I'm :+1: x98387483 on `_.transform`. 99% of the times that I use `_.reduce` is to re-key an object, and `return memo` always felt like needless cruft that should just be taken care of.

index.html
akre54: Right so immutable is the wrong word. I definitely don't understand how this is much different from `reduce`. (like sure, you can change the memo. but you shouldn't in most cases).

akre54: > The memo is mutable, but the return value doesn't affect it

Sure. So maybe "Another distinction is that the return value does not replace the `memo`, you simply mutate it in the iteratee" or something.

index.html
```diff
@@ -1593,6 +1778,20 @@ <h2 id="objects">Object Functions</h2>
 =&gt; true
 </pre>
 
+      <p id="isError">
+        <b class="header">isError</b><code>_.isError(object)</code>
+        <br />
+        Returns <i>true</i> if <b>object</b> inherrits from an Error.
```
elreimundo: inherits is misspelled

megawac:
@jashkenas can we wait til tomorrow afternoon for the release. I want to
get to a thing or two but Im busy tonight

jashkenas:
I'm afraid that ship has sailed. On to 1.8.1...

jridgewell:
:cry:

PR #2013, opened 2015-01-20
isEqual initialize stacks only if and when we need them

rndr:
no need for stacks if `a` and `b` can be compared by value

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -1125,6 +1125,11 @@
     }
     // Assume equality for cyclic structures. The algorithm for detecting cyclic
     // structures is adapted from ES 5.1 section 15.12.3, abstract operation `JO`.
+    
+    // Initializing stack of traversed objects.
+    // It's done here since we only need them for objects and arrays comparison.
+    aStack = aStack || [];
+    bStack = bStack || [];
     var length = aStack.length;
     while (length--) {
       // Linear search. Performance is inversely proportional to the number of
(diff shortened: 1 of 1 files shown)


megawac:
:+1:

underscore.js
```diff
@@ -1125,6 +1125,8 @@
     }
     // Assume equality for cyclic structures. The algorithm for detecting cyclic
     // structures is adapted from ES 5.1 section 15.12.3, abstract operation `JO`.
+    aStack || (aStack = []);
```
megawac: `aStack = aStack || [];`

Can you also add a comment. Thanks

jridgewell: This is actually [a](https://github.com/jashkenas/backbone/blob/master/backbone.js#L85-L86) [common](https://github.com/jashkenas/backbone/blob/master/backbone.js#L174-L175) [pattern](https://github.com/jashkenas/backbone/blob/master/backbone.js#L283) [in](https://github.com/jashkenas/backbone/blob/master/backbone.js#L358) [Backbone](https://github.com/jashkenas/backbone/blob/master/backbone.js#L446).

rndr: yeah, that's why i done it this way in first place, and test/index.html passing made me feel like it's ok.

rndr:
any ideas on better comment or it's ok?

megawac:
Yep, looks good - Nit can you add periods and I'll merge, otherwise I can do that when I get to a computer

rndr:
done

megawac:
Thanks

PR #2017, opened 2015-01-23
Add breakIf method to escape from a chain

### LINKED ISSUE #2009: chain breaking
Is it possible to break a chain? For example I have created a breakIfNull mixin, and I would like to stop executing the chain if some condition is met.

megawac:
Any better suggestions on a name?

Fixes #2009 

/cc @ppalladino

P.S. if there is interest in this ill clean up the implementation in a follow up pr

### DIFF
--- a/test/chaining.js
+++ b/test/chaining.js
@@ -85,4 +85,43 @@
     deepEqual(o.filter(function(i) { return i > 2; }).value(), [3, 4]);
   });
 
+  test('breakIf', 4, function() {
+    var call = _.partial(ok, true);
+    var dontCall = _.partial(ok, false);
+    var result = _.chain([1, 2, 3, 4, 5])
+     .breakIf(_.constant(false))
+     .tap(call)
+     .filter(function(val) {
+      return val > 1;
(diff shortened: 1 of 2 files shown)


jdalton:
Seems awkward to noop the chain. Looking at it my mind wants to complete the chain. Chaining is about creating a more natural feel to the syntax (like [].map(...).filter(...)) but this doesn't feel natural to me. If there's a possibility of breaking a chain wouldn't it make more sense to have two chain sequences?

ppalladino:
breakIf seems like a home run as far as naming goes

Florian-R:
> Any better suggestions on a name?

_.exit maybe?

akre54:
Agreed with @jdalton here. This doesn't make it especially clear there's a logic branch being created.

jashkenas:
Yes. I don't think leaving a no-op chain dangling is a nice way to go. Better to stop your chain and start another — or not to chain at all.

megawac:
Agreed, I've been meaning to close this as well

PR #2020, opened 2015-01-25
Precompute indexes of _.partial placeholders

jridgewell:
There's no reason to pay for the placeholder lookups every time.

http://jsperf.com/precompute-partial-placeholder

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -739,11 +739,15 @@
   // as a placeholder, allowing any combination of arguments to be pre-filled.
   _.partial = function(func) {
     var boundArgs = slice.call(arguments, 1);
+    var pIndexes = [];
+    for (var i = 0, length = boundArgs.length; i < length; i++) {
+      if (boundArgs[i] === _) pIndexes.push(i);
+    }
     return function bound() {
       var position = 0;
       var args = boundArgs.slice();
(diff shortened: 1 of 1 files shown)


megawac:
Hmm, I think a superior optimization here would be to manually do the slice call and placeholder lookup

underscore.js
```diff
@@ -739,11 +739,15 @@
   // as a placeholder, allowing any combination of arguments to be pre-filled.
   _.partial = function(func) {
     var boundArgs = slice.call(arguments, 1);
+    var pIndexes = [];
+    for (var i = 0, length = boundArgs.length; i < length; i++) {
+      if (boundArgs[i] === _) pIndexes.push(i);
+    }
     return function bound() {
       var position = 0;
       var args = boundArgs.slice();
```
megawac:
```js
args = _.map(boundArgs, function(val) {
   return val == _ ? arguments[position++] : val;
})
```

I feel that would be a superior optimization

jridgewell:
:clap: Mapping is faster, but only if you do it with an [inline for-loop](https://github.com/jashkenas/underscore/pull/2021): http://jsperf.com/precompute-partial-placeholder/2

PR #2022, opened 2015-01-26
Add check to createAssigner function so that _.defaults can use it (fixes #1984)

### LINKED ISSUE #1984: _.defaults should have for-in fixes
`_.defaults` should have `for-in` fixes like `_.extend`.

jmrog:
This commit expands the functionality of the `createAssigner` internal function so that it can be used by `_.defaults` in essentially the same way that it is already used by `_.extend` and `_.assign`. The expansion allows `createAssigner` to receive a second argument indicating whether or not the function that it returns should only assign properties and their values to a given object if the properties are undefined on that object. If `true` (or a truthy value) is passed as the value of that argument, the returned function does just that -- it only assigns properties to a given object if they are undefined on that object; otherwise, `createAssigner` returns a function that assigns properties whether they exist on the object or not (the default functionality). This modification allows `_.defaults` to enjoy the same fixes related to for-in loops as those enjoyed by `_.extend` and `_.assign`. 

See discussion at #1984 for additional details.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -95,7 +95,7 @@
   };
 
   // An internal function for creating assigner functions.
-  var createAssigner = function(keysFunc) {
+  var createAssigner = function(keysFunc, undefinedOnly) {
     return function(obj) {
       var length = arguments.length;
       if (length < 2 || obj == null) return obj;
@@ -105,7 +105,7 @@
             l = keys.length;
         for (var i = 0; i < l; i++) {
           var key = keys[i];
-          obj[key] = source[key];
+          if (!undefinedOnly || obj[key] === void 0) obj[key] = source[key];
         }
       }
       return obj;
@@ -1050,16 +1050,7 @@
   };
 
   // Fill in a given object with default properties.
-  _.defaults = function(obj) {
-    if (!_.isObject(obj)) return obj;
-    for (var i = 1, length = arguments.length; i < length; i++) {
-      var source = arguments[i];
-      for (var prop in source) {
-        if (obj[prop] === void 0) obj[prop] = source[prop];
-      }
-    }
-    return obj;
-  };
+  _.defaults = createAssigner(_.keysIn, true);
 
   // Creates an object that inherits from the given prototype object.
   // If additional properties are provided then they will be added to the


megawac:
+1 nice second pr!

jmrog:
Thanks, @megawac. 

Quick question for you or anyone else while I'm looking at this code... Do we have any idea why the loop through `arguments` in createAssigner starts at 0 rather than 1? It looks like starting the loop at 0 makes for an unnecessary step where an object's keys/values are just reassigned to that same object (since that object will be at position 0 in the `arguments` object). Is there something I'm missing there?

megawac:
I blame @thejameskyle :p (you're 100% right there @jmrog)

jmrog:
Actually, thinking through it a bit more, I think I'm *not* right about the loop!

When `createAssigner` is called for creating `_.extend`, it is passed `_.keysIn`, which returns all keys accessible on an object (including those higher up in the prototype chain) -- not just "own" keys. That means that the first iteration through the passed-in object itself can assign keys/values that were previously only on the object's prototype to the object itself. So, it does make a difference. The original pull request should therefore remain as is.

jmrog:
@megawac, continuing my previous comment, do you know whether it's the intended behavior to have `_.extend` take the first passed-in object's prototype properties and put them on the object itself, as it does now (per the explanation in my above comment)? It didn't used to do that, but maybe the change was intentional. I'll try to have a look through the history, if you're not sure. Should I open an issue about this?

Here's the "old" version of _.extend in the annotated source -- note that the loop starts at 1 rather than 0. (It also checks for "own" properties, but I think that's the purpose of _.assign now, so we can overlook that aspect of it.)

```javascript
  _.extend = function(obj) {
    if (!_.isObject(obj)) return obj;
    var source, prop;
    for (var i = 1, length = arguments.length; i < length; i++) {
      source = arguments[i];
      for (prop in source) {
        if (hasOwnProperty.call(source, prop)) {
            obj[prop] = source[prop];
        }
      }
    }
    return obj;
  };
```

jmrog:
An example might make things clearer. Is this intended?

```javascript
function Foo(){}

var foo = new Foo();
Foo.prototype.bar = "On the prototype!";
foo.hasOwnProperty('bar');
// => false

_.extend(foo, {});
foo.hasOwnProperty('bar');
// => true
```

megawac:
I got you but want to send a pr with a test case (I completely missed that when you mentioned it)

jmrog:
Sure, can do. It'll have to be tomorrow though, as it's time for me to get some sleep.

megawac:
I think its fine either way tbh, I'd have to sleep on it to decide whether the source should qualify for adding the inherited properties to (my gut says no and just leave them on the proto which would be faster ofc)

jmrog:
I'll open a separate issue about this before I head to bed. Maybe others can chime in about the preferred behavior.

PR #2024, opened 2015-01-26
Fix and test for _.extend (fixes #2023)

### LINKED ISSUE #2023: _.extend converts destination object's prototype properties to "own" properties
I'm trying to figure out (or at least get opinions on) whether this behavior is an undocumented "feature" of `_.extend` or a bug:

```javascript
function Foo(){}

var foo = new Foo();
Foo.prototype.bar = "On the prototype!";
foo.hasOwnProperty('bar');
// => false

_.extend(foo, {});
foo.hasOwnProperty('bar');
// => true
```

Basically, the current version of `_.extend` not only iterates through the properties of passed in "source" objects, but also through the properties of the "destination" object itself, including properties on the prototypes. `_.extend` is intended to include properties from prototypes, AFAIK (if you want only the "own" properties, there's now `_.assign` for that), so that isn't the issue. The issue is whether `_.extend` should be including the properties on the prototype chain for *the original destination object itself*, so that the destination object gets new "own" properties and ends up extended even when the "source" is an empty object.

I'm inclined to think this shouldn't be the case. When the "source" is empty (an empty object), the "destination" shouldn't be changed. You can just supply the prototypes as additional arguments to `_.extend` if you want the prototype properties on the destination object iself.

Thoughts?

Original notice of this issue and some related discussion is here: https://github.com/jashkenas/underscore/pull/2022

jmrog:
`_.extend` was iterating over the original (destination) object, going through its 'in' properties on the prototype chain, and putting all of those properties on the destination object (making them 'own' properties). That shouldn't happen. This fix modifies the `createAssigner` internal function (currently used by `_.extend`, `_.assign`, and `_.defaults`) so that the function it returns does not iterate over the destination object itself at all. Iterating over the original object seems unnecessary for any of the methods using createAssigner.

A test is included that makes sure that `_.extend` does not commit the aforementioned offense anymore.

### DIFF
--- a/test/objects.js
+++ b/test/objects.js
@@ -119,6 +119,8 @@
     var subObj = new F();
     subObj.c = 'd';
     deepEqual(_.extend({}, subObj), {a: 'b', c: 'd'}, 'extend copies all properties from source');
+    _.extend(subObj, {});
+    ok(!subObj.hasOwnProperty('a'), "extend does not convert destination object's 'in' properties to 'own' properties");
 
     try {
       result = {};
--- a/underscore.js
+++ b/underscore.js
@@ -99,7 +99,7 @@
     return function(obj) {
       var length = arguments.length;
       if (length < 2 || obj == null) return obj;
-      for (var index = 0; index < length; index++) {
+      for (var index = 1; index < length; index++) {
         var source = arguments[index],
             keys = keysFunc(source),
             l = keys.length;


megawac:
Great stuff @jmrog

jridgewell:
> [jmrog](https://github.com/jashkenas/underscore/issues/1984#issuecomment-71392479): Meager newb to the underscore.js source here

That [lead](https://github.com/jashkenas/underscore/pull/2022#issuecomment-71415009) to you catching a pretty serious bug that at least 6 others (including me) missed! Awesome job :clap:

jmrog:
Thanks, @jridgewell! :grinning: "Fresh eyes" and all that.

PR #2029, opened 2015-02-06
wrap constants or objects passed to _.times in _.constant

gka:
so you can do stuff like this:

```js
_.times(3, 'Foo') // ['Foo', 'Foo', 'Foo']
```

which can be useful for indenting text, for instance

```js
_.times(level, '  ').join('') + line
```

### DIFF
--- a/test/utility.js
+++ b/test/utility.js
@@ -98,6 +98,9 @@
     deepEqual(_.times(0, _.identity), []);
     deepEqual(_.times(-1, _.identity), []);
     deepEqual(_.times(parseFloat('-Infinity'), _.identity), []);
+    
+    // repeats constants
+    deepEqual([1,1,1], _.times(3, 1), 'repeats constants');
   });
 
   test('mixin', function() {
--- a/underscore.js
+++ b/underscore.js
@@ -1320,7 +1320,7 @@
   // Run a function **n** times.
   _.times = function(n, iteratee, context) {
     var accum = Array(Math.max(0, n));
-    iteratee = optimizeCb(iteratee, context, 1);
+    iteratee = optimizeCb(iteratee, _.isFunction(context) ? context : _.constant(context), 1);
     for (var i = 0; i < n; i++) accum[i] = iteratee(i);
     return accum;
   };
--- a/underscore.js
+++ b/underscore.js
@@ -1320,7 +1320,8 @@
   // Run a function **n** times.
   _.times = function(n, iteratee, context) {
     var accum = Array(Math.max(0, n));
-    iteratee = optimizeCb(iteratee, _.isFunction(context) ? context : _.constant(context), 1);
+    if (!_.isFunction(iteratee)) iteratee = _.constant(iteratee);
+    iteratee = optimizeCb(iteratee, context, 1);
     for (var i = 0; i < n; i++) accum[i] = iteratee(i);
     return accum;
   };


megawac:
Seems reasonable

megawac:
By the way you probably meant to type

```js
iteratee = _.isFunction(iteratee) ? optimizeCb(iteratee, context, 1) :  _.constant(context);
```

gka:
yeah, fixed that in 492834c

jashkenas:
Gregor says merge! ;)

jdalton:
Not sure I dig this, its behavior is different than other string shorthands.

jdalton:
Looks like a misplaced feature for _.repeat and _.fill

jdalton:
Before `_.times(4)` would produce `[0,1,2,3]` but now it would produce `[undefined, undefined, undefined, undefined]`.

jdalton:
What are the thoughts on the case of not providing an iteratee?

michaelficarra:
This was a hug mistake. This breaks parametricity. What if I want three of any `a` but I don't know what `a` is? `_.times(3, a)`? Whoops, `a` was a procedure and now I have 3 undefined values instead. Please revert this.

jdalton:
> Whoops, a was a procedure and now I have 3 undefined values instead.

Can you explain this with a code snippet.

michaelficarra:
```js
function makeThreeForMe(a) {
  return [a, a, a];
}
makeThreeForMe(function(){}); // three functions
```

```js
function makeThreeForMe(a) {
  return _.times(3, a);
}
makeThreeForMe(function(){}); // three undefined values
```

jdalton:
> `makeThreeForMe(function(){}); // three undefined values`

That's been Underscore's behavior. This PR doesn't change that.

michaelficarra:
Yes, it's the new behaviour I'm complaining about. `_.times` now behaves like a repeat function *except* if the value is a function, in which case it behaves like `_.times` used to.

For those unfamiliar with this common problem in jQuery-style JavaScript: `Array.prototype.concat` is a canonical example of this pitfall. Given any non-array value, it will produce a new array with that value on the end. But when the value is an array, each of the values it contains are put on the end of the new array instead. So we end up with the pattern `a.concat([b])` to make sure that, regardless of what `b` is, it will be the final element of the result. Similarly here, the "safe" pattern would be `_.times(n, _.constant(a))`. We shouldn't design interfaces like this.

jdalton:
> Yes, it's the new behaviour I'm complaining about. `_.times` now behaves like a repeat function except if the value is a function, in which case it behaves like `_.times` used to.

Before if you passed a non-function value it would error though so the scenario in your example wasn't possible.

> Array.prototype.concat is a canonical example of this pitfall. Given any non-array value, it will produce a new array with that value on the end. But when the value is an array, each of the values it contains are put on the end of the new array instead. 

I dig that flattening feature.

michaelficarra:
I strongly feel that this feature will be responsible for subtle bugs. We should just have a separate `_.repeat` function.

jdalton:
> I strongly feel that this feature will be responsible for subtle bugs. We should just have a separate `_.repeat` function.

lodash has a `_.repeat` for strings which mimics ES6 (2015) `String#repeat`. It'll probably get a `_.fill` to mimic `Array#fill` too.

jashkenas:
Thanks for looking it over. I'll revert.

michaelficarra:
Should we open an issue for adding an `_.repeat`?

PR #2031, opened 2015-02-07
Remove several unnecessary null checks

megawac:
Where they are handled implicitly by other underscore methods. I'd say this makes the source of a number of these functions quite a bit nicer.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -138,11 +138,10 @@
   // Handles raw objects in addition to array-likes. Treats all
   // sparse array-likes as if they were dense.
   _.each = _.forEach = function(obj, iteratee, context) {
-    if (obj == null) return obj;
     iteratee = optimizeCb(iteratee, context);
-    var i, length = obj.length;
+    var i, length;
     if (isArrayLike(obj)) {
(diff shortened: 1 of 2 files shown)


underscore.js
```diff
@@ -216,9 +213,8 @@
         accumulator = {};
       }
     }
-    if (obj == null) return accumulator;
     iteratee = optimizeCb(iteratee, context, 4);
-    var keys = obj.length !== +obj.length && _.keys(obj),
+    var keys = !isArrayLike(obj) && _.keys(obj),
```
jdalton: rogue `isArrayLike`

jdalton:
I can dig it. I kept similar checks in for the perf arms race but if they were removed here I'd likely remove them too.

underscore.js
megawac: Intentional call, same thing going on in [reduce](https://github.com/jashkenas/underscore/blob/master/underscore.js#L186), map, some and others

jdalton: Ah, I thought it was reverted, guess not. No worries.

PR #2033, opened 2015-02-08
Run test suite on node

megawac:


test/arrays.js
```diff
@@ -1,6 +1,7 @@
 (function() {
+  var _ = typeof require == 'function' ? require('../underscore') : window._;
```
michaelficarra: You can just use `require("../")` so if we change the main file name in `package.json`, we don't have to change it everywhere.

megawac: good point, done

jdalton: So what fixed the `isFinite` issue?

megawac: It was an issue with that cli exposing all the underscore methods to global scope

jdalton: Was there an option to prevent that or did they stop in a newer version?

megawac: Fixed by changing the cli to one that's just a wrapper for the official
qunit

PR #2035, opened 2015-02-09
Update _.range docs to clarify end is exclusive [fix #2002]

### LINKED ISSUE #2002: doc for _.range
http://underscorejs.org/#range
Explanation of 'stop' should make it crystal clear that end value is not included.

Returns a list of integers from start to stop
=>
Returns a list of integers from start up to but not including stop

amZotti:
I have updated the documentation for _.range to clarify explicitly that the end value _.range accepts is exclusive and the start value is inclusive. Attached is a screen shot of the changes:

screen shot 2015-02-08 at 5 27 13 pm


megawac:
Thanks!

amZotti:
You're welcome!

PR #2036, opened 2015-02-09
Update _.map docs to explain guarded iteratees

amZotti:
I am a regular user of Underscore and only after reading through the Underscore source code did I learn about _.map being able to handle guarded functions as iteratees. If I have been unaware of guarded functions while using Underscore regularly then chances are there are many other engineers who are equally unaware of this great functionality.

Below is a screen shot of the annotated _.map function documentation in index.html. No other changes have been made besides this. 

screen shot 2015-02-08 at 7 47 20 pm

### DIFF
--- a/index.html
+++ b/index.html
@@ -471,13 +471,15 @@ _.each({one: 1, two: 2, three: 3}, alert);
         through a transformation function (<b>iteratee</b>). The <tt>iteratee</tt>
         is passed three arguments: the <tt>value</tt>, then the <tt>index</tt>
         (or <tt>key</tt>) of the iteration, and finally a reference to the entire
-        <tt>list</tt>.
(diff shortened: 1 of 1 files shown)


megawac:
Does the documentation currently state which functions are gaurded?

amZotti:
The docco documentation page does. I was going to add a side note using similar CSS styling as alias.span on each method that was guarded. Do you have any advice on how I should make the design for adding the 'guarded' statement to the correct methods? 

I was thinking of either listing in bold next to their function signature "Guarded", or writing in bold and italics "Guarded: true"

megawac:
I'd say the bolded text gaurded would suffice. Huge +1 for this

arianf:
@amZotti  :+1: Great work

PR #2039, opened 2015-02-09
_.max and _.min map arrays correctly [fix #2037]

### LINKED ISSUE #2037: _.max and _.min return Infinity when used with _.map
screen shot 2015-02-08 at 8 27 10 pm

This behavior does not seem intuitive or appropriate. I intend to change these methods so they work correctly with _.map

amZotti:
Currently, mapping over a nested array and using _.max or _.min produces unexpected results. 
infinity

This occurs because when _.max or _.min is passed into _.map, the internal function optimizeCb attempts to account for the fact that _.max and _.min have multiple arguments by trying to pass an index value into them. This creates a misalignment between the parameters being passed into _.max and _.min and the parameters they are actually expecting. _.max and _.min expect an iteratee to be passed in as a second argument, but instead _.map is passing an index. 

The result of this arguments mismatch makes _.min and _.max fail the following conditional which checks to see if an iteratee was passed.
screen shot 2015-02-08 at 9 10 57 pm

I have corrected the above conditional by including a following check to ignore the iteratee if it is a number:
works

This additional conditional makes _.min and _.max produce the expected results with no side effects to their normal behavior. Nothing else has been changed and tests have been added to verify the code I have added is indeed valid and does not disrupt anything.

### DIFF
--- a/test/collections.js
+++ b/test/collections.js
@@ -563,6 +563,8 @@
     equal(3, _.max([1, 2, 3, 'test']), 'Finds correct max in array starting with num and containing a NaN');
     equal(3, _.max(['test', 1, 2, 3]), 'Finds correct max in array starting with NaN');
 
+    equal([3, 6], _.map([[1, 2, 3], [4, 5, 6]], _.max), 'Finds correct max in array when mapping through multiple arrays');
+
     var a = {x: -Infinity};
     var b = {x: -Infinity};
     var iterator = function(o){ return o.x; };
@@ -588,6 +590,8 @@
     equal(Infinity, _.min({}), 'Minimum value of an empty object');
     equal(Infinity, _.min([]), 'Minimum value of an empty array');
     equal(_.min({'a': 'a'}), Infinity, 'Minimum value of a non-numeric collection');
+
+    equal([1, 4], _.map([[1, 2, 3], [4, 5, 6]], _.min), 'Finds correct min in array when mapping through multiple arrays');
 
     var now = new Date(9999999999);
     var then = new Date(0);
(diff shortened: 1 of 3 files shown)


test/collections.js
```diff
@@ -563,6 +563,8 @@
     equal(3, _.max([1, 2, 3, 'test']), 'Finds correct max in array starting with num and containing a NaN');
     equal(3, _.max(['test', 1, 2, 3]), 'Finds correct max in array starting with NaN');
 
+    equal([3, 6], _.map([[1, 2, 3], [4, 5, 6]], _.max), 'Finds correct max in array when mapping through multiple arrays');
```
megawac: You want `deepEqual`

jdalton:
For reference see [isIterateeCall](https://github.com/lodash/lodash/blob/3.1.0-npm/internal/isIterateeCall.js).

test/collections.js
amZotti: I will switch the test to deepEqual, commit, and then include in my new commit message that it does not work when passed an array of objects.

After the commit is merged then we can annotate the index.html to reflect this change. I gathered from reading the contributions guidelines that we should merge changes and then change index.html. I assume we are following that procedure here?

megawac: Yep, thats the way we do it

amZotti: What course of action do I need to take in order to make this content high value enough to be merged?

Maybe I misunderstood something, I am new to contributing to open source; Do I need to refactor my code or my commit history to meet Underscore standards?

megawac: The implementation just needs to be discussed, we'll let you know if any
other changes are necessary

jgonggrijp:
Was fixed in #2211.

PR #2041, opened 2015-02-09
Fixed bug in _.sortedIndex that would put undefined at the start FIXES: https://github.com/jashkenas/underscore/issues/1834

arianf:
Changed _.sortedIndex function to hangle undefined values correctly (FIXES: https://github.com/jashkenas/underscore/issues/1834 )

Previously 

```javascript
  if (iteratee(array[mid]) < value) low = mid + 1; else high = mid;
```

Changed it to:

```javascript
var cur = iteratee(array[mid]);
if ( cur < value || (value === void 0 && cur <= Infinity || _.isNaN(cur))) low = mid + 1; else high = mid;
```

The logic works because Undefined should be placed after the highest number possible + 1.

(iteratee(array[mid]) <= Inifinity) // will always place undefined after the highest number, including Inifinity.

I also added in a check to see if the iteratee(array[mid]) was NaN, so NaN has a chance of stopping it.

```javascript
_.sortBy([undefined, 10, 1000, 0, NaN, NaN]);
// => [ 0, 10, 1000, NaN, NaN, undefined ]

_.sortedIndex([ 0, 10, 1000, NaN, NaN, undefined ], undefined);
// => 5
```

without `|| _.isNaN(cur)`

```javascript
_.sortedIndex([ 0, 10, 1000, NaN, NaN, undefined ], undefined);
// => 3
```

So that how I built 

```javascript
var cur = iteratee(array[mid]);
if ( cur < value || (value === void 0 && cur <= Infinity)) low = mid + 1; else high = mid;
```

Since no numeric value can be greater than Infinity, it will place undefined values to the correct index.

Also added edgeCaseTests, to verify it works:

```javascript
test('sortedIndex conforms to sortBy for undefined (#1834)', function() {
    var sorted = _([{a: NaN}, {a: NaN}, {a: void 0}, {a: 0}, {a: 1}, {a: 2}, {a: 3}])
                  .sortBy('a');
    _.each([0, 1, 2, 3, undefined], function(val) {
      equal(_.sortedIndex(sorted, {a: val}, 'a'), _.findIndex(sorted, {a: val}), 'For val: ' + val);
    });

    sorted = _.sortBy([undefined, 1, undefined, 2]);
    equal(_.sortedIndex(sorted, undefined, _.identity), 2);

    var edgeCaseNumbers = [-Infinity, -Infinity, 0, Number.MAX_VALUE, Infinity, Infinity, undefined, undefined];

[...]

Issue #1882, opened 2014-09-29
_.sortedLastIndex


jdalton:
Underscore has `_.sortedIndex` but this could be expanded to `_.sortedLastIndex` using a shared helper between them both (minimal code increase). This would allow `_.lastIndexOf` to support binary searches as well.

PR #2042, opened 2015-02-09
Fixed bug in _.sortedIndex that put undefined at the start (FIX: #1834)

arianf:
Changed _.sortedIndex function to hangle undefined values correctly (FIXES: https://github.com/jashkenas/underscore/issues/1834 )

Revised previous pull request: https://github.com/jashkenas/underscore/pull/2041

Previously 

``` javascript
if (iteratee(array[mid]) < value) low = mid + 1; else high = mid;
```

Changed it to:

``` javascript
var cur = iteratee(array[mid]);
if ( cur < value || (value === void 0 && cur <= Infinity) || (_.isNaN(value) && (cur <= Infinity || cur === void 0))) low = mid + 1; else high = mid;
```

The logic works because Undefined should be placed after the highest number possible + 1.

``` javascript
(iteratee(array[mid]) <= Inifinity)
// will always place undefined after the highest number, including Inifinity ( `<=` ).
```

Based on @jdalton request in https://github.com/jashkenas/underscore/pull/2041#issuecomment-73538789 I added `NaN` to always be placed after undefined.
Based off of https://github.com/jashkenas/underscore/issues/1768#issuecomment-53900794 

``` javascript
(_.isNaN(value) && (cur <= Infinity || cur === void 0)
```

The code aboves logic checks to see if the value is NaN, if it is go past `Infinity` or `undefined`.

``` javascript

_.sortedIndex([ 0, 10, 1000, undefined, NaN, NaN], undefined);
// => 3

_.sortedIndex([ 0, 10, 1000, undefined, NaN, NaN], NaN);
// => 4
```

So that how I built:

``` javascript
var cur = iteratee(array[mid]);
if ( cur < value || (value === void 0 && cur <= Infinity) || (_.isNaN(value) && (cur <= Infinity || cur === void 0))) low = mid + 1; else high = mid;
```

Since no numeric value can be greater than Infinity, it will place undefined values to the correct index.
And it will place all NaN past `Infinity` or `undefined`.

Also added edgeCaseTests, to verify it works:

``` javascript
test('sortedIndex conforms to sortBy for undefined (#1834)', function() {
    var sorted = _.sortBy([undefined, 1, undefined, 2]);
    equal(_.sortedIndex(sorted, undefined, _.identity), 2);

    var edgeCaseNumbers = [-Infinity, -Infinity, 0, Number.MAX_VALUE, Infinity, Infinity, undefined, undefined, NaN];

    var indexForUndefined = _.sortedIndex(edgeCaseNumbers, undefined);
    equal(indexForUndefined, 6, 'undefined should be inserted at index 6');

    var indexForNegInfinity = _.sortedIndex(edgeCaseNumbers, -Infinity);
    equal(indexForNegInfinity, 0, 'negative infinity should be inserted at index 0');

    var indexForInfinity = _.sortedIndex(edgeCaseNumbers, Infinity);
    equal(indexForInfinity, 4, 'infinity should be inserted at index 4');

    var indexForZero = _.sortedIndex(edgeCaseNumbers, 0);
    equal(indexForZero, 2, '0 should be inserted at index 2');

    var indexForNaN = _.sortedIndex(edgeCaseNumbers, NaN);
    equal(indexForNaN, 8, 'NaN should be inserted at index 8');

    var numbers = [10, 20, 30, 40, 50];

    var indexForUndefinedSimple = _.sortedIndex(numbers, undefined);
    equal(indexForUndefinedSimple, 5, 'undefined should be inserted at index 5');
  });
```

### DIFF
--- a/test/arrays.js
+++ b/test/arrays.js
@@ -141,7 +141,7 @@
     var sorted = _.sortBy([undefined, 1, undefined, 2]);
     equal(_.sortedIndex(sorted, undefined, _.identity), 2);
 
-    var edgeCaseNumbers = [-Infinity, -Infinity, 0, Number.MAX_VALUE, Infinity, Infinity, undefined, undefined, NaN];
+    var edgeCaseNumbers = [-Infinity, -Infinity, 0, 345345345234, Infinity, Infinity, undefined, undefined, NaN];
 
     var indexForUndefined = _.sortedIndex(edgeCaseNumbers, undefined);
     equal(indexForUndefined, 6, 'undefined should be inserted at index 6');
--- a/test/arrays.js
+++ b/test/arrays.js
@@ -137,6 +137,33 @@
     equal(_.sortedIndex(array, 2147483648), 2147483648, 'should work with large indexes');
   });
 
+  test('sortedIndex conforms to sortBy for undefined (#1834)', function() {
+    var sorted = _.sortBy([undefined, 1, undefined, 2]);
+    equal(_.sortedIndex(sorted, undefined, _.identity), 2);
+
+    var edgeCaseNumbers = [-Infinity, -Infinity, 0, Number.MAX_VALUE, Infinity, Infinity, undefined, undefined, NaN];
+
+    var indexForUndefined = _.sortedIndex(edgeCaseNumbers, undefined);
+    equal(indexForUndefined, 6, 'undefined should be inserted at index 6');
+
+    var indexForNegInfinity = _.sortedIndex(edgeCaseNumbers, -Infinity);
+    equal(indexForNegInfinity, 0, 'negative infinity should be inserted at index 0');
+
+    var indexForInfinity = _.sortedIndex(edgeCaseNumbers, Infinity);
+    equal(indexForInfinity, 4, 'infinity should be inserted at index 4');
+
+    var indexForZero = _.sortedIndex(edgeCaseNumbers, 0);
+    equal(indexForZero, 2, '0 should be inserted at index 2');
+
+    var indexForNaN = _.sortedIndex(edgeCaseNumbers, NaN);
+    equal(indexForNaN, 8, 'NaN should be inserted at index 8');
+
+    var numbers = [10, 20, 30, 40, 50];
+
+    var indexForUndefinedSimple = _.sortedIndex(numbers, undefined);
+    equal(indexForUndefinedSimple, 5, 'undefined should be inserted at index 5');
+  });
+
   test('uniq', function() {
     var list = [1, 2, 1, 3, 1, 4];
     deepEqual(_.uniq(list), [1, 2, 3, 4], 'can find the unique values of an unsorted array');
--- a/underscore.js
+++ b/underscore.js
@@ -685,7 +685,8 @@
     var low = 0, high = array.length;
     while (low < high) {
       var mid = Math.floor((low + high) / 2);
-      if (iteratee(array[mid]) < value) low = mid + 1; else high = mid;
+      var cur = iteratee(array[mid]);
+      if ( cur < value || (value === void 0 && cur <= Infinity) || (_.isNaN(value) && (cur <= Infinity || cur === void 0))) low = mid + 1; else high = mid;
     }
     return low;
   };


test/arrays.js
```diff
@@ -137,6 +137,33 @@
     equal(_.sortedIndex(array, 2147483648), 2147483648, 'should work with large indexes');
   });
 
+  test('sortedIndex conforms to sortBy for undefined (#1834)', function() {
+    var sorted = _.sortBy([undefined, 1, undefined, 2]);
+    equal(_.sortedIndex(sorted, undefined, _.identity), 2);
+
+    var edgeCaseNumbers = [-Infinity, -Infinity, 0, Number.MAX_VALUE, Infinity, Infinity, undefined, undefined, NaN];
```
megawac: Number.MAX_VALUE won't exist in all environments. Other than that we'll have to discuss the implications. May be worth while to add #1882 at the same time

jdalton: `MAX_VALUE` is like ES3 or smth.

arianf: I updated `Number.MAX_VALUE` to be just a really big number. What do you think of the rest of it?

arianf:
@megawac To make it work with #1882 what if I change the signature to include a last argument? ```javascript
_.sortedIndex = function(array, obj, iteratee, context, last)
```
where if last is set to true, it will return the last

File: underscore.js [hunk 2]
```diff
@@ -685,7 +685,8 @@
     var low = 0, high = array.length;
     while (low < high) {
       var mid = Math.floor((low + high) / 2);
-      if (iteratee(array[mid]) < value) low = mid + 1; else high = mid;
+      var cur = iteratee(array[mid]);
+      if ( cur < value || (value === void 0 && cur <= Infinity) || (_.isNaN(value) && (cur <= Infinity || cur === void 0))) low = mid + 1; else high = mid;
```
megawac: I've been pretty busy the past couple weeks (exam time) but I'm playing with this logic now. I've almost got it passing (1 failing test) with this

```js
if ( cur < value || (cur <= Infinity && !(value <= value)) ) low = mid + 1;
else high = mid;
```

test/arrays.js
```diff
@@ -137,6 +137,33 @@
     equal(_.sortedIndex(array, 2147483648), 2147483648, 'should work with large indexes');
   });
 
+  test('sortedIndex conforms to sortBy for undefined (#1834)', function() {
+    var sorted = _.sortBy([undefined, 1, undefined, 2]);
+    equal(_.sortedIndex(sorted, undefined, _.identity), 2);
+
+    var edgeCaseNumbers = [-Infinity, -Infinity, 0, 345345345234, Infinity, Infinity, undefined, undefined, NaN];
```
megawac: Thats slightly off of what `_.sortBy` would produce

```js
[-Infinity, -Infinity, 0, 345345345234, Infinity, Infinity, NaN, undefined, undefined];
```

megawac: With that fix my suggestion passes https://github.com/megawac/underscore/commit/45e7a4c78236fa98ece10422a51ca1754409f2a7

Feel free to cherry-pick that commit on to this branch

jdalton: Naw `NaN` should be after `undefined`. See https://github.com/jashkenas/underscore/pull/2041#issuecomment-73538789.

megawac: I don't think it matters what order they go in, the most important thing is `_.sortBy` and `_.sortedIndex` are consistent with one another for the values

jdalton: We've been through this before. The comment I linked to is your own comment.

megawac: Yeah, I know thats how I handled it in my implementation, but I don't think theres a particularly good argument for placing them one way or another, just that they should be consistently handled the same between the two methods, unless you have one.

Anyway, doing that would require changes to `_.sortBy` handling of the two values for likely minimal gain

jdalton: `_.sortBy` would still need to be modified to handle the values consistently:
```js
_.sortBy([1, 2, NaN, {}, NaN, undefined]);
// => [1, 2, NaN, {}, NaN, undefined]
```

and to add to that there's consistency with lodash which sorts `NaN` at the end (as per earlier discussions)

megawac:
Hows this look @jdalton https://github.com/jashkenas/underscore/compare/master...megawac:pr/2042?expand=1

PR #2043, opened 2015-02-09
Add "Guarded" on _.map iteratee's [fix #2036]

amZotti:
Certain underscore methods can be used as iteratee's to _.map. These
methods can be referred to as 'Guarded' methods. In pull request #2036 I
have requested that this functionality be included in the documentation
for _.map on index.html. I am following up with that pull request with
this one, which includes documentation that states exactly which methods
are guarded methods.

I am attempting to inform other Underscore developers of this great
functionality by marking appropriate functions as "Guarded" within
index.html. This "Guarded" statement uses the same CSS stylings as
span.alias.

Nothing has been changed besides the appropriate documentation shifts
described above. Screenshots will be included in pull request.

screen shot 2015-02-09 at 11 10 47 am
screen shot 2015-02-09 at 11 11 23 am
screen shot 2015-02-09 at 11 11 34 am

### DIFF
--- a/index.html
+++ b/index.html
@@ -471,7 +471,8 @@ _.each({one: 1, two: 2, three: 3}, alert);
         through a transformation function (<b>iteratee</b>). The <tt>iteratee</tt>
         is passed three arguments: the <tt>value</tt>, then the <tt>index</tt>
         (or <tt>key</tt>) of the iteration, and finally a reference to the entire
-        <tt>list</tt>. Selective <b>Underscore</b> functions marked as <tt>guarded</tt> can be used as an <tt>iteratee</tt> to <b>_.map</b>.
+        <tt>list</tt>. Selective <b>Underscore</b> functions marked as <tt>guarded</tt>
+        can be used as an <tt>iteratee</tt> to <b>_.map</b>.
       </p>
       <pre>
 _.map([1, 2, 3], function(num){ return num * 3; });
--- a/index.html
+++ b/index.html
@@ -471,13 +471,15 @@ _.each({one: 1, two: 2, three: 3}, alert);
         through a transformation function (<b>iteratee</b>). The <tt>iteratee</tt>
         is passed three arguments: the <tt>value</tt>, then the <tt>index</tt>
         (or <tt>key</tt>) of the iteration, and finally a reference to the entire
-        <tt>list</tt>.
+        <tt>list</tt>. Selective <b>Underscore</b> functions marked as <tt>guarded</tt> can be used as an <tt>iteratee</tt> to <b>_.map</b>.
       </p>
       <pre>
 _.map([1, 2, 3], function(num){ return num * 3; });
 =&gt; [3, 6, 9]
 _.map({one: 1, two: 2, three: 3}, function(num, key){ return num * 3; });
-=&gt; [3, 6, 9]</pre>
+=&gt; [3, 6, 9]
+_.map([[1, 2], [3, 4]], _.first);
+=&gt; [1, 3]</pre>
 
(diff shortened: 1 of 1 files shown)


megawac:
This is sweet! Thanks a bunch @amZotti

index.html
```diff
@@ -471,13 +471,15 @@ <h2 id="collections">Collection Functions (Arrays or Objects)</h2>
         through a transformation function (<b>iteratee</b>). The <tt>iteratee</tt>
         is passed three arguments: the <tt>value</tt>, then the <tt>index</tt>
         (or <tt>key</tt>) of the iteration, and finally a reference to the entire
-        <tt>list</tt>.
+        <tt>list</tt>. Selective <b>Underscore</b> functions marked as <tt>guarded</tt> can be used as an <tt>iteratee</tt> to <b>_.map</b>.
```
megawac: Mind putting this on two lines

amZotti: Of course, and thanks for getting back to me so quickly!

index.html
```diff
@@ -471,13 +471,16 @@ <h2 id="collections">Collection Functions (Arrays or Objects)</h2>
         through a transformation function (<b>iteratee</b>). The <tt>iteratee</tt>
         is passed three arguments: the <tt>value</tt>, then the <tt>index</tt>
         (or <tt>key</tt>) of the iteration, and finally a reference to the entire
-        <tt>list</tt>.
+        <tt>list</tt>. Selective <b>Underscore</b> functions marked as <tt>guarded</tt>
+        can be used as an <tt>iteratee</tt> to <b>_.map</b>.
```
jdalton: It's not just `_.map`, but `_.filter`, `_.every`, and others.

amZotti: @jdalton The annotated source code makes it explicit that these functions are guarded specifically for _.map. There is no mention anywhere of _.filter, _.every, or any others, being used with guarded functions.

You can pass guarded functions into methods like _.filter, but it will not produce the expected results:
screen shot 2015-02-09 at 11 54 18 am

jdalton: There is nothing in the guard that prevents it from working with any other method like `_.map` that triggers it. For example:
```js
_.filter([[0,1,2],[3,4,5]], _.first);
// => [[3,4,5]]
```

amZotti: Yeah, I see your point. But I don't think the example you are referencing is the type of behavior I would expect. The fact that guarded methods can be passed into functions like `_.filter` is more of a side effect rather than something which was explicitly implemented. Hence, why the annotated source explicitly states guarded functions work with `_.map`.

jdalton: Imagine Underscore get's a `_.trim` or other like method. `_.filter(arrayOfStrings, _.trim)`. It's better to be clear about the guard rather than boxing in. You could do this by simply saying "used as an `iteratee` to `_.map` and the like."

amZotti: Where do you think we should put that statement? In index.html or the annotated source?

megawac: I think this is fine, we can move it around later. Thanks for the awesome documentation @amZotti

amZotti: You're welcome!

jashkenas:
No, this isn't very awesome, I'm afraid — although I appreciate the effort.

"Guarded" is a weird word to use for this, and won't be clear at first read.  It would be better for us to remove this little flag, and simply demonstrate a useful example of using the documented function as an iteratee for each function where that applies.

megawac:
@jashkenas I think the explanation can defintely be improved (e.g. a section and each *gaurded* linking to it), but the documentation should be included for what is gaurded and isn't

PR #2044, opened 2015-02-09
_.matches only validates own properties[fix #1986]

### LINKED ISSUE #1986: _.matches and _.isEqual behavior
Thoughts on aligning `_.matches` behavior closer to `_.isEqual` by matching only own properties of objects? This would allow easier adoption in the future for partial deep equality comparisons.

amZotti:
I have changed `_.matches` to only match properties which are declared
directly on the object being passed to the predicate function generated
by `_.matches`.

I have accomplished this by testing the property on the object passed in
using the method `hasOwnProperty` method instead of the `in` operator.

I have included specs to demonstrate how this functionality works and
when it should return `true` or `false`.

### DIFF
--- a/index.html
+++ b/index.html
@@ -471,13 +471,15 @@ _.each({one: 1, two: 2, three: 3}, alert);
         through a transformation function (<b>iteratee</b>). The <tt>iteratee</tt>
         is passed three arguments: the <tt>value</tt>, then the <tt>index</tt>
         (or <tt>key</tt>) of the iteration, and finally a reference to the entire
-        <tt>list</tt>.
+        <tt>list</tt>. Selective <b>Underscore</b> functions marked as <tt>guarded</tt> can be used as an <tt>iteratee</tt> to <b>_.map</b>.
       </p>
       <pre>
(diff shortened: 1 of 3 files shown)


megawac:
I'm -1 for this, it's pretty common in my experience for variables to be set on the prototype that I would expect to be found by `matches`

Here's a shotty example (as I'm in the middle of writing a report and can't think of something more appropriate)

```js
var Cla = Backbone.Model.extend({
   idAttribute: "foo"
})

var c = new Cla();
_.matches(c, {
  idAttribute: 'foo'
})
```

I would expect that to yield true not false

jdalton:
I think that having a consistent behavior with comparisons (`matches` & `isEqual`) is a good thing and allows for expansion into partial deep comparisons. The addition of #2045 would handle your case in a much more efficient way. This also keeps a clear distinction between `property` style methods like `pluck`, `omit`, `pick` and comparison methods like `matches` and `isEqual`.

jridgewell:
Closing with [#1986](https://github.com/jashkenas/underscore/issues/1986#issuecomment-75144504).

Issue #1583, opened 2014-04-16
_.intersection should work with arrays regardless of if the first arg is an array


jdalton:
Currently `_.intersection(null, [1, 2, 3]);` returns an empty array `[]` but it should return `[1, 2, 3]` just as `_.intersection(null, [1, 2], null, [2, 3])` should return `[2]`.

This would allign with the current behavior of `_.union` which `_.union(null, [1, 2], null, [3]);` returns `[1, 2, 3]`.

jashkenas:
The intersection of nothing and something is nothing. The union of nothing and something is something. I think the current behavior is the correct behavior.

PR #2050, opened 2015-02-14
_.intersection should work with if the first arg is null.

chrishilditch:
Patch for issue #1583 - _.intersection should work with arrays regardless of if the first arg is an array.

### DIFF
--- a/test/arrays.js
+++ b/test/arrays.js
@@ -204,11 +204,13 @@
     result = _.intersection([2, 4, 3, 1], [1, 2, 3]);
     deepEqual(result, [2, 3, 1], 'preserves order of first array');
     result = _.intersection(null, [1, 2, 3]);
-    equal(Object.prototype.toString.call(result), '[object Array]', 'returns an empty array when passed null as first argument');
-    equal(result.length, 0, 'returns an empty array when passed null as first argument');
-    result = _.intersection([1, 2, 3], null);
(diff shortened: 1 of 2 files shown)


test/arrays.js
```diff
@@ -204,11 +204,13 @@
...
-    result = _.intersection([1, 2, 3], null);
-    equal(Object.prototype.toString.call(result), '[object Array]', 'returns an empty array when passed null as argument beyond the first');
-    equal(result.length, 0, 'returns an empty array when passed null as argument beyond the first');
+    deepEqual(result, [1, 2, 3], 'ignores a null value when it is the first item passed');
+    result = _.intersection([1, 2], null, [2, 3])
+    deepEqual(result, [2], 'ignores null values when they are passed as arguments beyond the first');
+    result = _.intersection([1, 2], [2, 3], null)
+    deepEqual(result, [2], 'ignores a null value when it is passed as the last argument');
+    result = _.intersection(null, null);
+    deepEqual(result, [], 'only passing null values returns an empty array');
```
megawac: None of the tests cover `null` as the first argument

megawac:
Also, it's not just `null|undefined` that are of issue in #1583 but other objects `{}` etc. One simple way to do this would to just filter the arguments :sleepy:

jridgewell:
Closed by https://github.com/jashkenas/underscore/issues/1583#issuecomment-75154103.

PR #2054, opened 2015-02-17
Refactor tests; don't install phantom by default; karma for browser tests

megawac:
This makes the node tests the default instead of headless mode and uses karma to run the phantomjs/whatever browser tests

Added a package.json script `npm run test-browser` to run phantomjs tests, otherwise just lints and runs the suite in node. `Travis` will still run the tests in phantom + firefox

Work up to #1835 (hopefully)


jashkenas:
Can you explain what this gives us?

megawac:
Makes installing much faster as phantomjs is no longer installed. Also lets us delete that vendor/phantomjs script that has gotten dated now

The main perk is karma integration which will make it easier to set up sauce in a follow up pr

jdalton:
Any reason to test in Firefox 10?

megawac:
No particularly good one, just did that to prove it works. Going to be trying to get sauce working hopefully, so it can be removed soon

package.json
```diff
@@ -19,15 +19,16 @@
...
     "uglify-js": "2.4.x"
   },
   "scripts": {
-    "test": "npm run test-node && npm run test-browser && npm run lint",
-    "lint": "eslint underscore.js test/*.js test/vendor/runner.js",
+    "test": "npm run test-node && npm run lint",
+    "lint": "eslint underscore.js test/*.js",
     "test-node": "qunit-cli test/*.js",
-    "test-browser": "phantomjs test/vendor/runner.js test/index.html?noglobals=true",
+    "test-browser": "npm i karma-phantomjs-launcher && ./node_modules/karma/bin/karma start",
```
michaelficarra: Install every time? Why not add a `devDependency`?

michaelficarra: Also, no need to `./node_modules/...`. You should be able to just `karma start` as long as the package marks them as binaries.

megawac: npm won't install it every time, it will check compat. Also this avoids having to install phantomjs every time someone npm installs

And karma doesn't add its cli to bin

megawac: I would rather have users test on FF and add a devDependency on `karma-firefox-launcher` as it doesn't require a big install

PR #2056, opened 2015-02-19
replaced paragraph symbols with section symbols

chharvey:
This pull request will make the following replacements:

old | new
--- | ---
&#x00b6; | &#x00a7;
`&#182;` | `&#167;`
`.pilcrow` | `.ss`
`.pilwrap` | `.sswrap`
`class="pilcrow"` | `class="ss"`
`class="pilwrap "` | `class="sswrap "`

Note: Should use Unicode hex instead of decimal: `&#182;` should be `&#x00b6;` and `&#167;`should be `&#x00a7;`, but that is another issue.

### DIFF
--- a/docs/docco.css
+++ b/docs/docco.css
@@ -210,7 +210,7 @@ ul.sections > li > div {
 
 /*---------------------- Low resolutions (> 320px) ---------------------*/
 @media only screen and (min-width: 320px) {
-  .pilwrap { display: none; }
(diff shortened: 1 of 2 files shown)


michaelficarra:
:+1:

akre54:
Wrong repository. Try docco.

jashkenas:
Yes, but it's a nice idea. Please do send it.

Issue #2008, opened 2015-01-14
Feature request: _.matches accepting an array as the argument, or multiple arguments


afanasy:
So it's possible to filter by multiple values of the same key:

```javascript
_.matches([{name: 'Bob'}, {name: 'Alice'}])
//matches both {name: 'Bob'} and {name: 'Alice'}

//and this syntax would be nice to have too :)
_.matches({name: 'Bob'}, {name: 'Alice'})
//matches both {name: 'Bob'} and {name: 'Alice'}
```

megawac:
I would sooner have a `anyPass` and `allPass` which call each function in an array with provided arguments

Then we could do
```js
var isBobOrAlice = _.partial(_.anyPass, _.map([{name: 'Bob'}, {name: 'Alice'}], _.matches));

isBobOrAlice({name: 'Bob'})
```

afanasy:
This looks good too, but I wanted to use _.where in the same manner:
```javascript
_.where(person, [{name: 'Bob'}, {name: 'Alice'}])
```
returns array with Bob and Alice
and the new syntax is backward compatible as far as I can see, it just adds the special handling case if argument is an array

jashkenas:
Yeah, `anyPass` and `allPass` would be better than this. If you want boolean logic, just use it!

    _.matches(a) or _.matches(b)

    _.matches(a) and _.matches(b) and not _.matches(c)

Issue #1985, opened 2014-12-19
_.isMatch


jdalton:
There's been a request for a function like `_.matches` that accepts the object to compare to without having to create the function. So kind-of like `_.isEqual`. Do you all have thoughts on an  `_.isMatch` method that accepts an object and a source to compare it against?

```js
var object = {'a':1,'b':2,'c':3};
var source = {'b':2};
_.isMatch(object, source);
// => true
```

jashkenas:
-> #2057

PR #2057, opened 2015-02-19
Add proper Object function matches

jridgewell:
This takes `_.matches` and turns it into a proper Object Function (takes the object as the first parameter). This should satisfy the performance problems seen in [jashkenas/backbone#3487](https://github.com/jashkenas/backbone/pull/3487#issuecomment-75001423).

I think it’s pretty apparent we jumbled the arguments originally. See the need to special case [`Backbone.Model#matches`](https://github.com/jashkenas/backbone/blob/master/backbone.js#L390-L393) instead of auto proxying like the other [`_` methods](https://github.com/jashkenas/backbone/blob/master/backbone.js#L669-L680). We even have to special case [`Collection#matches`](https://github.com/jashkenas/backbone/blob/master/backbone.js#L917-L924) so we don’t kill performance doing simple `where`s.

The legacy `_.matches(obj)` predicate matcher has been extracted into `_.matcher`. `_.matches` will return the predicate matcher if only one argument is provided. This legacy support should be removed in v2.0.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -87,7 +87,7 @@
   var cb = function(value, context, argCount) {
     if (value == null) return _.identity;
     if (_.isFunction(value)) return optimizeCb(value, context, argCount);
-    if (_.isObject(value)) return _.matches(value);
+    if (_.isObject(value)) return _.matcher(value);
     return _.property(value);
   };
   _.iteratee = function(value, context) {
@@ -300,13 +300,13 @@
   // Convenience version of a common use case of `filter`: selecting only objects
   // containing specific `key:value` pairs.
   _.where = function(obj, attrs) {
-    return _.filter(obj, _.matches(attrs));
+    return _.filter(obj, _.matcher(attrs));
   };
 
   // Convenience version of a common use case of `find`: getting the first object
   // containing specific `key:value` pairs.
   _.findWhere = function(obj, attrs) {
-    return _.find(obj, _.matches(attrs));
+    return _.find(obj, _.matcher(attrs));
   };
 
   // Return the maximum element (or element-based computation).
@@ -1091,6 +1091,20 @@
     return obj;
   };
 
+  // Returns whether an object has a given set of `key:value` pairs.
+  _.matches = function(object, attrs) {
+    if (arguments.length < 2) return _.matcher(object);
+    var keys = _.keys(attrs), length = keys.length;
+    if (object == null) return !length;
+    var obj = Object(object);
+    for (var i = 0; i < length; i++) {
+      var key = keys[i];
+      if (attrs[key] !== obj[key] || !(key in obj)) return false;
+    }
+    return true;
+  };
+
+
   // Internal recursive comparison function for `isEqual`.
   var eq = function(a, b, aStack, bStack) {
     // Identical objects are equal. `0 === -0`, but they aren't identical.
@@ -1306,16 +1320,10 @@
   };
 
   // Returns a predicate for checking whether an object has a given set of `key:value` pairs.
-  _.matches = function(attrs) {
-    var pairs = _.pairs(attrs), length = pairs.length;
+  _.matcher = function(attrs) {
+    attrs = _.assign({}, attrs);
     return function(obj) {
-      if (obj == null) return !length;
-      obj = new Object(obj);
-      for (var i = 0; i < length; i++) {
-        var pair = pairs[i], key = pair[0];
-        if (pair[1] !== obj[key] || !(key in obj)) return false;
-      }
-      return true;
+      return _.matches(obj, attrs);
     };
   };
 
--- a/test/objects.js
+++ b/test/objects.js
@@ -767,6 +767,103 @@
   test('matches', function() {
     var moe = {name: 'Moe Howard', hair: true};
     var curly = {name: 'Curly Howard', hair: false};
+
+    equal(_.matches(moe, {hair: true}), true, 'Returns a boolean');
+    equal(_.matches(curly, {hair: true}), false, 'Returns a boolean');
+
+    equal(_.matches(5, {__x__: undefined}), false, 'can match undefined props on primitives');
+    equal(_.matches({__x__: undefined}, {__x__: undefined}), true, 'can match undefined props');
+
+    equal(_.matches(null, {}), true, 'Empty spec called with null object returns true');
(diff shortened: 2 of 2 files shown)


underscore.js
```diff
@@ -1091,6 +1091,20 @@
     return obj;
   };
 
+  // Returns whether an object has a given set of `key:value` pairs.
+  _.matches = function(object, attrs) {
+    if (arguments.length < 2) return _.matcher(object);
+    var keys = _.keys(attrs), length = keys.length;
+    if (object == null) return !length;
+    var obj = Object(object);
```
jridgewell: Note: I'm intentionally **not** assigning to `object` here to avoid an [arguments deopt](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#31-reassigning-a-defined-parameter-while-also-mentioning-arguments-in-the-body-typical-example).

jdalton:
lodash has [_.isMatch](https://lodash.com/docs#isMatch) it'd be nice to align to that.

jridgewell:
Sorry, I meant to link https://github.com/jashkenas/underscore/issues/1985 as well.

I overloaded `_.matches` because the current functionality isn't implied by the name (I think `_.matcher` fits it better). I'm not tied to it, and I think getting this (or `_.isMatch`) merged is more important than my dislike of the current name.

jashkenas:
Very nice.

`_.matches` and `_.matcher` are nice names for it.

jdalton:
Way to be a team player.

jdalton:
`_.matches` already exists it seems odd to change the behavior now. And most methods in Underscore that return a boolean are prefixed with `is`.

megawac:
Yeah, I agree with @jdalton I think it makes sense for this to be `matches` (old behaviour) and `isMatch` for bc

arthurvr:
> Way to be a team player.

Ya :(

jridgewell:
> _.matches already exists. It seems odd to change the behavior now.

But it would make the Backbone method ugly. `Model#isMatch` isn't as nice as `Model#matches`.

> And most methods in Underscore that return a boolean are prefixed with is.

Could be aliased?

> This will actually break some suggestions I've made on SO and the issues as matches is not gaurded.

We can add a guard here, and remove that with v2.0 as well.

- - -

I just **really** dislike having `_.matches` return a function... :smile:

megawac:
I feel like that is more effort than its worth: I'd say just leave matches be

jdalton:
> I just really dislike having `_.matches` return a function... :smile:

I've gotten used to it. It makes sense since it produces a function that checks if an object matches the source. This feel inconsistent since there's `isEqual` and not `equals`.

jdalton:
And again the overloading is gross, sometimes returning a function or a boolean is yuck, @michaelficarra is rolling over in his seat.

jridgewell:
@megawac: how were you using `_.matches` as an `iteratee`? `_.filter(arr, _.matches)` doesn't make sense.

@jashkenas: Thoughts on overloading?

megawac:
@jridgewell it was something like


```js
var obj = {a: 1};
var hasMatch = _.chain([{b: 1}, 'a']).map(_.iteratee).any((matcher) => matcher(obj));
```

I don't recall exactly what the code was

jashkenas:
`_.matches` was a crappy name for what it was doing — it shouldn't have been added with that name in the first place. Mea culpa.

    _.matches(options, {async: true})    // => true

Reads nicely. 

The overloading is certainly gross, but can be removed over time. Leaving the function named poorly would be worse in the long term.

megawac:
@jashkenas I disagree, a poorly named function is much better than changing the contract. This will force some users using `matches` to update their code for no particularly good reason

jdalton:
> `_.matches` was a crappy name for what it was doing 

Naming is hard and it can be seen to work as both. `_.isMatch` better follows the naming conventions in Underscore.

jdalton:
> 
```js
_.matches(options, {async: true})    // => true
```
> Reads nicely.

I'd argue that it doesn't read nicely as devs are used to it having completely different behavior.

braddunbar:
Yea, returning a function or a boolean with an optional first argument, not second, is about an confusing as it gets. If we dislike the name `_.matches` then let's alias and deprecate it. I don't think a bad name is a good reason to add a breaking change with completely new functionality.

jridgewell:
> I don't think a bad name is a good reason to add a breaking change with completely new functionality.

I don't think any breaking change was introduced. @megawac mentioned guarded behavior but you would never use `_.matcher` as an iteratee, only it's returned method. I also searched his [StackOverflow answers](http://stackoverflow.com/search?q=user%3A1517919+%5Bunderscore%5D+matches) (sorry that's kind of stalker-y) and only found one answer pertaining to `_.matches`, but it isn't affected by this. And if this is the concern, we can add a `guard` flag until the legacy support is removed.

The optional first arg is clunky, but it'll go away in the next major (hopefully) release.

megawac:
@jridgewell here's a (currently non working) example of something that would break with this

> ```js
var isBobOrAlice = _.partial(_.anyPass, _.map([{name: 'Bob'}, {name: 'Alice'}], _.matches));
isBobOrAlice({name: 'Bob'})
```

https://github.com/jashkenas/underscore/issues/2008#issuecomment-70552484

jridgewell:
Damn. I'll work a `guard` PR.

megawac:
> Yea, returning a function or a boolean with an optional first argument, not second, is about an confusing as it gets. If we dislike the name _.matches then let's alias and deprecate it. I don't think a bad name is a good reason to add a breaking change with completely new functionality.

:+1:

The current way this is being implemented wastes LOC for a name. Just implement `_.isMatch` instead :sleepy:

jashkenas:
Alright then — `isMatch`, and `matcher`, with `matches` aliased and deprecated, and removed from the docs. 

@jridgewell — you got this?

jridgewell:
Yah, I'll get a new PR tonight.

Can we get a 2.0 milestone with `_.isMatch` aliased to `_.matches` (and possibly `_.isEqual` -> `_.equals`)?

jashkenas:
Ah, I was thinking about sooner than that ;) I'll get it.

Re: 2.0 — nah. A little naming jiggery-pokery isn't a good enough reason to start rolling that ball down the hill.

flyswatter:
> Model#isMatch isn't as nice as Model#matches.

How about pick the grammatically correct `isMatching`?

PR #2065, opened 2015-02-20
Refine _.pick

jridgewell:
Not sure how this’ll do performance wise. I’ll write a jsperf tomorrow.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -988,23 +988,20 @@
 
   // Return a copy of the object only containing the whitelisted properties.
   _.pick = function(obj, iteratee, context) {
-    var result = {}, key;
+    var result = {}, keys;
     if (obj == null) return result;
     if (_.isFunction(iteratee)) {
+      keys = _.allKeys(obj);
       iteratee = optimizeCb(iteratee, context);
-      var keys = _.allKeys(obj);
-      for (var i = 0; i < keys.length; i++) {
-        var key = keys[i];
-        var value = obj[key];
-        if (iteratee(value, key, obj)) result[key] = value;
-      }
     } else {
-      var keys = flatten(arguments, false, false, 1);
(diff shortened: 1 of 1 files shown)


megawac:
Any advantage to this refactor? Did `_.pick({constructor: 1}, 'constructor')` not work or something?

jridgewell:
It was a lot of duplicated code between the two branches to support an inline `key in obj`.

https://github.com/jashkenas/underscore/commit/173a7fa817e17b3cc5d555180ba0ea4d35655172 fixed issues with adding new keys as you're iterating, but it created linting errors (duplicate `var` declarations).

- - -

I think this wraps up all the removals changes from 1.8 to 1.8.1. I can't spell anymore, time for bed. :sleeping:

megawac:
I tried a similar refactor before for reference #1639

jridgewell:
That's probably because of [Line 903](https://github.com/jashkenas/underscore/pull/1639/files#diff-0f36b362a0b81d6f4d4bfd8a7413c75dR903). The only perf hit this should take is invoking (and this is sure to be inlined) in the `_.pick(obj, 'key', 'key2')` case.

```javascript
var iteratee = function(value, key, object) {
  return key in object;
};
```

megawac:
Can you post benchmarks when you get a chance

jridgewell:
[jsperf](http://jsperf.com/underscore-1639/6).

Unfortunately, it's a bit slower for the `_.pick(obj, 'key')` case. I was able to improve performance a bit in an [opt branch](https://github.com/jridgewell/underscore/compare/refine-pick).

jashkenas:
That's a weird optimization, no?

jridgewell:
Are you talking about https://github.com/jashkenas/underscore/pull/2068?

Issue #2061, opened 2015-02-20
Discuss latest changes to master before releasing


thejameskyle:
There are a lot of changes happening right now that have not been discussed, including breaking changes.

https://github.com/jashkenas/underscore/compare/6d3100ed758818ba41cde2dc178524e24f236b23...master

At the very least these should be pull requests that other contributors can discuss.

Changes include:
- Removing `_.transform`
- Removing `_.create`
- Renaming `_.keysIn` to `_.allKeys`
- Renaming `_.mapValues` to `_.mapObject`
- Renaming `_.assign` to `_.extendOwn`
- Renaming `_.matches` to `_.isMatch`
- Adding `_.matcher` alias `_.matches`
- Reverted some changes in 1.7 (ie. `_.extend`)

thejameskyle:
Well, too late I guess https://twitter.com/jashkenas/status/568564455937839106

jashkenas:
Too late indeed. We can follow up with a 1.8.1 pronto if needed. Hopefully it won't be...

jridgewell:
> Renaming _.matches to _.isMatch
> Adding _.matcher alias _.matches

Those were both fixed. 1.7's `_.matches` is 1.8's `_.matcher = _.matches`. 1.8's `_.isMatch` is what you expect it is.

jridgewell:
> Reverted some changes in 1.7 (ie. _.extend)

1.8 has the same behavior as 1.6, which is [intentional](https://github.com/jashkenas/underscore/pull/1799#issuecomment-53581153) https://github.com/jashkenas/underscore/commit/b17bd671ca49e1b76b43f29cee40352fb4651c9f.

jridgewell:
Do you mind cleaning up the ones that weren't changes made last minute by @jashkenas? `_.matches`, `_.matcher`, `_.isMatch`, and `_.extend` undiscussed issues.

I'd love for this to be the jumping off point to fix 1.8 with 1.8.1. And this time, let's [actually](https://github.com/jashkenas/underscore/issues/1855) fix it.

jridgewell:
Additionally, we have a breaking change to a brand new feature in 1.7: Overriding `_.iteratee` no longer changes the behavior of the collection iterators. Fix is https://github.com/jashkenas/underscore/pull/1965.

@jashkenas Please revoke 1.8 from npm. Let's get these issues worked out, and ship a 1.8.1.

jashkenas:
[...]

PR #2070, opened 2015-02-20
Assign extendOwn

jridgewell:
“assign” is an ES6 standard, and naming should match it. Any dev who remotely follows ES will immediately recognize and understand what it does.

Aliases `_.assign` to `_.extendOwn`, for 1.8.0 compatibility.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1436,8 +1436,9 @@ _.extend({name: 'moe'}, {age: 50});
 =&gt; {name: 'moe', age: 50}
 </pre>
 
-      <p id="extendOwn">
-        <b class="header">extendOwn</b><code>_.extendOwn(destination, *sources)</code>
+      <p id="assign">
+        <b class="header">assign</b><code>_.assign(destination, *sources)</code>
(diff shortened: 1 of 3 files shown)


jashkenas:
Whoa there folks. `extendOwn` is a nicer name alongside `extend`. We can add an `assign` alias — but let's use `extendOwn` as the primary documented name, and internally.

jridgewell:
> let's use extendOwn as the primary documented name, and internally.

I strongly disagree. I'm not sure why we're actively discouraging ES6 compatibility, but keeping the two in sync is an awesome win.

jashkenas:
ES6 is going to very soon — at least for simple things like assign — be viable. If you want to use Object.assign, just use it. If we can improve on it, let's improve on it. One of the important areas where we can improve on things is naming — a more internally consistent Underscore API is nicer than a franken-Under-ES6-score one.

thejameskyle:
The ideal API in my opinion would be for `_.assign` to behave like `Object.assign` and for `_.extend` to be an alias of `_.assign`. I don't see a need for the `_.keysIn` version of an assign method.

PR #2071, opened 2015-02-20
Alias allKeys to keysIn

jridgewell:
I’m only mad because it wasn’t discussed. And, I think `_.keysIn` is a better name, in part because of your [own argument](https://github.com/jashkenas/underscore/commit/4f771e0a04e8b071b4857e8e8da13fcb79c5c0d5#commitcomment-9833268).

### DIFF
--- a/index.html
+++ b/index.html
@@ -1346,6 +1346,7 @@ _.keys({one: 1, two: 2, three: 3});
 
       <p id="allKeys">
         <b class="header">allKeys</b><code>_.allKeys(object)</code>
+        <span class="alias">Alias: <b>keysIn</b></span>
         <br />
         Retrieve <i>all</i> the names of <b>object</b>'s own and inherited properties.
       </p>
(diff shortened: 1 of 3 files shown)


jashkenas:
No — let's leave this be. `ownKeys` would have been preferable, but that's already what `keys` is. Hence `allKeys`.

`keys` vs. `keysIn` — you have to memorize which is which.

megawac:
`allKeys` makes no sense and is inconsistent with the rest of the library. At least `keysIn` aligns with `for..in`

jridgewell:
But you just released `_.extendOwn`?

For the record, I'm :-1: on the change. I've updated to just alias `_.allKeys` to `_.keysIn`.

thejameskyle:
`_.allKeys` is pretty ambiguous IMO, `_.keysIn` makes sense as it's a `for..in` loop that fixes IE issues.

raghumadhavan:
@jridgewell Issue seems to be open from a long time, could you consider closing it.

jgonggrijp:
I'm sorry to meddle in other people's old sores, but I'm just closing this.

PR #2072, opened 2015-02-20
Contains fromIndex

jridgewell:
Aligns with ES6’s `contains`.

Re: https://github.com/jashkenas/underscore/pull/1962


jashkenas:
`fromIndex` is a really strange thing to have in a `contains`. Is there a good use-case for it — or is it only because you get it for free with indexOf? If there's no good reason, why clutter the API?

PR #2073, opened 2015-02-20
Alias mapObject to mapValues

jridgewell:
For the record, I like `_.mapObject` better. But this change was undiscussed.


thejameskyle:
`_.mapValues` makes more sense to me as you are only manipulating the values in the object and not the keys. `_.mapObject` implies a limited set of functionality that `_.transform` has.

paulfalgout:
:+1:

captbaritone:
I like the alias, but the PR should also update tests.

jridgewell:
@captbaritone: You got it. :wink:

PR #2074, opened 2015-02-20
Add _.create method

jridgewell:
Another undiscussed removal.

Re: https://github.com/jashkenas/underscore/pull/1907

### DIFF
--- a/test/objects.js
+++ b/test/objects.js
@@ -282,6 +282,32 @@
     equal(_.clone(null), null, 'non objects should not be changed by clone');
   });
 
+  test('create', function() {
+    var Parent = function() {};
+    Parent.prototype = {foo: function() {}, bar: 2};
+
+    _.each(['foo', null, undefined, 1], function(val) {
+      deepEqual(_.create(val), {}, 'should return empty object when a non-object is provided');
+    });
+
+    ok(_.create([]) instanceof Array, 'should return new instance of array when array is provided');
+
+    var Child = function() {};
+    Child.prototype = _.create(Parent.prototype);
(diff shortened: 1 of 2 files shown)


michaelficarra:
:+1:

jashkenas:
Underscore should not be a generic OOP library for JavaScript — and even if it were, we should add a function to easily set up constructors ... not one that promotes inefficient proliferation of identical prototypes. This is something we should not add to the API.

thejameskyle:
I use this all the time as a more efficient clone that respects the prototype.

megawac:
@jashkenas this could be useful in BB `extend` at a glance

jashkenas:
Yes — it's only a piece of the puzzle. To use it "correctly" ... you still have to monkey around with prototypes yourself.

If we're going to add an inheritance function, shouldn't it be user-friendly, and actually do the whole job for you?

    B = _.subclass(A)

megawac:
> `B = _.subclass(A)`

That wouldn't enable B to set a different constructor. `create` handles the tricky part of setting the prototype to an instance, adding the constructor and adding some properties to the prototype.

Anyway it's mirrors a well understood ES5 interface that is present in a lot of literature, is useful for more than just subclassing, and would be useful in BB as demonstrated

jashkenas:
Alright.

alanhussey:
`_.create` is being shown in the sidebar on [underscorejs.org](http://underscorejs.org/#extendOwn), but there isn't any documentation available for it. Not sure if this is a known issue. screenshot

PR #2075, opened 2015-02-20
Add _.transform method

jridgewell:
Another undiscussed removal.

I’m not so hot on the `Ctor` conditional, or early breaking. But I [**want**](https://github.com/jashkenas/underscore/pull/2012#discussion_r23318181) this method.

Re: https://github.com/jashkenas/underscore/pull/1901

### DIFF
--- a/test/collections.js
+++ b/test/collections.js
@@ -252,6 +252,53 @@
     strictEqual(_.reduceRight, _.foldr, 'alias for reduceRight');
   });
 
+  test('transform', function() {
+    var list = _.transform(['foo', 'bar', 'baz'], function(accumulator, value, index){ accumulator[index] = value + '2'; });
+    deepEqual(list, ['foo2', 'bar2', 'baz2'], 'handles arrays with no accumulator');
+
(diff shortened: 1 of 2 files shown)


megawac:
:+1:

jridgewell:
Whoops, left in the `_.create` commit. Rebased and removed it.

jashkenas:
I think that the API is currently too magical — and not something we're going to want to end up supporting long term.

thejameskyle:
What would make it less magical?

jgonggrijp:
Really, this does not belong in a functional library. The only reason for having this instead of `_.reduce` in the first place is the option to break out of the loop, which in turn requires modifying the accumulator in place, but both of those things are procedural and completely against the conventions of the other collection functions. Besides, I don't see why one would want to transform an object only partially.

"Magical" probably referred to the attempt to conjure an appropriate accumulator out of thin air when none is provided. The `if`/`else` tree to support this will grow out of proportions over time and any attempt to draw a line on that will be arbitrary. Indeed, this has [grown into a cluster of arbitrary special cases](https://github.com/lodash/lodash/blob/74b09d6272354b804dc9fa9f0703c566736cf8f5/transform.js#L37-L53) in the Lodash implementation.

PR #2077, opened 2015-02-21
updated mapObject example text in index.html

91bananas:
Previously had mismatching keys from the example compared to its result.


akre54:
Thanks

PR #2079, opened 2015-02-23
fixed isArrayLike returning true for 0

smelnikov:
before: 

```
_.size(0) //undefined
_.size(1) //0
```

after:

```
_.size(0) //0
_.size(1) //0
```

There are probably other methods that are broken for the value `0` but `_.size` is how I discovered the bug.

### DIFF
--- a/test/collections.js
+++ b/test/collections.js
@@ -789,6 +789,7 @@
     equal(_.size(new String('hello')), 5, 'can compute the size of string object');
 
     equal(_.size(null), 0, 'handles nulls');
+    equal(_.size(0), 0, 'handles numbers');
   });
 
   test('partition', function() {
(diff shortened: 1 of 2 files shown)


jridgewell:
The only other affected method is `_.shuffle`:

```javascript
_.shuffle(0) == [undefined];
_.shuffle(1) == [];
```

megawac:
This test is sufficient and a good catch. Thanks @smelnikov!

jridgewell:
Also found `_.sample` is affected:

```javascript
_.sample(0, 1) == [undefined];
_.sample(1, 1) == []
```

@jashkenas: Mind cutting a 1.8.3 bugfix release?

jashkenas:
Sure — but let's wait for a medium period of time before doing it. (Like, end of the week?) I'd imagine there are a number of minor things that can go in...

jridgewell:
:+1:

PR #2084, opened 2015-02-24
Refactor _.intersection, More readable, Less lines

BSteephenson:
Uses underscore to make _.intersection shorter and more understandable.


jridgewell:
Re: https://github.com/jashkenas/underscore/pull/1517

BSteephenson:
Oops my bad

PR #2089, opened 2015-02-27
small example added to _.extendOwn

eduardojribeiro:


jashkenas:
That doesn't demonstrate the `own` nature of `extendOwn`. We'll need something a little more interesting.

eduardojribeiro:
this demonstrate?
https://github.com/jashkenas/underscore/pull/2091

jridgewell:
He wants something that uses a prototype:

```javascript
var Class = function() {};
Class.prototype.own = false;

var obj = _.extendOwn({}, {own: true}, new Class());
obj.own; // => true
```

eduardojribeiro:
I had not Understood. Thanks.

Issue #1850, opened 2014-09-15
_.sortedIndex problem using >>> 1


jdalton:
`_.sortedIndex` uses `>>> 1` as a shortcut for `Math.floor((low + high) / 2)` but this can cause an infinite loop on large arrays because `>>>` will eventually resolve to `0` when exceeding `Math.pow(2, 32)-1`. For example:

``` js
var values = [0, 1, 3, 7, 15, 31, 63, 127, 255, 511, 1023, 2047, 4095, 8191, 16383, 32767, 65535, 131071, 262143, 524287, 1048575, 2097151, 4194303, 8388607, 16777215, 33554431, 67108863, 134217727, 268435455, 536870911, 1073741823, 2147483647];
var array = Array(Math.pow(2, 32) - 1);
var length = values.length;

while (length--) {
    array[values[length]] = values[length];
}
_.sortedIndex(array, 2147483648); // locks up
```

Also the method should ensure not to return an index greater than `Math.pow(2, 32) - 2`.

PR #2093, opened 2015-03-01
speed up the bisect

xch89820:
speed up the bisect like crossfilter has done


megawac:
This would undo the fix for https://github.com/jashkenas/underscore/issues/1850 and work done in #1954

PR #2094, opened 2015-03-03
Prevent iOS from improperly JIT-ing isArrayLike

### LINKED ISSUE #2081: Wrong number of iterations for objects that contain numeric keys
This is an ios only bug that also happened on lodash. 

https://github.com/lodash/lodash/issues/799

Here is the original testcase updated for underscore: http://jsfiddle.net/rcsz27v1/2/

The result should be 3, but on ios it's 51.

It happens on underscore 1.8.2 as well as 1.7.0.

jridgewell:
Fixes #2081. Special cases with `_.isArray`, since it’s so damn fast and engines [really dislike](http://jsperf.com/ios-jit-isarraylike/5) doing `'length' in array`.

[jsperf](https://jsperf.com/ios-jit-isarraylike/2), and with [longer collections](http://jsperf.com/ios-jit-isarraylike/3).

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -125,9 +125,10 @@
   // Helper for collection methods to determine whether a collection
   // should be iterated as an array or as an object
   // Related: http://people.mozilla.org/~jorendorff/es6-draft.html#sec-tolength
-  var MAX_ARRAY_INDEX = Math.pow(2, 53) - 1;
+  // Avoids a very nasty iOS 8 JIT bug on ARM-64. #2094
+  var MAX_ARRAY_INDEX = Math.pow(2, 53) - 1, LENGTH = 'length';
   var isArrayLike = function(collection) {
-    var length = collection != null && collection.length;
+    var length = collection != null && collection[LENGTH];
     return typeof length == 'number' && length >= 0 && length <= MAX_ARRAY_INDEX;
   };
 


megawac:
Mind adding a failing test case as @osolo describes in #2081 I want to see if sauce catches it

jridgewell:
> Mind adding a failing test case as @osolo describes in #2081 I want to see if sauce catches it

https://github.com/jashkenas/underscore/pull/2088 :stuck_out_tongue:

jridgewell:
The failure has something to do with karma-cli not liking node v0.8. Everything else works fine.

jridgewell:
Also, discussion happening in https://github.com/jridgewell/underscore/commit/8a6728c40d13cfe33f65960ee0dc57c749ca8f5a.

jridgewell:
Boo yah, @jdalton came up with a brilliant `collection['length']` fix that's much faster.

michaelficarra:
Will the minifier not inline this?

jridgewell:
Appears not. Tested with our uglify flow, and it worked fine. Also tested it with closure compiler on crazy `ADVANCED_OPTIMIZATIONS` mode, and it worked as well.

michaelficarra:
Wow, we really need to improve the state of our minifiers. That is sad.

megawac:
I would still be hesitant to use the brace fix for the cases where users
minify underscore themselves. I know closure converts brackets to dot
notation

I would suggest using result(x,length)

jridgewell:
@megawac See my above comment.

> Also tested it with closure compiler on crazy ADVANCED_OPTIMIZATIONS mode, and it worked as well.

I can do another PR that can use `_.property`, but I'm not sure it's necessary.

megawac:
@jridgewell I just tested minifying underscore with closurecompiler.js@1.5.1 and it coerced it to using dot notation

jridgewell:
Agh, I'll submit another PR.

akre54:
I seem to remember that the closure minifier required you to do the quoted keys rain dance to keep property names from getting minified `obj['prop']` vs `obj.prop` and to do proper typechecking but I could be totally wrong.

megawac:
@akre54 yeah, but uglify won't respect it then xD

PR #2099, opened 2015-03-10
Speeding up _.uniq method by using hash table lookups for primitive data...

ch-ms:
... types.

It is now works 200 times faster on my macbook air on an unsorted array with primitive data types.

I used these tests with 1000000 elements to measure perfomance:

function testSpeed(n){
    var tarr = [];

```
for(var i = 0; i<=n; i++){
    tarr.push(Math.floor(Math.random()*10000));
}

var startTime = Date.now();
_.uniq(tarr);

console.log("Ended by:", Date.now()-startTime);
```

}

function testSpeedIter(n){
    var tarr = [];

```
for(var i = 0; i<=n; i++){
    tarr.push(Math.floor(Math.random()*10000));
}

var iter = function(v, i, array){
    return v+"some";
}

var startTime = Date.now();

_.uniq(tarr, iter);

console.log("Ended by:", Date.now()-startTime);
```

}

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -520,19 +520,41 @@
     if (iteratee != null) iteratee = cb(iteratee, context);
     var result = [];
     var seen = [];
+    var seenPrimitive = {};
     for (var i = 0, length = array.length; i < length; i++) {
       var value = array[i],
           computed = iteratee ? iteratee(value, i, array) : value;
+
+      var vType = typeof computed,
+          vKey = vType[0] + computed,
+          isPrimitive = _.contains(['number', 'string', 'boolean', 'undefined'], vType);
+
       if (isSorted) {
         if (!i || seen !== computed) result.push(value);
         seen = computed;
       } else if (iteratee) {
-        if (!_.contains(seen, computed)) {
+
+        if (isPrimitive){
+          if (seenPrimitive[vKey] === undefined){
+            seenPrimitive[vKey] = true;
+            result.push(value);
+          }
+        } else if (!_.contains(seen, computed)){
           seen.push(computed);
           result.push(value);
         }
-      } else if (!_.contains(result, value)) {
-        result.push(value);
+
+      } else {
+
+        if (isPrimitive){
+          if (seenPrimitive[vKey] === undefined){
+            seenPrimitive[vKey] = true;
+            result.push(value);
+          }
+        } else if (!_.contains(result, value)){
+          result.push(value);
+        }
+
       }
     }
     return result;
@@ -1134,7 +1156,7 @@
     }
     // Assume equality for cyclic structures. The algorithm for detecting cyclic
     // structures is adapted from ES 5.1 section 15.12.3, abstract operation `JO`.
-    
+
     // Initializing stack of traversed objects.
     // It's done here since we only need them for objects and arrays comparison.
     aStack = aStack || [];
@@ -1298,7 +1320,7 @@
     };
   };
 
-  // Returns a predicate for checking whether an object has a given set of 
+  // Returns a predicate for checking whether an object has a given set of
   // `key:value` pairs.
   _.matcher = _.matches = function(attrs) {
     attrs = _.extendOwn({}, attrs);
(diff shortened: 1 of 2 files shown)


d10:
You should use [jsPerf](http://jsperf.com) for performance tests.

underscore.js
```diff
@@ -520,19 +520,35 @@
...
+
       if (isSorted) {
         if (!i || seen !== computed) result.push(value);
         seen = computed;
-      } else if (iteratee) {
-        if (!_.contains(seen, computed)) {
-          seen.push(computed);
+      } else {
+        var vType = typeof computed,
+          isPrimitive = _.contains(['number', 'string', 'boolean', 'undefined'], vType);
```
akre54: how about `!_.isObject(computed)`

michaelficarra: That will catch `null`, but will fail for symbols.

ch-ms: @akre54 I have implemented this, thanks for advice!
@michaelficarra I think I can patch isObject function to recognize symbols as primitives. Is it ok for a symbol to be a primitive?

michaelficarra: @ch-ms `_.isObject` already treats symbols as primitives. The problem is that unique symbols don't have a unique toString, so this method would want to either use the symbol directly (dangerous) or treat them as objects.

akre54: I think it's fine to treat it as an object, no? `indexOf` should work.

michaelficarra: Yeah, that would be fine, as I said, but that's not what this code does.

megawac:
How does this compare with using `Set` see https://github.com/jashkenas/underscore/pull/1868#issuecomment-56561228

underscore.js
```diff
@@ -520,19 +520,34 @@
     if (iteratee != null) iteratee = cb(iteratee, context);
     var result = [];
     var seen = [];
+    var seenPrimitive = {};
```
megawac: worth doing `nativeCreate ? nativeCreate(null) : {}`?

underscore.js
ch-ms: @michaelficarra yeah, now I see problem with isObject, rollback to indexOf

underscore.js
ch-ms: @megawac reason?

ch-ms:
@d10 I made a test for integers http://jsperf.com/underscore-uniq-hash-table-lookup

underscore.js
akre54: why not `!_.isObject(computed) && !_.isSymbol(computed)` for the hypothetic future where we have that method? 

Regardless we should play around with various ways of making this method faster.

megawac: > The problem is that unique symbols don't have a unique toString, so this method would want to either use the symbol directly (dangerous) or treat them as objects.

Placing a symbol in an object is pretty safe - you can treat them as primitives and not have to worry about shim implementations being caught due to the `typeof` checks.

akre54: Yeah actually after testing it out...

```js
> a = Symbol()
Symbol()
> b = Symbol()
Symbol()
> obj = {}
{}
> obj[a] = 1
1
> obj[b] = 2
2
> obj[a]
1
```

michaelficarra: I wasn't saying it was unsafe because setting/getting them would not work. I was saying it was unsafe because well-known symbols have behavioural changes.

megawac:
Whats the hit on an array of objects @ch-ms

underscore.js
megawac: Can you provide an example @michaelficarra

underscore.js
megawac: Some browsers used to optimize for it. Doesn't seem necessary any more based on some benchmarks

underscore.js
michaelficarra: @megawac See https://people.mozilla.org/~jorendorff/es6-draft.html#sec-properties-of-the-symbol-constructor. Each of the well-known symbols there changes how the object behaves in some way. It is not safe in a future with additional well-known symbols to modify these fields arbitrarily.

underscore.js
jridgewell: It'll actually be necessary if anything pollutes Object's `prototype`.

Issue #1615, opened 2014-04-29
_.contains does not work for arrays


joaomilho:
So contains does not work for array...

``` js
_.contains([[]], [])
=> false

_.contains([[1]], [1])
=> false
```

Since arrays are broken...

``` js
[] == []
=> false

[] === []
=> false

[1] === [1]
=> false
```

Anyway it's pretty misleading not to work with arrays... any thoughts?

allenan:
It seems like there should be a more natural way to test if an array includes values from a sub-array. If the implementation of includes/contains is not intended to handle this case, what about a `_.containsAll(collection, [value1, value2])` method?

PR #2100, opened 2015-03-12
Add functions _.containsAll and _.containsAny

allenan:
`contains` does not currently work with arrays (https://github.com/jashkenas/underscore/issues/1615) and the current workarounds for this aren't very readable.

This PR adds two methods for testing whether a collection includes all or any values from another collection.

```javascript
_.containsAll([1, 2, 3], [1, 3]); //=> true
_.containsAll([1, 2, 3], [1, 4]); //=> false

_.containsAny([1, 2, 3], [1, 4]); //=> true
_.containsAny([1, 2, 3], [4, 5]); //=> false
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -208,6 +208,8 @@
       <li>- <a href="#every">every</a></li>
       <li>- <a href="#some">some</a></li>
       <li>- <a href="#contains">contains</a></li>
+      <li>- <a href="#containsAll">containsAll</a></li>
+      <li>- <a href="#containsAny">containsAny</a></li>
       <li>- <a href="#invoke">invoke</a></li>
       <li>- <a href="#pluck">pluck</a></li>
       <li>- <a href="#max">max</a></li>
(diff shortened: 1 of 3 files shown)


jridgewell:
What's wrong with just using `_.every` and `_.any`?

```javascript
_.every(subArray, function(i) { return _.contains(array, i); });
_.any(subArray, function(i) { return _.contains(array, i); });
```

allenan:
@jridgewell Thanks for the update to _.contains. I profiled using _.intersection vs _.every + _.partial and found the latter to be much more performant so I went ahead and changed my implementation to use that approach. 

I'll leave this PR open in case anyone feels it would be valuable to have these methods as shortcuts for readability and also to promote a more performant solution for array comparison.

megawac:
Though this can definitely be useful, I believe this belongs in mixin land for now. Thanks!

slavafomin:
Plus one on this. I think it will be very useful and easy to implement based on existing functions.

PR #2103, opened 2015-03-13
Fix _.indexOf with NaN and isSorted

jridgewell:
Slight breaking change from 1.8.0. Prioritizes finding `NaN` over using a sorted search.

This may not be a good change, since using `isSorted` is only defined to find matches if the array is actually sorted. Before this change, `[1, 5, 2, NaN]` wouldn't find `NaN` (if it were `===`), but afterwards it would.


megawac:
Is this still necessary?

jridgewell:
Never necessary, since it would be changing the behavior of `isSorted`. But, I personally hate that it's included in `indexOf`.

jashkenas:
Then let's leave it be.

PR #2105, opened 2015-03-13
Combine _.indexOf and _.lastIndexOf

jridgewell:
Exactly what the name implies: removes the duplication.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -603,40 +603,8 @@
     return result;
   };
 
-  // Return the position of the first occurrence of an item in an array,
-  // or -1 if the item is not included in the array.
-  // If the array is large and already in sort order, pass `true`
-  // for **isSorted** to use binary search.
-  _.indexOf = function(array, item, isSorted) {
-    var i = 0, length = array && array.length;
-    if (typeof isSorted == 'number') {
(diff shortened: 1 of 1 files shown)


megawac:
Mind forking some benchmarks and including them?

jashkenas:
> removes the duplication

The most noble goal ;)

underscore.js
```diff
@@ -603,40 +603,8 @@
...
-    if (item !== item) {
-      return _.findLastIndex(slice.call(array, 0, idx), _.isNaN);
-    }
-    while (--idx >= 0) if (array[idx] === item) return idx;
-    return -1;
-  };
-
   // Generator function to create the findIndex and findLastIndex functions
-  function createIndexFinder(dir) {
+  function createPredicateIndexFinder(dir) {
```
akre54: Why do we have NFEs here? This seems to be recent (`createReduce` too). Older style (`group` specifically) was to assign to a var.

michaelficarra: That's an FD, not an NFE.

akre54: whatever. pedantry aside, can we swap this back to the style of the rest of the code? It looks ugly.

PR #2106, opened 2015-03-13
Add _.sortedLastIndex

jridgewell:
Closes https://github.com/jashkenas/underscore/issues/1882.

### DIFF
--- a/test/arrays.js
+++ b/test/arrays.js
@@ -112,9 +112,9 @@
   });
 
   test('sortedIndex', function() {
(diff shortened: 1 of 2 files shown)


jashkenas:
Parity sure — but are there any real-world use cases for this?

jridgewell:
Just `_.lastIndexOf` supporting binary search.

jgonggrijp:
I'm favor of having both lower bound and upper bound variants of binary search, but the underlying implementation should take a binary `compare` predicate instead of a boolean flag that lets you choose only between `<` and `<=`.

PR #2107, opened 2015-03-13
Optimize _.range

jridgewell:
Prevents an [arguments deopt](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#31-reassigning-a-defined-parameter-while-also-mentioning-arguments-in-the-body-typical-example).

http://jsperf.com/range-optimization


akre54:
Cool, I can dig it.

Issue #1862, opened 2014-09-19
Consider switching memoize cache to be compatible with ES6 Map/WeakMap.


jdalton:
At the moment the `_.memoize` cache is just a plain object. If we were to switch it to being a simple wrapper around the cache object (still doing the same param to string key use) but with the interface of ES6 [Map](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-map-constructor), [WeakMap](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-weakmap-constructor) that would allow devs to swap in ES6 Maps/WeakMaps for their cache. The thin wrapper would have an interface to mimic Map/WreakMap, so [has](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-weakmap.prototype.has), [set](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-weakmap.prototype.set), [get](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-weakmap.prototype.set), and optionally ([delete](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-weakmap.prototype.delete), [clear](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-weakmap.prototype.clear)).

`_.memoize` could have a `Cache` constructor bolted on to it `_.memoize.Cache` to allow devs to swap it out with a Map/WeakMap or equiv shim as well. 

Map/WeakMap are available on all modern browsers, node --harmony.

Thoughts?

jgonggrijp:
Maybe depends on #2453.

jgonggrijp:
2.0 Should just use `Map` internally. No need for fancy wrappers.

PR #2108, opened 2015-03-13
Use Map shim with _.uniq

jridgewell:
Rounding out my PRs for today: An alternative implementation of https://github.com/jashkenas/underscore/pull/2099, using a Map shim. Performs just as well, but simplifies `_.uniq` instead of making it more complex.

This also [paves the way](https://github.com/jashkenas/underscore/compare/master...jridgewell:maps-memoize?expand=1) for using a Map in `_.memoize` (#1862).


megawac:
> This also paves the way for using a Map in _.memoize (#1862).

This won't be possible (by default) as theres no reasonable way to shim a `WeakMap` so in older environments (such as node 0.10) we'll end up creating some potentially severe memory leaks

megawac:
Why are you using `Map` instead of `Set` here?

See https://github.com/jashkenas/underscore/commit/2ce6f62f32f8bbc8d308748cbef6c9480bf2faf1

jridgewell:
> This won't be possible (by default)

It's just a `Map`, not a `WeakMap`. But this'll allow dev's to use a `WeakMap` if they want.

> Why are you using Map instead of Set here?

I wanted to kill two birds with one stone. And, this is a direct port of #2099.

jgonggrijp:
I think this would be the right thing to do in principle, but I hope it can be done in a way that takes up less space.

Issue #1715, opened 2014-07-02
Inconsistent results when passing empty jQuery object to _.isEmpty


elias6:
I am noticing some strange behavior when running _.isEmpty on an empty jQuery object. At first, _.isEmpty returns false, but if I add an empty collection to it the collection, it stays empty, but _.isEmpty returns true after that. Here is an example:

```
var result = $();
console.log("before", _(result).isEmpty(), result.length);    // before true 0
result = result.add($("p"));
console.log("after", _(result).isEmpty(), result.length);    // after false 0
```

This is on a page with no paragraphs, so the result should stay empty after adding `$("p")`. Is this a bug in Underscore, or is it in jQuery, or am I using isEmpty wrong? It seems like adding the empty collection creates some new properties in the result (length, prevObject, context), but it seems confusing that isEmpty sometimes returns true on empty array-like objects, and sometimes returns false.

I would try to diagnose the problem myself but I am busy. Here is a fiddle: http://jsfiddle.net/56kaw/

Issue #1138, opened 2013-06-05
_.isEmpty return true on the latest version of FireFox if validating an IMG element that failed loading


iMoses:
The title says it all.
Bug occurs on Window, Ubuntu and OSX when using FireFox 21.

PR #2116, opened 2015-03-19
fix _.isEmpty for empty NodeList object in WebKit

rolfb:
After noticing that `_.isEmpty` fails in WebKit (Chrome/Safari) due to `Object.keys(emptyNodeList)` returning `length` as a key I added a failing test and implemented a fix which checks if the object has `length` implemented as a final check before trying `keys` as I'm assuming that if length is available it should be used regardless of the type of object it is.

Please consider it.

Thanks.

### DIFF
--- a/test/objects.js
+++ b/test/objects.js
@@ -566,7 +566,9 @@
     ok(_.isEmpty(), 'undefined is empty');
     ok(_.isEmpty(''), 'the empty string is empty');
     ok(!_.isEmpty('moe'), 'but other strings are not');
-
+    if (testElement) {
+      ok(_.isEmpty(testElement.childNodes), 'NodeList is empty');
+    }
     var obj = {one : 1};
     delete obj.one;
     ok(_.isEmpty(obj), 'deleting all the keys from an object empties it');
--- a/underscore.js
+++ b/underscore.js
@@ -1188,7 +1188,7 @@
   // An "empty" object has no enumerable own-properties.
   _.isEmpty = function(obj) {
     if (obj == null) return true;
-    if (isArrayLike(obj) && (_.isArray(obj) || _.isString(obj) || _.isArguments(obj))) return obj.length === 0;
+    if (isArrayLike(obj) && (_.isArray(obj) || _.isString(obj) || _.isArguments(obj) || _.has(obj, 'length'))) return obj.length === 0;
     return _.keys(obj).length === 0;
   };
 


jridgewell:
Duplicate of https://github.com/jashkenas/underscore/issues/1138. See https://github.com/jashkenas/underscore/issues/1138#issuecomment-19110640 in particular.

rolfb:
I'm not sure if it constitutes a complete duplicate. While it solves a problem related to NodeList it adds a feature to generic collection objects which implements `length`. 

I'm also curious why types are tested so heavily and not just solely relying on `isArrayLike`.
All tests pass with 

```  _.isEmpty = function(obj) {
    if (obj == null) return true;
    if (isArrayLike(obj)) return obj.length === 0;
    return _.keys(obj).length === 0;
  };
```

jridgewell:
`_.isEmpty` is truly meant for enumerable own-properties. The special cases for strings, arrays, and arguments are for cross-compatibility (and speed?).

It's a dup because of the [comment](https://github.com/jashkenas/underscore/issues/1138#issuecomment-19110640) I linked to (with emphasis):

> Second, underscore is attempting to address javascript objects (including arrays), **not host objects**.  An "empty" DOM element could mean many things. Number of children? Enumerable properties? Zero length value? This is not the appropriate place for that check, so it's not a supported argument type.

megawac:
I personally don't get `isEmpty`

```js
_.size(jQuery()) === 0
_.isEmpty(jQuery()) === false;
```

rolfb:
Could you define enumerable own-properties? Would love to understand what the proper use of `_.isEmpty` is.

Thanks.

jridgewell:
Enumerable own-properties are properties that are `enumerable` and set directly on the object, not on it's prototype chain.

```javascript
var prototype = {};
prototype.a = 'enum prototype';
Object.defineProperty(prototype, 'b', { value: 'non-enum prototype', enumerable: false });

var obj = Object.create(prototype);
obj.c = 'enum own'
Object.defineProperty(obj, 'd', { value: 'non-enum own', enumerable: false });

// Enumerable properties of obj: ['a', 'c']
// Enumerable own-properties of obj: ['c']

jridgewell:
Also, https://github.com/jashkenas/underscore/issues/1715 is related.

rolfb:
Appreciated :-)






On lørdag 21. mar. 2015 at 22.15, Justin Ridgewell <notifications@github.com>, wrote:

Also, #1715 is related.

PR #2120, opened 2015-03-25
Make _.uniq doc more precise

phadej:


michaelficarra:
> In particular only the first occurence of each value is kept.

This is not an observable property. Your order example is not in support of this because _.uniq makes no claims about the order of the elements in the output.

phadej:
Should the  documentation, say that `uniq` *might* shuffle the input array, so no-one rely on the input ordering. Especially, how `uniq` picks the elements when `iteratee` is not `_.identity`?

michaelficarra:
Ah yes, with `iteratee` this property is observable. In that case, if we're okay sticking with this behaviour, :+1:.

megawac:
:+1: thanks

jashkenas:
Before merging, we should phrase these things more naturally. In particular, this added sentence doesn't read well.

megawac:
The wording made enough sense to me. Anyway something like this may be better?

> The resulting list will keep the same order of values as the one given.

phadej:
My wording is inspired by Haskell's `[nub](https://hackage.haskell.org/package/base-4.7.0.2/docs/Data-List.html#v:nub)`, which is documented as:

> The nub function removes duplicate elements from a list. In particular, it keeps only the first occurrence of each element.

Yet in JS *removes* implies mutation of the parameter, which `uniq` doesn't do.

PR #2124, opened 2015-03-26
Add Composite Predicate Functions

jridgewell:
Cross posting from lodash/lodash#1072. We already have `_.negate` for negating predicate functions, how about adding something similar to `and` and `or` logic for composing predicate functions. This would allow pretty nifty filtering like:

``` javascript
_.filter(array, _.allPass(_.isObject, _.negate(_.isArray))); // #2123
```

### DIFF
--- a/test/functions.js
+++ b/test/functions.js
@@ -515,6 +515,32 @@
     equal(_.negate(isOdd)(3), false, 'should return the complement of the given function');
   });
 
+  test('allPass', function() {
+    var isOdd = function(n){ return n & 1; };
+    var gt10 = function(n) { return n > 10; };
+    var gt20 = function(n) { return n > 20; };
+    var isOddGt10AndGt20 = _.allPass(isOdd, gt10, gt20);
+    equal(isOddGt10AndGt20(21), true, 'true if all predicates pass');
+    equal(isOddGt10AndGt20(22), false, 'false if any predicate fails');
+    equal(isOddGt10AndGt20(20), false, 'false if any predicate fails');
+    equal(isOddGt10AndGt20(19), false, 'false if any predicate fails');
+    equal(isOddGt10AndGt20(10), false, 'false if any predicate fails');
+    equal(isOddGt10AndGt20(9), false, 'false if any predicate fails');
+  });
+
+  test('allPass', function() {
+    var isOdd = function(n){ return n & 1; };
+    var gt10 = function(n) { return n > 10; };
+    var gt20 = function(n) { return n > 20; };
+    var isOddGt10OrGt20 = _.anyPass(isOdd, gt10, gt20);
+    equal(isOddGt10OrGt20(22), true, 'true if any predicates pass');
+    equal(isOddGt10OrGt20(21), true, 'true if any predicates pass');
+    equal(isOddGt10OrGt20(20), true, 'true if any predicates pass');
+    equal(isOddGt10OrGt20(19), true, 'true if any predicates pass');
+    equal(isOddGt10OrGt20(9), true, 'true if any predicates fail');
+    equal(isOddGt10OrGt20(10), false, 'false if all predicates fail');
+  });
+
   test('compose', function() {
     var greet = function(name){ return 'hi: ' + name; };
     var exclaim = function(sentence){ return sentence + '!'; };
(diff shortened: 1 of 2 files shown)


jridgewell:
Re: https://github.com/jashkenas/underscore/issues/2008

jridgewell:
Note: I canned the `_.none` and `_.nonePass`, I was doing the wrong logic tables. :grimacing:

megawac:
:+1:

jashkenas:
Got a few more real-world use case examples up your sleeve?

jridgewell:
Searching through the codebase for the right `&&` is impossible. :pensive:

I've used patterns like the following, vs their functional eqivalent:

```javascript
_.filter(collection, function(model) {
    return model.isActive && model.method();
});

// or
var call = function(method) {
    return function(obj) { return obj[method](); };
};
_.filter(collection, _.allPass(_.property('isActive'), call('method')));
```

and

```javascript
_.findIndex(array, function(obj) {
    obj.isActive && !_.isEmpty(obj.els);
});

// or
_.findIndex(array, _.allPass(_.property('isActive'), _.compose(_.isEmpty, _.property('els'))));

**Comment by jashkenas** [comment]
Right, and in both of those cases, the former is way nicer than the latter. Not super persuasive for a core feature...

**Comment by jridgewell** [comment]
Neither of the functional examples would look bad if the helper methods were already defined.

```javascript
_.filter(collection, _.allPass(isActive, call('method')));
_.findIndex(array, _.allPass(isActive, emptyEls));
```

Maybe @megawac will be able to find one of his uses quicker than I am? But I like this because of how nifty they don't require me writing a function. Kind of like `_.filter(collection, 'isActive')`.

megawac:
My use cases generally tend to need `anyPass` more often than `allPass` are usually to determine if `a`, `b` or `c` are true

For example, from some acorn AST parsing I'm working on

```js
_.chain(findImports(ast.body))
 .filter(_.anyPass({type: 'AMDImport'}, {type: 'ImportDeclaration'}, {type: 'CJSImport'}))
 .filter(...)
 .map(...)
```

aearly:
I'm :+1: for this.  I would find it useful for creating predicates that verify types.

```js
var isCollection = _.anyPass(Array.isArray, _.isPlainObject, _.isArguments);

function takesACollection(arg1) {
  assert(isCollection(arg1));
  //...
}
```

I would think of them as `and` and `or` operations.

jashkenas:
Should we call them `_.and` and `_.or` then?

aearly:
`_.and(f, g)` might be confused for `(f, g) => f && g` rather than `(f, g) => f() && g()` -- a simple logic operation, rather than combining predicates.

Ramda calls the simple logic operator [`R.and`](http://ramdajs.com/docs/#and), and the function that combines predicates [`R.allPass`](http://ramdajs.com/docs/#allPass).

Personally, I like `_.and` and `_.or` better, since they're shorter.  I also see myself using predicate combinators more often than wrappers around operators.

jdalton:
Are there any other lang equiv names in haskell or smth?

aearly:
In Clojure there is [some-fn](http://clojure.github.io/clojure/clojure.core-api.html#clojure.core/some-fn) and [every-pred](http://clojure.github.io/clojure/clojure.core-api.html#clojure.core/every-pred).  (A bit strange that the names are not symmetric.)

michaelficarra:
@jdalton Haskell actually has a concept that is far more general than this. Because functions are applicative functors, we can "lift" any function that operates on some type to a function that operates on functions that operate on that type. This example should make that explanation clearer:

```ghci
λ> let a = const True
λ> let b = const False
λ> import Control.Applicative
λ> (liftA2 (&&)) a b 0
False
λ> (liftA2 (||)) a b 0
True
λ>
```

A JavaScript version of `liftA2` specialised to functions would look like this:

```js
function liftA2(combine) {
  return function(f, g) {
    return function(){
      return combine(f.apply(null, arguments), g.apply(null, arguments));
    };
  };
}
let and = (a, b) => a && b;
let f = x => x > 0;
let g = y => y > 2;
liftA2(and)(f, g)(5) // true
liftA2(and)(f, g)(1) // false
```

Of course, being JavaScript, we could allow a `liftA` that accepts a combining function of any arity and any number of functions to apply to the lifted function.

tilmanpotthof:
+1

jdalton:
After much bikeshedding I went with the names `_.conj` and `_.disj` for [conjunction](https://en.wikipedia.org/wiki/Logical_conjunction) and [disjunction](https://en.wikipedia.org/wiki/Logical_disjunction). This avoids the `any`/`all` - `some`/`every` issue and connects well with [_.negate](https://en.wikipedia.org/wiki/Negation).

robinpokorny:
Thi PR can be closed now.

PR #2134, opened 2015-04-01
Fix iOS JIT Bug, Part 2

jridgewell:
Given Closure Compiler’s [need to inline](https://github.com/jashkenas/underscore/pull/2094#issuecomment-88636347),
we need to grab the length using a function (that uses dynamic property access).

@megawac also pointed out a few other methods would be affected.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -122,13 +122,20 @@
     return result;
   };
 
+  var property = function(key) {
+    return function(obj) {
+      return obj == null ? void 0 : obj[key];
+    };
+  };
+
   // Helper for collection methods to determine whether a collection
   // should be iterated as an array or as an object
   // Related: http://people.mozilla.org/~jorendorff/es6-draft.html#sec-tolength
   // Avoids a very nasty iOS 8 JIT bug on ARM-64. #2094
-  var MAX_ARRAY_INDEX = Math.pow(2, 53) - 1, LENGTH = 'length';
+  var MAX_ARRAY_INDEX = Math.pow(2, 53) - 1;
+  var getLength = property('length');
   var isArrayLike = function(collection) {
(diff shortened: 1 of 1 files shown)


underscore.js
```diff
@@ -1506,7 +1507,7 @@
     _.prototype[name] = function() {
       var obj = this._wrapped;
       method.apply(obj, arguments);
-      if ((name === 'shift' || name === 'splice') && obj.length === 0) delete obj[0];
+      if ((name === 'shift' || name === 'splice') && getLength(obj) === 0) delete obj[0];
```
megawac: I don't think it's necessary here as this is a fix for IE<9

megawac:
:+1:

underscore.js
```diff
@@ -1287,11 +1292,7 @@
 
   _.noop = function(){};
 
-  _.property = function(key) {
-    return function(obj) {
-      return obj == null ? void 0 : obj[key];
-    };
-  };
+  _.property = property;
```
megawac: Any reason you don't just hoist the definition to where its necessary

underscore.js
jridgewell: Grep and replace. :grin:

underscore.js
jridgewell: To keep it inline with the other Object functions. Same as with the `create*` helper functions, which we then [assign to methods](https://github.com/jashkenas/underscore/blob/master/underscore.js#L1001-L1005).

jridgewell:
And once again, it's node v0.8 that's failing the test while installing deps, not the PR failing.

jridgewell:
@megawac: dropped.

megawac:
https://github.com/jashkenas/underscore/pull/2135

Issue #1594, opened 2014-04-19
Implement `_.partition` by way of `group`


jdalton:
You can implement [_.partition](https://github.com/jashkenas/underscore/blob/813d39212fcde8446caee455b109daa618ab87b9/underscore.js#L453-L460) by way of [group()](https://github.com/jashkenas/underscore/blob/813d39212fcde8446caee455b109daa618ab87b9/underscore.js#L323-L333) if you modify `group` to accept an `initializer` function. Something like:

``` js
  var group = function(behavior, initializer) {
    return function(obj, iterator, context) {
      var result = initializer ? initializer() : {};
      iterator = lookupIterator(iterator, context);
      _.each(obj, function(value, index) {
        var key = iterator(value, index, obj);
        behavior(result, key, value);
      });
      return result;
    };
  };

  _.partition = group(function(result, key, value) {
    result[key ? 0 : 1].push(value);
  }, function() { return [[], []]; });
```

PR #2137, opened 2015-04-02
Implement partion via group

### LINKED ISSUE #2060: Go on a Diet
Underscore is getting a little chunky around the edges...

We're up to 5.8k, minified and gzipped. It would be lovely to get that back down under 5...

megawac:
Fixes https://github.com/jashkenas/underscore/issues/1594

#2060...

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -391,9 +391,9 @@
   };
 
   // An internal function used for aggregate "group by" operations.
-  var group = function(behavior) {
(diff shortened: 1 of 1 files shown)


jridgewell:
Is there a reason to pass an initializer instead of a simple boolean like [createAssigner](https://github.com/jashkenas/underscore/blob/master/underscore.js#L98)?

megawac:
No particular one

PR #2140, opened 2015-04-03
Add _.restParam

jridgewell:
`_.restParam` is the equivalent of ES6's rest parameter. It's particularly useful as an alternative to `slice.call(arguments, index)` (numerous examples in this PR), and it's a reusable, very fast version of inline for-loop slicing, eg. https://github.com/jashkenas/backbone/pull/3489.

```javascript
var test = function(a, b, ...args) {
   console.log(arguments);
}

test(1, 2, 3, 4, 5); // => [1, 2, [3, 4, 5]]

var test = _.restParam(function(a, b, args) {
   console.log(arguments);
});

test(1, 2, 3, 4, 5); // => [1, 2, [3, 4, 5]]
```

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -98,7 +98,7 @@
   var restParam = function(func, startIndex) {
     startIndex = startIndex == null ? func.length - 1 : +startIndex;
     return function() {
-      var length = arguments.length > startIndex ? arguments.length - startIndex : 0;
+      var length = Math.max(arguments.length - startIndex, 0);
       var rest = Array(length);
       for (var index = 0; index < length; index++) {
         rest[index] = arguments[index + startIndex];
--- a/underscore.js
+++ b/underscore.js
@@ -95,7 +95,7 @@
 
   // Similar to ES6's rest params (http://ariya.ofilabs.com/2013/03/es6-and-rest-parameter.html)
   // This accumulates the arguments passed into an array, after a given index.
-  _.restParam = function(func, startIndex) {
+  var restParam = function(func, startIndex) {
     startIndex = startIndex == null ? func.length - 1 : +startIndex;
     return function() {
       var length = arguments.length > startIndex ? arguments.length - startIndex : 0;
@@ -112,7 +112,7 @@
       for (index = 0; index < startIndex; index++) {
         args[index] = arguments[index];
       }
-      args[index] = rest;
+      args[startIndex] = rest;
       return func.apply(this, args);
     };
   };
@@ -274,7 +274,7 @@
   };
 
   // Invoke a method (with arguments) on every item in a collection.
-  _.invoke = _.restParam(function(obj, method, args) {
+  _.invoke = restParam(function(obj, method, args) {
     var isFunc = _.isFunction(method);
     return _.map(obj, function(value) {
       var func = isFunc ? method : value[method];
@@ -511,7 +511,7 @@
   };
 
   // Return a version of the array that does not contain the specified value(s).
-  _.without = _.restParam(function(array, otherArrays) {
+  _.without = restParam(function(array, otherArrays) {
     return _.difference(array, otherArrays);
   });
 
@@ -711,7 +711,7 @@
     if (nativeBind && func.bind === nativeBind) return nativeBind.apply(func, slice.call(arguments, 1));
     if (!_.isFunction(func)) throw new TypeError('Bind must be called on a function');
     var args = slice.call(arguments, 2);
-    var bound = _.restParam(function(callArgs) {
+    var bound = restParam(function(callArgs) {
       return executeBound(func, bound, context, this, args.concat(callArgs));
     });
     return bound;
@@ -721,7 +721,7 @@
   // arguments pre-filled, without changing its dynamic `this` context. _ acts
   // as a placeholder by default, allowing any combination of arguments to be
   // pre-filled. Set `_.partial.placeholder` for a custom placeholder argument.
-  _.partial = _.restParam(function(func, boundArgs) {
+  _.partial = restParam(function(func, boundArgs) {
     var placeholder = _.partial.placeholder;
     var bound = function() {
       var position = 0, length = boundArgs.length;
@@ -740,7 +740,7 @@
   // Bind a number of an object's methods to that object. Remaining arguments
   // are the method names to be bound. Useful for ensuring that all callbacks
   // defined on an object belong to it.
-  _.bindAll = _.restParam(function(obj, keys) {
+  _.bindAll = restParam(function(obj, keys) {
     if (keys.length < 1) throw new Error('bindAll must be passed function names');
     return _.each(keys, function(key) {
       obj[key] = _.bind(obj[key], obj);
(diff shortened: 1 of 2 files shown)


megawac:
This is heavily inspired by jashkenas/backbone#3489 - related jsperf http://jsperf.com/trigger-with-for-loop-vs-slice/8
Also an awesome pattern it turns out :)

jashkenas:
Super nifty.

How about naming it `_.arguments` or `_.varArgs` instead. It is the "arguments" object, after all.

megawac:
@jashkenas the name is based off of ES6 `rest` parameters - we debated the
name pretty heavily over the past couple months before submitting this

Also arguments is ~ a reserved word
http://ecma-international.org/ecma-262/5.1/#sec-12.2.1

jashkenas:
The "parameter/argument" naming dichotomy would be unfortunate to introduce — Underscore has historically called them "arguments".

How about `_.restArgs`?

jridgewell:
I'm not against `_.restArgs`, though I think `_.restParam` is a much better name since it mirrors ES6's rest parameter.

Also, updated to it merges nicely.

jashkenas:
"parameters" and "arguments" are synonyms for the purposes of JavaScript. "Arguments" is already the established word choice — both for JavaScript with the `arguments` object, and for Underscore with the `_.isArguments` function. `_.restArgs` would be a more consistent name.

jridgewell:
Sigh, updated.

jashkenas:
Groovy. No way to combine those two similar loops, eh?

megawac:
> Groovy. No way to combine those two similar loops, eh?

Not without perf penalties (see the initial 2 commits)

jridgewell:
Yah, there's a pretty bad performance hit, which would defeat the goal of adding this to Backbone's `#trigger`.

michaelficarra:
`_.rest` wouldn't have worked?

jashkenas:
`_.rest` would have been great, but alas:

http://underscorejs.org/#rest

underscore.js
```diff
@@ -94,6 +93,30 @@
     return cb(value, context, Infinity);
   };
 
+  // Similar to ES6's rest param (http://ariya.ofilabs.com/2013/03/es6-and-rest-parameter.html)
+  // This accumulates the arguments passed into an array, after a given index.
+  var restArgs = function(func, startIndex) {
+    startIndex = startIndex == null ? func.length - 1 : +startIndex;
```
akre54: Using function.length means functions bound with the `_.bind` fallback will work inconsistently with ones bound with `Function.prototype.bind` btw. Is this acceptable? @megawac @jridgewell 

```js
var func = function(a,b,c,d,e,f,g) {};

func.length; // 7
func.bind(null).length; // 7

Function.prototype.bind = null;
_.bind(func, null).length; // 0
```

underscore.js
```diff
@@ -1489,11 +1508,10 @@
   _.mixin = function(obj) {
     _.each(_.functions(obj), function(name) {
       var func = _[name] = obj[name];
-      _.prototype[name] = function() {
-        var args = [this._wrapped];
-        push.apply(args, arguments);
+      _.prototype[name] = restArgs(function(args) {
```
akre54: Is this faster? It's certainly not clearer.

underscore.js
jridgewell: I think the usefulness of not having to give the `startIndex` every times outweighs this trip-up. A lot of what I imagine using it for is in parse-time. You would then bind the returned function.

```javascript
var args = _.restArgs(function(args) { return args; }).bind(null);
```

underscore.js
jridgewell: It actually looks like it's [slower](http://jsperf.com/restargs-mixin). @megawac?

underscore.js
akre54: Ok cool. You could still pass `startIndex` to restArgs if you needed the fallback behavior, that seems reasonable.

underscore.js
megawac: It seemed faster when I profiled it locally on node - I'll check again  this weekend

akre54:
I'm :-1: on this change. It may be ugly but  `slice.call(arguments, 1)` is idiomatic JS. It's not yet another underscore-specific helper function that makes code harder to read and reason about.

I'd also argue all of these extra helper methods make Underscore itself harder to read. I really wish we'd simplify a lot of these.

jashkenas:
I agree that this one is a little bit weird — but it's a neat pattern to automatically capture varargs, and to avoid dealing with the arguments object. I think it's worth having.

jridgewell:
> I'm :-1: on this change. It may be ugly but slice.call(arguments, 1) is idiomatic JS. It's not yet another underscore-specific helper function that makes code harder to read and reason about.


The point of this was to get us closer to idiomatic ES6, and provide a speed boost since `slice.call(arguments)` [murders performance](https://github.com/jashkenas/backbone/pull/3489).

akre54:
That's great, I just really don't like the "lodashification of Underscore" we've been seeing recently -- helper methods everywhere and a cargo-cult-like focus on microbenchmarks and minor speed improvements at the expense of Underscore's readability. `arguments` de-opt fixes are fine but with this change on any given method you have to read several levels of helpers down to actually understand what it's doing.

One of the best things about Underscore of days past was that it looked like code you might actually write in your project, and in a pinch you could pull out just the parts you needed. Lodash already exists as a separate approach, I'm not sure we need to be copying them here.

megawac:
> One of the best things about Underscore of days past was that it looked like code you might actually write in your project, and in a pinch you could pull out just the parts you needed. Lodash already exists as a separate approach, I'm not sure we need to be copying them here.

I agree with you and thats why I've avoided several refactors in the past. Me and @jridgewell have just been refactoring in the places which will gain maximum benefit for the library in terms of size and speed. The purpose of a library such as underscore/lodash should be to help devs write quality code

Also a distinction is that nearly all of the helpers used internally are exposed and documented in the API. I'd say its fine for a library to eat its own dog food like this

akre54:
> murders performance

Is this [the jsperf](http://jsperf.com/trigger-with-for-loop-vs-slice/6) you're talking about? Sure twice as fast is always awesome, but when you need to trigger 400,000 events in a second, I highly doubt the trigger method (or its args de-opt) is going to be your bottleneck.

akre54:
> its fine for a library to eat its own dog food like this

Sure. I'm more referring to the internal `createX` methods (`reduce` comes to mind as particularly egregious) that make Underscore's source nearly impossible to understand if you aren't already intimately familiar.

jashkenas:
> Sure. I'm more referring to the internal createX methods (reduce comes to mind as particularly egregious) that make Underscore's source nearly impossible to understand if you aren't already intimately familiar.

Those are indeed much more problematic. So perhaps this particular issue isn't the right place to make this argument?

It would be great to look at a PR that conceptually simplified the create* mega-methods...

akre54:
> So perhaps this particular issue isn't the right place to make this argument?

Totally. This was kinda the straw that broke the camel's back. I wanted to register that I see this change as part of a larger more unfortunate trend. A PR or discussion of how to fix this would be excellent.

jridgewell:
@akre54 Mind creating it with your thoughts?

akre54:
Will do when I get some time. Work business and non work busyness and all
that. It's been on my to do list

jridgewell:
:+1:

PR #2142, opened 2015-04-03
Combine _.min and _.max

jridgewell:
http://jsperf.com/min-max-code-sharing

There's a hit on arrays without an iteratee. Looking into speeding that up, but everything else is faster.


jashkenas:
We really gotta fix those `create*` names...

jridgewell:
> We really gotta fix those create* names...

It'd probably be good. Want to do that afterwards?

jridgewell:
Hmm, I can get speeds to roughly the same levels, but brings back some duplication.


```
max (master, array, no iteratee) x 5,666,595 ops/sec ±1.46% (89 runs sampled)
max (combine, array, no iteratee) x 5,237,906 ops/sec ±2.17% (86 runs sampled)
max (master, array, iteratee) x 5,631,853 ops/sec ±1.51% (88 runs sampled)
max (combine, array, iteratee) x 5,169,098 ops/sec ±1.82% (83 runs sampled)
max (master, obj, no iteratee) x 342,269 ops/sec ±1.55% (82 runs sampled)
max (combine, obj, no iteratee) x 358,430 ops/sec ±1.95% (82 runs sampled)
max (master, obj, iteratee) x 324,928 ops/sec ±3.26% (82 runs sampled)
max (combine, obj, iteratee) x 361,950 ops/sec ±1.87% (86 runs sampled)
Fastest is max (master, array, no iteratee),max (master, array, iteratee)
```

```javascript
var createExtremumFinder = function(dir) {
  return function(obj, iteratee, context) {
    var keys = !isArrayLike(obj) && _.keys(obj),
        length = (keys || obj).length,
        result = dir * Infinity,
        lastComputed = result,
        index = 0, currentKey, value;
    if (iteratee == null) {
        currentKey = keys ? keys[index] : index;
        value = obj[currentKey];
        if (dir < 0 ? result > value : result < value) {
          result = value;
        }
      }
    } else {
        currentKey = keys ? keys[index] : index;
        value = obj[currentKey];
        var computed = iteratee(value, currentKey, obj);
        if ((dir < 0 ? computed > lastComputed : computed < lastComputed) || index === 0 && computed === lastComputed) {
          result = value;
          lastComputed = computed;
        }
      }
    }
    return result;
  };
};
```

underscore.js
```diff
@@ -277,56 +277,33 @@
...
+          result = max ? -Infinity : Infinity,
+          lastComputed = result,
+          found = false;
+      iteratee = iteratee == null ? null : cb(iteratee, context);
+      for (var index = 0; index < length; index++) {
+        var currentKey = keys ? keys[index] : index;
+        var value = obj[currentKey];
+        var computed = iteratee ? iteratee(value, currentKey, obj) : value;
+        if ((max ? computed > lastComputed : computed < lastComputed) || (!found && computed === result)) {
+          found = true;
```
megawac: is the found variable necessary? Can't you just infer it based on index

underscore.js
```diff
@@ -277,56 +277,33 @@
...
-        if (computed > lastComputed || computed === -Infinity && result === -Infinity) {
+  // Generator function to create the max and min functions
+  var createExtremumFinder = function(max) {
+    return function(obj, iteratee, context) {
+      var keys = !isArrayLike(obj) && _.keys(obj),
+          length = (keys || obj).length,
+          result = max ? -Infinity : Infinity,
+          lastComputed = result,
+          found = false;
+      iteratee = iteratee == null ? null : cb(iteratee, context);
```
megawac: When I was playing around with it a while ago I found the `identity` iteratee was often faster than all the branching

underscore.js
jridgewell: Well, the iteratee can return anything. So, if the first iteration returns `undefined`, that won't be greater than `-Infinity`. But, the next iteration might be equal.

underscore.js
jridgewell: Looks like it, updated.

megawac:
Its faster for the `function` and all object cases so thats good enough for me :+1:

jashkenas:
Rebase for a clean merge?

jridgewell:
Rebased.

PR #2146, opened 2015-04-10
Change flatten to iterative version to be good for deeply nested arrays

davidxi:
One of my use cases is to flatten some deeply nested arrays. Then I studied the code of that function, and found it is using recursive method. So this code change attempt is to change it to iterative way.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -485,24 +485,35 @@
     return _.filter(array, _.identity);
   };
 
-  // Internal implementation of a recursive `flatten` function.
+  // Internal implementation of a iterative `flatten` function.
   var flatten = function(input, shallow, strict, startIndex) {
-    var output = [], idx = 0;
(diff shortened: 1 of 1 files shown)


underscore.js
```diff
@@ -485,24 +485,35 @@
     return _.filter(array, _.identity);
   };
 
-  // Internal implementation of a recursive `flatten` function.
+  // Internal implementation of a iterative `flatten` function.
```
michaelficarra: It still recursively flattens, regardless of the implementation.

michaelficarra:
Can we see a jsperf comparison? This looks like it will be terribly inefficient.

davidxi:
yes I saw it's actually slower when running the unit test cases, cause in the unit test case, there are test cases of super long arrays, which is actually weak of this implementation (it uses array reverse + clone). Since this one would be more beneficial to super deep arrays and eliminate function call costs. So it depends on which type of complexity of array is more frequently used.

PR #2153, opened 2015-04-14
Detect global in strict mode and WebWorkers

### LINKED ISSUE #2152: Use Function("return this") to get global?
Underscore currently breaks in `'strict'` mode while trying to access properties on the global object. The global object is being set using the `(function(root) { ... })(this)` pattern. However, in this case, `root` remains `undefined` and hence the issue.

Using the `Function("return this")()` pattern instead alleviates the issue, though arguably ugly. Thoughts?

**Steps to reproduce**: Import underscore in a `babel-jest` test harness. (uses strict mode)

megawac:
fixes #2152

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -8,8 +8,10 @@
   // Baseline setup
   // --------------
 
-  // Establish the root object, `window` in the browser, or `exports` on the server.
-  var root = this;
+  // Establish the root object, `window` (`self`) in the browser, or `global` on the server.
+  // We use `self` instead of `window` for `WebWorker` support.
+  var root = (typeof self == 'object' && self.self == self && self) ||
+            (typeof global == 'object' && global.global == global && global);
 
   // Save the previous value of the `_` variable.
   var previousUnderscore = root._;
@@ -1562,4 +1564,4 @@
       return _;
     });
   }
-}.call(this));
+}());


michaelficarra:
It bothers me that the only detection we do now is via de facto self-references on the global object. I wish there was a more reliable way to do this, though at the same time I don't want it to be possible to get a reference to the global object in strict mode.

underscore.js
```diff
@@ -8,8 +8,9 @@
   // Baseline setup
   // --------------
 
-  // Establish the root object, `window` in the browser, or `exports` on the server.
-  var root = this;
+  // Establish the root object, `window` in the browser, or `global` on the server.
+  var root = typeof self == 'object' && self.Object == Object && self;
```
michaelficarra: `self.Object == Object` would be better as `self.self === self`.

megawac: Good call

jridgewell:
I'm not entirely sure how to test, but what are `self` and `global` in ES6 modules? I assume they're either the browser's `window` or node's `global` and the only difference is the context of the module is `undefined` instead of one of those?

megawac:
Yeah `this` will be undefined so instead we detect the global scope variable, as you mentioned `self` in the browser & `global` in node. I think your fix would work as well `root = this || {}`

michaelficarra:
@megawac no, I believe `this` will be a ReferenceError.

jridgewell:
:+1:. We should also apply this to Backbone.

jashkenas:
I've just fixed this in Backbone like so (a little looser): jashkenas/backbone@5a6ffcb

jridgewell:
It probably won't hurt much in Backbone, but avoiding `self` will prevent underscore from working in WebWorkers.

> workers run in another global context that is different from the current window. Thus, using the window shortcut to get the current global scope (instead of self) within a Worker will return an error. - [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers)

jridgewell:
Merging in light of https://github.com/jashkenas/backbone/pull/3602.

megawac:
@jashkenas may justify a patch release

On Thu, May 14, 2015 at 12:02 PM, Justin Ridgewell <notifications@github.com

zertosh:
I would like to point out that if you use a build tool like browserify, a reference to `global`, causes the module to be [wrapped](https://github.com/substack/insert-module-globals/blob/68ec279f6365bfe34a37ef5136ebfeec3f7794a1/index.js#L14-L19) in:

```js
(function (global){
/** your original module **/
}).call(this,typeof global !== "undefined" ? global : typeof self !== "undefined" ? self : typeof window !== "undefined" ? window : {})
```

I'm only pointing this out for the benefit of those that use browserify and wonder why underscore gets that wrapper. I'm not suggesting to build underscore with browserify, and I'm not suggesting this is a bad change.

dcocchia:
This fix was merged with master in May, 2015. Almost 2 years later, can we release a patch?

Kosta-Github:
I second the request from @dcocchia ? can you distribute the fix as a new release, please?

Issue #466, opened 2012-02-09
setImmediate() with _.defer()


jareware:
Any plans on delegating to window.setImmediate() where available in _.defer()..?

https://developer.mozilla.org/en/Document_Object_Model_(DOM)/window.setImmediate
http://ie.microsoft.com/testdrive/Performance/setImmediateSorting/Default.html

Would match the documented use case of "Defers invoking the function until the current call stack has cleared, similar to using setTimeout with a delay of 0" though it's perhaps not Underscore's place to start shimming for experimental browser features.

PR #2154, opened 2015-04-15
Modify defer to use native task queue via promise if available

MidnightWonderer:
Hi there
I wonder if we can use native defer when run on environment that has Promise interface available?
Take advantage from browser optimized task queue.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -768,8 +768,20 @@
   });
 
   // Defers a function, scheduling it to run after the current call stack has
-  // cleared.
-  _.defer = _.partial(_.delay, _, 1);
+  // cleared. Use native promise if available.
+  _.defer = (function() {
+    if (!Promise) {
+      return _.partial(_.delay, _, 1);
+    }
+    var _promise = new Promise(function(resolve) {
+      resolve();
+    });
+    return restArgs(function(func, args) {
+      _promise.then(function() {
+        func.apply(null, args);
+      });
+    });
+  })();
 
   // Returns a function, that, when invoked, will only be triggered at most once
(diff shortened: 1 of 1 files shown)


MidnightWonderer:
Hi @jdalton 
I can not get this to pass TravisCI but I glad you got the idea.

This approach allow some room for browser to optimize in particular area.
In my opinion the accuracy should come first in setTimeout optimiztion. (Best effort for accuracy)
In the other hand promise optimization will focus on performance and maybe memory consumption.

Please refer to ES6 draft
http://people.mozilla.org/~jorendorff/es6-draft.html#sec-jobs-and-job-queues
where they introduce the idea of task queue in ES6 to use in Promise.

For the inconsistent:
Native defer will be triggered before setTimeout.
Please refer to some test on node.js repo
https://github.com/joyent/node/issues/7714#issuecomment-51993171

For the polyfill promise that comply to Promises/A+ specification (not jQuery one) my best guess is that it also use setTimeout to simulate the effect. (AFAIK no other mechanism to accomplish this in ES5) So for polyfill Promise it will do essentially the same as setTimeout 1.

MidnightWonderer:
I don't know what to do with the code in this case but this will do just to demonstrate the idea.
Should I just add lint directive to tell that Promise is a global object?

I'm open to hear review / suggestion / recommendation from you guys.

megawac:
I'd be more keen on using `setImmediate` over the `Promise` based implementation, but I can relate to the arguments in https://github.com/jashkenas/underscore/issues/466. Either way I would be skeptical of using a Promise based implementation of defer to avoid issues from Promise polyfills

PR #2155, opened 2015-04-15
adding advice on when to prefer <%- ... -%> over <%= ... =>

ttelang:
From my experience using underscore template. Adding advice on how to sanitize your template usage against cross-site scripting vulnerability.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1963,7 +1963,9 @@ _.now();
         data sources. Template functions can both interpolate values, using
         <tt>&lt;%= &hellip; %&gt;</tt>, as well as execute arbitrary JavaScript code, with
         <tt>&lt;% &hellip; %&gt;</tt>. If you wish to interpolate a value, and have
-        it be HTML-escaped, use <tt>&lt;%- &hellip; %&gt;</tt>. When you evaluate a
+        it be HTML-escaped, use <tt>&lt;%- &hellip; %&gt;</tt>. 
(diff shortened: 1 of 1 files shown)


jridgewell:
Ugh. We get this so often, it may be worth rewording to present `<%- %>` as the "normal", and `<%= %>` as the "insecure" version -- as opposed to `<%= %>` as the "normal" and `<%- %>` as the "secure".

How about

> Template functions can both interpolate values, using `<%- … %>`, as well as execute arbitrary JavaScript code, with `<% … %>`. If you wish to interpolate a value without HTML-escaping, use `<%= … %>`.

jashkenas:
Nah, let's leave it be. Kind of an ideological thing, but — for the types of uses that I personally do at work (NYT), and Underscore was used for in its original project (DocumentCloud) — `<%=` is almost always what you want, and `<%-` is the rare exception.

Of course, use cases may vary, but thinking of Underscore templates as *basically* simple string concatenation is a nice way to start.

ttelang:
@jashkenas - I and many other co-developers with whom I interact with use mostly <%- and very rarely <%= (as malicious JavaScript code can be injected within this).

michaelficarra:
@ttelang You're probably using underscore templates only for HTML templating. They are supposed to provide a more general facility than that, so the default should be simple interpolation.

PR #2158, opened 2015-04-21
Make _.sample more efficient by exiting early

marekventur:
This makes ```_.sample``` more efficient by re-implementing ```_.shuffle```'s  logic and exiting early when enough elements are found. The speedup to the previous method of shuffling the whole input is especially large when the input is a long array and only a few sample elements are required. 

It is safe to exit a Fisher-Yates shuffle early: http://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#Comparison_with_other_shuffling_algorithms

There are two alternatives to the way I implemented it which would require either factoring the main ```_.shuffle``` logic out into a private function and using it in both ```_.sample``` and ```_.shuffle``` or adding an internal parameter to ```_.shuffle``` that limits the result to a certain number, thus making ```_.shuffle``` even more similar to ```_.sample```. Please let me know if any of those two options are preferred over the one I have here.

### DIFF
--- a/test/collections.js
+++ b/test/collections.js
@@ -737,12 +737,12 @@
   });
 
   test('shuffle', function() {
-    var numbers = _.range(10);
+    deepEqual(_.shuffle([1]), [1], 'behaves correctly on size 1 arrays');
+    var numbers = _.range(20);
     var shuffled = _.shuffle(numbers);
+    notDeepEqual(numbers, shuffled, 'does change the order'); // Chance of false negative: 1 in ~2.4*10^18
     notStrictEqual(numbers, shuffled, 'original object is unmodified');
-    ok(_.every(_.range(10), function() { //appears consistent?
-      return _.every(numbers, _.partial(_.contains, numbers));
-    }), 'contains the same members before and after shuffle');
+    deepEqual(numbers, _.sortBy(shuffled), 'contains the same members before and after shuffle');
 
     shuffled = _.shuffle({a: 1, b: 2, c: 3, d: 4});
     equal(shuffled.length, 4);
@@ -750,6 +750,8 @@
   });
 
   test('sample', function() {
+    strictEqual(_.sample([1]), 1, 'behaves correctly when no second parameter is given');
+    deepEqual(_.sample([1, 2, 3], -2), [], 'behaves correctly on negative n');
     var numbers = _.range(10);
     var allSampled = _.sample(numbers, 10).sort();
     deepEqual(allSampled, numbers, 'contains the same members before and after sample');
@@ -761,6 +763,9 @@
     notStrictEqual(_.sample([1, 2, 3], 0), [], 'sampling an array with 0 picks returns an empty array');
     deepEqual(_.sample([1, 2], -1), [], 'sampling a negative number of picks returns an empty array');
     ok(_.contains([1, 2, 3], _.sample({a: 1, b: 2, c: 3})), 'sample one value from an object');
+    var partialSample = _.sample(_.range(1000), 10);
+    var partialSampleSorted = partialSample.sort();
+    notDeepEqual(partialSampleSorted, _.range(10), 'samples from the whole array, not just the beginning');
   });
 
   test('toArray', function() {
--- a/underscore.js
+++ b/underscore.js
@@ -349,21 +349,13 @@
     return result;
   };
 
-  // Shuffle a collection, using the modern version of the
-  // [Fisher-Yates shuffle](http://en.wikipedia.org/wiki/Fisher–Yates_shuffle).
+  // Shuffle a collection.
   _.shuffle = function(obj) {
-    var set = isArrayLike(obj) ? obj : _.values(obj);
-    var length = set.length;
-    var shuffled = Array(length);
-    for (var index = 0, rand; index < length; index++) {
-      rand = _.random(0, index);
-      if (rand !== index) shuffled[index] = shuffled[rand];
-      shuffled[rand] = set[index];
-    }
-    return shuffled;
+    return _.sample(obj, Infinity);
   };
 
-  // Sample **n** random values from a collection.
+  // Sample **n** random values from a collection using the modern version of the
+  // [Fisher-Yates shuffle](http://en.wikipedia.org/wiki/Fisher–Yates_shuffle).
   // If **n** is not specified, returns a single random element.
   // The internal `guard` argument allows it to work with `map`.
   _.sample = function(obj, n, guard) {
@@ -371,7 +363,18 @@
       if (!isArrayLike(obj)) obj = _.values(obj);
       return obj[_.random(obj.length - 1)];
     }
-    return _.shuffle(obj).slice(0, Math.max(0, n));
+
+    var sample = isArrayLike(obj) ? _.clone(obj) : _.values(obj);
+    var length = getLength(sample);
+    n = Math.max(Math.min(n, length), 0);
+    var rand, temp;
+    for (var index = length - 1; index > length - n - 1; index--) {
+      rand = _.random(0, index);
+      temp = sample[index];
+      sample[index] = sample[rand];
+      sample[rand] = temp;
+    }
+    return sample.slice(length - n);
   };
 
   // Sort the object's values by a criterion produced by an iteratee.
--- a/test/.eslintrc
+++ b/test/.eslintrc
@@ -12,6 +12,7 @@
     "strictEqual": false,
     "notStrictEqual": false,
     "notEqual": false,
+    "notDeepEqual": false,
     "throws": false,
     "asyncTest": false,
     "start": false,


marekventur:
I think I made have made a mistake here, please give me a second while I investigate.

marekventur:
Turns out you can't just take the loop from ```_.shuffle``` and exit early, it's not the type of fisher-yates I remember (not even sure it's one, but I'm no expert). 

Anyway, I've rewritten the function and added a test to check it's actually sampling from the whole range rather than just from the first ```n``` elements.

underscore.js
```diff
@@ -366,12 +366,23 @@
...
+  // Works similar to `shuffle` but exits early to be more efficient.
   _.sample = function(obj, n, guard) {
     if (n == null || guard) {
       if (!isArrayLike(obj)) obj = _.values(obj);
       return obj[_.random(obj.length - 1)];
     }
-    return _.shuffle(obj).slice(0, Math.max(0, n));
+
+    var set = isArrayLike(obj) ? obj : _.values(obj);
+    n = Math.max(Math.min(n, set.length), 0);
```
megawac: `set.length` shouldn't ever be less than 0 and if it is it won't really matter

underscore.js
```diff
@@ -366,12 +366,23 @@
...
+    var sample = _.clone(set);
```
megawac: Why clone it? Just construct a new array

megawac:
If this is done `shuffle` should probably be implemented via `sample`

underscore.js
jridgewell: You'll need to use `getLength(set)`, to avoid iOS JIT.

underscore.js
```diff
@@ -366,12 +366,23 @@
...
+    for (var length = set.length, index = length - 1, rand, temp; index > (length - n - 1); index--) {
```
jridgewell: Do you mind moving `rand` and `temp` out of here?

marekventur:
Thanks for the review. I've changed the bits mentioned and made ```shuffle``` call through to ```sample```. I've also did a few minor changes to the test for ```shuffle``` to make it look a bit cleaner and to test for one naive failure case.

I wasn't sure what your policy on git commits is, so I squashed the whole thing together into one commit.

underscore.js
michaelficarra: You still need to use max because `n` may be negative.

underscore.js
```diff
@@ -371,7 +363,19 @@
       if (!isArrayLike(obj)) obj = _.values(obj);
       return obj[_.random(obj.length - 1)];
     }
-    return _.shuffle(obj).slice(0, Math.max(0, n));
+
+    var set = isArrayLike(obj) ? obj : _.values(obj);
+    var length = getLength(set);
+    n = Math.min(n, getLength(set));
+    var sample = _.clone(set);
```
jridgewell: One last nitpick: I think we can use a [modified version](http://en.wikipedia.org/wiki/Reservoir_sampling#Relation_to_Fisher-Yates_shuffle) of Fisher-Yates to only allocate an array of `n`.

```javascript
    var set = isArrayLike(obj) ? obj : _.values(obj);
    var length = getLength(set);
    n = Math.min(n, length);
    var sample = Array(n);
    var rand;
    for (var index = 0; index < n; index++) {
      rand = _.random(0, index);
      if (rand !== index) sample[index] = sample[rand];
      sample[rand] = set[index];
    }
      rand = _.random(0, index);
      if (rand < n) sample[rand] = set[index];
    }
    return sample;
```

marekventur: sounds good. I've added another test to cater for the n<0 case as well.

underscore.js
```diff
@@ -349,21 +349,13 @@
...
-    var set = isArrayLike(obj) ? obj : _.values(obj);
-    var length = set.length;
-    var shuffled = Array(length);
-    for (var index = 0, rand; index < length; index++) {
-      rand = _.random(0, index);
-      if (rand !== index) shuffled[index] = shuffled[rand];
-      shuffled[rand] = set[index];
-    }
-    return shuffled;
+    return _.sample(obj, _.size(obj));
```
jridgewell: Ok, one more :wink:: Using `Infinity` instead of `_.size(obj)` will accomplish the same thing without a potentially costly `_.keys`.

marekventur: Neat, hadn't even thought about that!

megawac:
Have you done any profiling yet @marekventur?

test/collections.js
```diff
@@ -737,12 +737,12 @@
...
-    var numbers = _.range(10);
+    deepEqual(_.shuffle([1]), [1], 'behaves correctly on size 1 arrays');
+    var numbers = _.range(20);
     var shuffled = _.shuffle(numbers);
+    notDeepEqual(numbers, shuffled, 'does change the order'); // Chance of false negative: 1 in ~2.4*10^18
     notStrictEqual(numbers, shuffled, 'original object is unmodified');
-    ok(_.every(_.range(10), function() { //appears consistent?
-      return _.every(numbers, _.partial(_.contains, numbers));
-    }), 'contains the same members before and after shuffle');
+    deepEqual(numbers, _.sortBy(shuffled, function(n) { return n; }), 'contains the same members before and after shuffle');
```
megawac: `_.sortBy(shuffled)` will work just as well

marekventur:
Good call. This is how long 1M operations on a 1000 elements array take on my machine:
```
Master
input = _.range(1000);
1000000x _.shuffle(input):      15321ms
1000000x _.sample(input, 1000): 32637ms
1000000x _.sample(input, 500):  25765ms
1000000x _.sample(input, 2):    16937ms

My version (before https://github.com/jashkenas/underscore/pull/2158#discussion_r29190358)
input = _.range(1000);
1000000x _.shuffle(input):      15931ms
1000000x _.sample(input, 1000): 15326ms
1000000x _.sample(input, 500):   8331ms
1000000x _.sample(input, 2):     1127ms

After https://github.com/jashkenas/underscore/pull/2158#discussion_r29190358:
input = _.range(1000);
1000000x _.shuffle(input):      15896ms
1000000x _.sample(input, 1000): 15310ms
1000000x _.sample(input, 500):  17067ms
1000000x _.sample(input, 2):    10645ms

```
profiling code is here: https://gist.github.com/marekventur/455c7bffb00d9759ec7e

In lights of those numbers I'm going to revert back to my code pre the changes @jridgewell suggested, it might be more efficient in terms of memory usage, but it's slower.

test/collections.js
marekventur: I didn't know that. Is that documented behaviour or a side-effect?

megawac: Documented from iteratee

marekventur: Ok, I'll drop that function.

jridgewell:
I don't believe 92a9d2c is a proper shuffle, that's why my suggestions run slower.

```javascript
// Let's return predictable results
_.random = function (min, max) {
  if (max == null) {
    max = min;
    min = 0;
  }
  return min + Math.floor(Math.min(Math.random(), .1) * (max - min + 1));
}

var array = _.range(10);

// master
_.shuffle(array); // => [8, 9, 1, 2, 3, 4, 5, 6, 7, 0];

// Yours
_.shuffle(array); // => [9, 2, 3, 4, 5, 6, 7, 8, 0, 1];

// Mine
_.shuffle(array); // => [8, 9, 1, 2, 3, 4, 5, 6, 7, 0];
```

It's not definitive proof, but it feels fishy to me.

underscore.js
```diff
@@ -371,7 +363,19 @@
...
+    n = Math.max(Math.min(n, getLength(set)), 0);
+    var sample = _.clone(set);
+    var rand, temp;
+    for (var index = length - 1; index > length - n - 1; index--) {
+      rand = _.random(0, index);
+      temp = sample[index];
+      sample[index] = sample[rand];
+      sample[rand] = temp;
+    }
+    return sample.slice(length - n);
```
megawac: Killing the elements after `n` should be faster

```js
sample.length = n;
```

underscore.js
```diff
@@ -371,7 +363,19 @@
       if (!isArrayLike(obj)) obj = _.values(obj);
       return obj[_.random(obj.length - 1)];
     }
-    return _.shuffle(obj).slice(0, Math.max(0, n));
+
+    var set = isArrayLike(obj) ? obj : _.values(obj);
+    var length = getLength(set);
+    n = Math.max(Math.min(n, getLength(set)), 0);
+    var sample = _.clone(set);
```
megawac: Can you write this `slice.call(set)`

underscore.js
```diff
@@ -371,7 +363,19 @@
...
+    var rand, temp;
+    for (var index = length - 1; index > length - n - 1; index--) {
```
megawac: Can we write this

```js
var index = length, fin = length - n - 1;
while (--index > fin) {
```

underscore.js
megawac: Actually can we not clone the set and use `push` instead?

megawac:
> In lights of those numbers I'm going to revert back to my code pre the changes @jridgewell suggested, it might be more efficient in terms of memory usage, but it's slower.

I think it should be comparable if it is implemented like this https://github.com/megawac/lodash/commit/d7d1a3d3eea811c0211f3bc3a9fcb1c5365ebdf9

underscore.js
jridgewell: The problem is, the shuffled elements are at the end of the array not the beginning.

marekventur: no, that wouldn't work. You want the end of the array, not the beginning.

underscore.js
marekventur: sure, but why not use a for loop for that? Iterating between two integers is the textbook example for when to use ```for``` instead of ```while```.

jridgewell:
> I think it should be comparable/better if it is implemented like this megawac/lodash@d7d1a3d

Putting an `if` in the hot loop will make it slower, no? http://jsperf.com/samples-clone-vs-for v http://jsperf.com/samples-clone-vs-for/5.

His is so much faster because it only runs `n` iterations (with `length` space), whereas ours does the opposite.

marekventur:
Like @jridgewell says, megawac/lodash@d7d1a3d still runs ```input.length``` iterations, right? Avoiding that is the whole point of this PR :)

megawac:
Right, my bad I misread earlier comments

jridgewell:
> Like @jridgewell says, megawac/lodash@d7d1a3d still runs input.length iterations, right? Avoiding that is the whole point of this PR :)

Well, before we were running `length` iterations and using `length` space, the worst of both worlds. I think yours is good enough since it's no worse space-wise and _so_ much faster.

marekventur:
Is there anything else that needs doing before it can be merged? I'm happy to do it, I just don't know what it is :)

jashkenas:
Thanks for contributing!

Issue #2159, opened 2015-04-22
Example of _.unzip on _.homepage wrong?


smartmuki:
_.unzip(["moe", 30, true], ["larry", 40, false], ["curly", 50, false]);
=> [['moe', 'larry', 'curly'], [30, 40, 50], [true, false, false]]

This is actually the output of zip. 
Unzip returns : 
=> [['m', undefined, undefined], ['o', undefined, undefined], ['e', undefined, undefined]]

The test was run on  Underscore.js 1.8.3 

Thanks.

carltonf:
I think it's a homepage typo.

PR #2160, opened 2015-04-22
Fix array typo in home page

carltonf:
_.unzip accepts a single array of arrays.

As smartmuki found in github issue #2159.


akre54:
Thanks.

PR #2161, opened 2015-04-23
described correct behavior of _.result in docs

smelnikov:
`_.result({foo: undefined}, 'foo', 'bar') //bar`

The behavior described in the _.result docs specifies that the defaultValue (if provided) will be returned if an objects property does not exists. Though the implementation will also return the default value when the property exists but has the value `undefined`. Changed docs to reflect this.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1930,8 +1930,8 @@ _.unescape('Curly, Larry &amp;amp; Moe');
         <br />
         If the value of the named <b>property</b> is a function then invoke it
         with the <b>object</b> as context; otherwise, return it. If a default value
-        is provided and the property doesn't exist than the default will be returned.
-        If <tt>defaultValue</tt> is a function its result will be returned.
+        is provided and the property doesn't exist or is undefined than the default
+        will be returned. If <tt>defaultValue</tt> is a function its result will be returned.
       </p>
       <pre>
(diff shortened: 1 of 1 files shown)


michaelficarra:
Instead of changing the documentation to match the implementation, shouldn't we be changing the implementation to match the documentation?

megawac:
Agreed this is an improvement

jridgewell:
I think we can open another PR if we want to change the implementation (and now, documentation). For now, I agree that this is an improvement.

index.html
```diff
@@ -1930,8 +1930,8 @@ <h2 id="utility">Utility Functions</h2>
         <br />
         If the value of the named <b>property</b> is a function then invoke it
         with the <b>object</b> as context; otherwise, return it. If a default value
-        is provided and the property doesn't exist than the default will be returned.
-        If <tt>defaultValue</tt> is a function its result will be returned.
+        is provided and the property doesn't exist or is undefined than the default
```
michaelficarra: ", then", not " than".

jridgewell: Fixed by https://github.com/jashkenas/underscore/commit/2433b0c1d1998271a755ad366d85fd5191f6cebc.

PR #2162, opened 2015-04-23
Fix documents for zip and unzip.

carltonf:
Mismatch with code. And some docs seem to be misplaced.

### DIFF
--- a/index.html
+++ b/index.html
@@ -957,8 +957,9 @@ _.uniq([1, 2, 1, 4, 1, 3]);
         Merges together the values of each of the <b>arrays</b> with the
         values at the corresponding position. Useful when you have separate
         data sources that are coordinated through matching array indexes.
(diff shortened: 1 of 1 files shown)


carltonf:
A follow up to #2159.

index.html
```diff
@@ -957,8 +957,9 @@ <h2 id="arrays">Array Functions</h2>
         Merges together the values of each of the <b>arrays</b> with the
         values at the corresponding position. Useful when you have separate
         data sources that are coordinated through matching array indexes.
-        If you're working with a matrix of nested arrays, <tt>_.zip.apply</tt>
-        can transpose the matrix in a similar fashion.
+        Use with <tt>apply</tt> to pass in an array of arrays.
```
akre54: "use with `apply`" isn't really clear here. Can we revert this back to `_.zip.apply`?

PR #2167, opened 2015-05-06
Add _.isNodeList, support for NodeLists in _.difference

### LINKED ISSUE #2156: Difference of two NodeLists returns every element
The _.difference method should return an array containing only elements which are present in the first array but not in any of the subsequent array.  If you pass it 2 NodeLists it always returns all the elements in the first list regardless of which elements appear in both.

```javascript
    divs = document.querySelectorAll("div");
   _.difference(divs, divs);
```

returns an array which contains every element from 'divs'.  It would be super if difference could handle NodeLists as if they were arrays, as it's a useful case. The core logic of the function works on NodeLists so

```javascript
_.filter(divs, function(el){ return !_.contains(divs, el); });
```

returns the expected result.

PatrickStankard:
Closes #2156

### DIFF
--- a/test/objects.js
+++ b/test/objects.js
@@ -4,6 +4,7 @@
   QUnit.module('Objects');
 
   var testElement = typeof document === 'object' ? document.createElement('div') : void 0;
+  var testNodeList = typeof document === 'object' ? document.querySelectorAll('div') : void 0;
 
   test('keys', function() {
     deepEqual(_.keys({one : 1, two : 2}), ['one', 'two'], 'can extract the keys from an object');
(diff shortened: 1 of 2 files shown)


test/objects.js
```diff
@@ -4,6 +4,7 @@
   QUnit.module('Objects');
 
   var testElement = typeof document === 'object' ? document.createElement('div') : void 0;
+  var testNodeList = typeof document === 'object' ? document.querySelectorAll('div') : void 0;
```
megawac: `querySelector` isn't the best choice as it won't be available in older browsers. Try `document.getElementsByTagName`

PatrickStankard:
Since a `NodeList` is a host object, and [Underscore isn't for host objects](https://github.com/jashkenas/underscore/issues/1138#issuecomment-19110640), I'm just going to close this

PatrickStankard:
@jdalton: I also noted that you could just `_.toArray()` a `NodeList` before using it with an array or collection function in https://github.com/jashkenas/underscore/issues/2156#issuecomment-101385472, which was the inspiration for this

PR #2170, opened 2015-05-07
Reword comment to not use "require"

zertosh:
Longshot I know...

browserify does a [quick](https://github.com/substack/node-detective/blob/75946ee23cb7abb974a2053235ecb2663b879c07/index.js#L4) regexp match for `\brequire\b` before deciding on whether to build an AST to get the require calls. By rewording this comment so that the word "require" is nowhere in the source, you'd be improving everyones build times. Browserify provides a `noparse` escape hatch but this is an easy win.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -42,7 +42,7 @@
   };
 
   // Export the Underscore object for **Node.js**, with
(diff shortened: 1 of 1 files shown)


megawac:
Cool stuff and good to know

PR #2176, opened 2015-05-13
Resolve: Uncaught TypeError: Cannot read property '_' of undefined

JBerendes:
resolve: Uncaught TypeError: Cannot read property '_' of undefined

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -11,8 +11,10 @@
   // Establish the root object, `window` in the browser, or `exports` on the server.
   var root = this;
 
+  if(root._) {
   // Save the previous value of the `_` variable.
-  var previousUnderscore = root._;
+    var previousUnderscore = root._;
+  }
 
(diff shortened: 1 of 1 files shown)


jridgewell:
Hi @JBerendes, this PR isn't going to fix a `read property of undefined` error. How are you including underscore?

JBerendes:
Using React 0.13.3 with Browserify and Babelify (ES6+ to ES5 friendly conversion)
The code is combined into one file for a static documentation site.
in the source of one my components is the line

When opening the app Chrome then throws the error as described.

jridgewell:
Ah, dupe of https://github.com/jashkenas/underscore/issues/2152 then. Right now, underscore isn't compatible with ES6 modules.

JBerendes:
Alright I'll come up with a different solution for what I need. Thanks.

PR #2178, opened 2015-05-15
Use restArgs in flatten functions

jridgewell:
http://jsperf.com/restargs-flatten

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -488,9 +488,9 @@
   };
 
   // Internal implementation of a recursive `flatten` function.
-  var flatten = function(input, shallow, strict, startIndex) {
+  var flatten = function(input, shallow, strict) {
     var output = [], idx = 0;
-    for (var i = startIndex || 0, length = getLength(input); i < length; i++) {
+    for (var i = 0, length = getLength(input); i < length; i++) {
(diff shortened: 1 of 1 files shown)


akre54:
The only win looks like it's in `keys`. The others look about the same or slower, no?

jridgewell:
Everything the same or faster (with the exception of `_.pick` on Safari, which is slower).

I just prefer the style of using restArgs so the code shows the "normal" case. `union(...arrays)`, `difference(array, ...others)`, `pick(obj, ...keys)` reads better.

akre54:
I disagree. I think these kinds of changes hurt readability for negligible gains. I wouldn't bet against the future of optimized JS engines and I certainly don't think the mental overhead and extra indirection of these helper functions is worth it.

jashkenas:
The readability is hurt where we start peeking into `rest[1]` for what previously was a `context` variable, and things like that. Possible to avoid?

jdalton:
> The readability is hurt where we start peeking into `rest[1]` for what previously was a `context` variable, and things like that. Possible to avoid?

I think in the context of overloaded API the readability issue is a wash as it's just favoring the other supported argument signature.

PR #2179, opened 2015-05-16
optimized isEqual function, 30x faster on integers

gaperton:
and 1,5 times faster than lodash.isEqual
http://jsperf.com/underscore-isequal/3

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -1203,7 +1203,10 @@
 
   // Perform a deep comparison to check if two objects are equal.
   _.isEqual = function(a, b) {
-    return eq(a, b);
+      var typeA = typeof a;
+      if( typeA != typeof b ) return false;
+      if( typeA == 'object' ) return a && b ? eq( a, b ) : a === b;
+      return a === 0 || a != a ? eq(a, b) : a === b;
   };
 
   // Is a given array, string, or object empty?
--- a/underscore.js
+++ b/underscore.js
@@ -1110,11 +1110,6 @@
 
   // Internal recursive comparison function for `isEqual`.
   var eq = function(a, b, aStack, bStack) {
-    // Identical objects are equal. `0 === -0`, but they aren't identical.
-    // See the [Harmony `egal` proposal](http://wiki.ecmascript.org/doku.php?id=harmony:egal).
-    if (a === b) return a !== 0 || 1 / a === 1 / b;
-    // A strict comparison is necessary because `null == undefined`.
-    if (a == null || b == null) return a === b;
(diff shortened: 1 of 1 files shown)


gaperton:
Some problems in unit tests. Next time then.

underscore.js
```diff
@@ -1203,7 +1198,13 @@
...
   // Perform a deep comparison to check if two objects are equal.
   _.isEqual = function(a, b) {
-    return eq(a, b);
+      // Identical objects are equal. `0 === -0`, but they aren't identical.
+      // See the [Harmony `egal` proposal](http://wiki.ecmascript.org/doku.php?id=harmony:egal).
+      if (a === b) return a !== 0 || 1 / a === 1 / b;
+      var typeA = typeof a;
+      if( typeA != typeof b ) return false;
+      if( typeA == 'object' ) return a && b ? eq( a, b ) : a === b;
+      return a != a ? b != b : a === b;
```
megawac: You can't move this code so `[NaN] == [NaN]`

underscore.js
```diff
@@ -1110,11 +1110,6 @@
 
   // Internal recursive comparison function for `isEqual`.
   var eq = function(a, b, aStack, bStack) {
-    // Identical objects are equal. `0 === -0`, but they aren't identical.
-    // See the [Harmony `egal` proposal](http://wiki.ecmascript.org/doku.php?id=harmony:egal).
-    if (a === b) return a !== 0 || 1 / a === 1 / b;
-    // A strict comparison is necessary because `null == undefined`.
-    if (a == null || b == null) return a === b;
```
megawac: Sorry I meant this code

underscore.js
```diff
@@ -1203,7 +1198,13 @@
 
   // Perform a deep comparison to check if two objects are equal.
   _.isEqual = function(a, b) {
-    return eq(a, b);
+      // Identical objects are equal. `0 === -0`, but they aren't identical.
+      // See the [Harmony `egal` proposal](http://wiki.ecmascript.org/doku.php?id=harmony:egal).
+      if (a === b) return a !== 0 || 1 / a === 1 / b;
+      var typeA = typeof a;
+      if( typeA != typeof b ) return false;
```
jridgewell: Unfortunately, types aren't enough since `_.isEqual('test', new String('test')) === true`.

gaperton:
Unfortunately, this code doesn't work at all breaking underscore unit tests, so I've prepared different implementation. Which appears to be correct, and it's faster.

https://github.com/jashkenas/underscore/pull/2180

PR #2180, opened 2015-05-17
isEqual performance optimization

gaperton:
Optimized isEqual for speed on primitive types. Now it's multiple times faster than lodash in all browsers.

Performance of isEqual is critical for Backbone.Model.set performance.

http://jsperf.com/underscore-isequal/5

Integer comparison:
Chrome: 70x faster 
Firefox: 50x faster
Safari: 140x faster

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -1113,7 +1113,7 @@
     // Identical objects are equal. `0 === -0`, but they aren't identical.
     // See the [Harmony `egal` proposal](http://wiki.ecmascript.org/doku.php?id=harmony:egal).
     if (a === b) return a !== 0 || 1 / a === 1 / b;
-    if( typeof a != 'object' && typeof b != 'object' ) return a != a ? b != b : a === b;
+    if( typeof a != 'object' && typeof b != 'object' ) return a != a && b != b;
 
     if (a == null || b == null) return a === b;
 
--- a/underscore.js
+++ b/underscore.js
@@ -1114,8 +1114,7 @@
     // See the [Harmony `egal` proposal](http://wiki.ecmascript.org/doku.php?id=harmony:egal).
     if (a === b) return a !== 0 || 1 / a === 1 / b;
     if( typeof a != 'object' && typeof b != 'object' ) return a != a && b != b;
-
-    if (a == null || b == null) return a === b;
+    if (a == null || b == null) return false;
 
     // Unwrap any wrapped objects.
     if (a instanceof _) a = a._wrapped;
@@ -1205,7 +1204,9 @@
 
   // Perform a deep comparison to check if two objects are equal.
   _.isEqual = function(a, b) {
-      return eq( a, b );
+      if (a === b) return a !== 0 || 1 / a === 1 / b;
+      if( typeof a != 'object' && typeof b != 'object' ) return a != a && b != b;
+      return a != null && b != null && eq( a, b );
   };
 
   // Is a given array, string, or object empty?
--- a/underscore.js
+++ b/underscore.js
@@ -1210,7 +1210,9 @@
 
   // Perform a deep comparison to check if two objects are equal.
   _.isEqual = function(a, b) {
-      return eq( a, b );
+      if (a === b) return a !== 0 || 1 / a === 1 / b;
+      if( typeof a != 'object' && typeof b != 'object' ) return a != a && b != b;
+      return a != null && b != null && deepEq( a, b );
   };
 
   // Is a given array, string, or object empty?
--- a/underscore.js
+++ b/underscore.js
@@ -1110,15 +1110,10 @@
 
   // Internal recursive comparison function for `isEqual`.
   var deepEq, eq = function(a, b, aStack, bStack) {
-      // Identical objects are equal. `0 === -0`, but they aren't identical.
-      // See the [Harmony `egal` proposal](http://wiki.ecmascript.org/doku.php?id=harmony:egal).
       if (a === b) return a !== 0 || 1 / a === 1 / b;
-
-      // if objects are of primitive types, they are not equal unless both of them are NaNs.
-      if (typeof a != 'object' && typeof b != 'object') return a != a && b != b;
-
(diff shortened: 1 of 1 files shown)


underscore.js
```diff
@@ -1203,7 +1210,9 @@
 
   // Perform a deep comparison to check if two objects are equal.
   _.isEqual = function(a, b) {
-    return eq(a, b);
+      if (a === b) return a !== 0 || 1 / a === 1 / b;
+      if (typeof a != 'object' && typeof b != 'object') return a != a && b != b;
+      return a != null && b != null && deepEq(a, b);
```
gaperton: You might wonder why eq function is inlined here.

Answer is that it gives 25-30% more performance in firefox, and about 10-15% in Chrome.

underscore.js
```diff
@@ -1109,12 +1109,19 @@
 
 
   // Internal recursive comparison function for `isEqual`.
-  var eq = function(a, b, aStack, bStack) {
-    // Identical objects are equal. `0 === -0`, but they aren't identical.
-    // See the [Harmony `egal` proposal](http://wiki.ecmascript.org/doku.php?id=harmony:egal).
-    if (a === b) return a !== 0 || 1 / a === 1 / b;
-    // A strict comparison is necessary because `null == undefined`.
-    if (a == null || b == null) return a === b;
```
gaperton: It's easy to see that a === b would never be true here. If it would, it would go through return in 1115.

underscore.js
```diff
@@ -1109,12 +1109,19 @@
...
+  var deepEq, eq = function(a, b, aStack, bStack) {
+      // Identical objects are equal. `0 === -0`, but they aren't identical.
+      // See the [Harmony `egal` proposal](http://wiki.ecmascript.org/doku.php?id=harmony:egal).
+      if (a === b) return a !== 0 || 1 / a === 1 / b;
+
+      // if objects are of primitive types, they are not equal unless both of them are NaNs.
+      if (typeof a != 'object' && typeof b != 'object') return a != a && b != b;
```
gaperton: Values are not objects, and they are not strictly equal. Only option they could be equal is that they are NaNs.

underscore.js
```diff
@@ -1109,12 +1109,19 @@
...
+
+      // if any of a or b is null, objects are not equal.
+      return a != null && b != null && deepEq(a, b, aStack, bStack);
+  };
+
+  deepEq = function(a, b, aStack, bStack) {
```
gaperton: eq is recursive function, so all the logic we create must be recursive too. Also, splitting this function into two dramatically improved performance in Firefox JIT. It was slower than lodash before this change.

underscore.js
megawac: Why is this not implemented as `eq(a, b)` looks like a lot of redundant code

underscore.js
```diff
@@ -1109,12 +1109,19 @@
 
 
   // Internal recursive comparison function for `isEqual`.
-  var eq = function(a, b, aStack, bStack) {
-    // Identical objects are equal. `0 === -0`, but they aren't identical.
-    // See the [Harmony `egal` proposal](http://wiki.ecmascript.org/doku.php?id=harmony:egal).
-    if (a === b) return a !== 0 || 1 / a === 1 / b;
-    // A strict comparison is necessary because `null == undefined`.
-    if (a == null || b == null) return a === b;
+  var deepEq, eq = function(a, b, aStack, bStack) {
```
megawac: is `eq` actually used?

underscore.js
gaperton: Intentionally redundant code, which makes firefox significantly faster. I have implemented it in eq initially, and measured the result. Was not good in FF.

underscore.js
gaperton: Yes, it is. It's called from deepEq. Again, splitting eq by two functions significantly improves JIT performance, as well as inlining eq directly to _.isEqual.

This is completely unintuitive, but it's how JIT actually works. I measured every change. You could try yourself.

jridgewell:
I think this is getting more complicated than is needed. @megawac's original code is able to get the [same performance](rawgit.com/megawac/underscore/eq-primitives/underscore.js) with common usage without another function or duplicating all the checks already in `eq()`. https://github.com/jashkenas/underscore/compare/master...jridgewell:isEqual

gaperton:
@jridgewell, in fact, it's quite oppoiste. This is an only line of additional logic this change adds, all other are optimizations of existing code for modern browser's JIT.

```
if (typeof a != 'object' && typeof b != 'object') return a != a && b != b;
```

And it's way faster in case if you compare objects too, cause it's eq function optimization, not a hack on a top of it, as "original  change" is.

gaperton:
Hey. Looks like lodash guys are already testing the fix, improving performance to match this function. :)

http://jsperf.com/underscore-isequal/6

That's enough. I'm switching to lodash. Bye.

gaperton:
Ok, that was my last commit. Now you have one of the fastest recursive deepEq implementations currently available. Unfortunately, I have no time to deal with that issue any more.

jashkenas:
Should this be closed then?

underscore.js
```diff
@@ -1203,7 +1205,10 @@
 
   // Perform a deep comparison to check if two objects are equal.
   _.isEqual = function(a, b) {
-    return eq(a, b);
+      if (a === b) return a !== 0 || 1 / a === 1 / b;
+      var typeA = typeof a;
+      return typeA != 'function' && typeA != 'object' && typeof b != 'object' ?
+        a != a && b != b : a != null && b != null &&  deepEq(a, b);
```
jridgewell: This is going to fail `_.isEqual(null, null)`.

jdalton:
Yap. Micro-opt is micro-opt.

underscore.js
gaperton: Of course it's not. _.isEqual(null, null) will go through 1208, which in fact is the first line of the original eq function unchanged.

Sometimes it's so hard to find a valid reason to close PR, right? :)

PR #2183, opened 2015-05-21
Add Composer support

RobLoach:
This adds support for the [Composer](http://getcomposer.org) package manager for PHP.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,25 @@
+{
+  "name": "jashkenas/underscore",
+  "description": "JavaScript's functional programming helper library.",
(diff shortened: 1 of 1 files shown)


michaelficarra:
All of this information is duplicated from `package.json`. Composer should learn how to read it from `package.json` instead of requiring us to maintain a file for every package manager. PRs like this have been denied in the past, so :-1: from me.

akre54:
Underscore is a Javscript library. We shouldn't add a PHP package manager file for the same reasons we shouldn't add a `Gemfile` or a `setup.py`. Use the right tool for the job.

PR #2185, opened 2015-05-22
Optimize _.isEqual for primatives

jridgewell:
#2180 got micro-opty, but we can still speed up `_.isEqual` with acceptable changes.

I'm stealing a page from lodash and splitting `eq()` and `deepEq()`. Yes, I [waffled](https://github.com/jashkenas/underscore/pull/2180#issuecomment-102730045) about creating another function, but we're able to not duplicate checks and get a [30x speed boost](http://jsperf.com/underscore-isequal/9) with primitives.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -1115,6 +1115,15 @@
     if (a === b) return a !== 0 || 1 / a === 1 / b;
     // A strict comparison is necessary because `null == undefined`.
     if (a == null || b == null) return a === b;
+    // `NaN`s are equivalent, but non-reflexive.
+    if (a !== a) return b !== b;
+    // Exhaust primitive checks
+    var type = typeof a;
+    if (type !== 'function' && type !== 'object' && typeof b !== 'object') return false;
+    return deepEq(a, b, aStack, bStack);
+  };
+
+  var deepEq = function(a, b, aStack, bStack) {
     // Unwrap any wrapped objects.
     if (a instanceof _) a = a._wrapped;
     if (b instanceof _) b = b._wrapped;


jdalton:
Nice! I broke the method up into more focused parts because it made it easier to digest.

underscore.js
```diff
@@ -1115,6 +1115,15 @@
     if (a === b) return a !== 0 || 1 / a === 1 / b;
     // A strict comparison is necessary because `null == undefined`.
     if (a == null || b == null) return a === b;
+    // `NaN`s are equivalent, but non-reflexive.
+    if (a !== a) return b !== b;
+    // Exhaust primitive checks
+    var type = typeof a;
+    if (type !== 'function' && type !== 'object' && typeof b !== 'object') return false;
```
jdalton: Are you missing the other `typeof b` for `!== 'function'` ?

jdalton: In ld I simplified the check to:
```js
if (value == null || other == null || (!isObject(value) && !isObject(other))) {
  return value !== value && other !== other;
}
```

which squishes the `NaN` and nullish checks into 1 condition.

jridgewell: I'm unsure if it's needed:

| type a | type b | equals |
| -------- |  -------- |  -------- | 
| primative | primative | primitive checks |
| primative | function | false |
| primative  | object |  deepEq |
| function | primative | false |
| function | function | deepEq |
| function | object | deepEq |
| object | primative | deepEq |
| object | function | false |
| object | object | deepEq |

There's a difference between `(function, object)` and `(object, function)`, but I think both those could be handled by primitive checks only? It's was just easier to check for `type === 'function'` for the `(function, function)` case (Safari regexs).

jdalton: > I'm unsure if it's needed:

It just misses the early exit. Could be tackled and simplified with `_.isObject` use.

jridgewell: Which? I think it actually adds an early exit (`(primitive, function)`).

I could be more verbose:
```js
    // Exhaust primitive checks
    var typea = typeof a;
    var typeb = typeof b;
    var eitherObjects = typea === 'object' || typeb === 'object';
    var eitherFunctions = typea === 'function' || typeb === 'function';
    var bothFunctions = typea === 'function' && typeb === 'function';
    if (eitherObjects && eitherFunctions) return false;
    if (!eitherObjects && !bothFunctions) return false;
    return deepEq(a, b, aStack, bStack);
```

Which would add all possible early exits:

| type a | type b | equals |
| -------- |  -------- |  -------- | 
| primitive | primitive | primitive checks |
| primitive | function | false |
| primitive  | object |  deepEq |
| function | primitive | **false** |
| function | function | deepEq |
| function | object | **false** |
| object | primitive | deepEq |
| object | function | false |
| object | object | deepEq |

mqyqingfeng:
I am curious why the current code is 

```js
if (type !== 'function' && type !== 'object' && typeof b != 'object') return false;
```

why `type` use `!==` and `type b` use `!=` ?

PR #2187, opened 2015-05-24
Feature remove item

1984weed:
```_.removeItem``` method. It would remove item from array if predicate is true.

### CODE REVIEW

**Comment by megawac** [comment]
This is a mixture of `filter` and `without` that only supports arrays (or context). I don't think theres demand for this but thanks anyway. P.s. you can mix this in like

```js
_.removeItem = function(array, item){
   var method = _.isFunction(item) ? _.filter : _.without
   return method(array, item);
}
```

PR #2192, opened 2015-05-25
Added a _.reverse array method to underscore.js.

kengz:
Added a _.reverse (aliased _.flip) array method to underscore.js – reverses the first **n** entries.
Uses while(shift / unshift) for fast performance in V8 engine.
Example usage:
_.reverse([0, 1, 2, 3, 4])
=> [4, 3, 2, 1, 0]
_.reverse([0, 1, 2, 3, 4], 3)
=> [2, 1, 0, 3, 4]

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -483,6 +483,21 @@
   _.rest = _.tail = _.drop = function(array, n, guard) {
     return slice.call(array, n == null || guard ? 1 : n);
   };
+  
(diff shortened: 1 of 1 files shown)


ElegantSudo:
Why isn't this official?

ElegantSudo:
Ah, I see. We overlooked the prototypal method on all arrays.

PR #2193, opened 2015-05-25
Added a _.reverse array method with a variable argument.

kengz:
Added a `_.reverse` (aliased `_.flip`) array method to *underscore.js*; and the unit tests to *test/array.js*.
Returns a copy of the reversed array. Aliased as `flip`.
Passing an **n** will return a copy of array with the first N entries reversed.
Example usage:
```
_.reverse([0, 1, 2, 3, 4])
=> [4, 3, 2, 1, 0]
_.reverse([0, 1, 2, 3, 4], 3)
=> [2, 1, 0, 3, 4]
```

### DIFF
--- a/test/arrays.js
+++ b/test/arrays.js
@@ -41,6 +41,17 @@
     deepEqual(result, [2, 3, 4], 'works on arguments object');
   });
 
+  test('reverse', function() {
+    var numbers = [0, 1, 2, 3 ,4];
+    deepEqual(_.reverse(numbers), [4, 3, 2, 1, 0], 'working reverse()');
+    deepEqual(_.reverse(numbers, -1), [0, 1, 2, 3 ,4], 'working reverse(-1)');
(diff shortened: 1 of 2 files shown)


michaelficarra:
When would someone want to reverse a portion of an array?

kengz:
@michaelficarra not very frequently; most of the times just the entire array.
I used it in time series analysis, where the tail of an array has some other entries – bad data design on the provider's part, so just to be able to deal with that. I agree with you – it will not be the most frequently used function, but can be there when needed.
The example was:
```
[t1, t2, t3, ..., tn, bloodtype, bloodpressure1, ... ] //raw data
```
Where I wanted to reverse the time series at the front to feed to an analyzer, yet I can't discard the array or take the effort to manually split it, since the time series can be of varying length.

jridgewell:
> not very frequently; most of the times just the entire array.

I think this is better left as a mixin, then. JS already provides the base case (`array.reverse()`, or `array.slice().reverse()` for immutable) without any browser incompatibilities.

PR #2194, opened 2015-05-26
upgrade eslint version and fix all current warnings/errors

michaelficarra:


jashkenas:
Whew. Nice one!

michaelficarra:
This took nearly 4 hours. It was so tedious!

jdalton:
For the most part I find stuff like this is largely a unnecessary. Does it stop bad releases? Not really. Does it add noise & stumbling blocks to PRs? Yeah.

jashkenas:
I agree, but now that it's done — don't you think it's a net positive to have?

jdalton:
Naw, the PR hurdles alone make it a negative for me. I see this as a crutch to brace maintainer activity.

braddunbar:
Maybe it doesn't directly stop bad releases but I disagree that's it's unnecessary. Consistency of style and use of tools to prevent trivial bugs is one of the first things I use to gauge the health of a project and, consequently, my willingness and desire to contribute to it.

That said, I don't care if we run it on pulls or block them because it fails, if that's what you mean. Accepting good pulls with different style and not cleaning up afterwards are two different things.

jridgewell:
Agh, it's always my code that has sneaky 4 spaces. :frowning: 

Why do we align colons, but not equals?

```diff
-    nativeBind         = FuncProto.bind,
-    nativeCreate       = Object.create;
+    nativeIsArray = Array.isArray,
+    nativeKeys = Object.keys,

-        value: value,
-        index: index,
+        value   : value,
+        index   : index,
         criteria: iteratee(value, index, list)
```

michaelficarra:
@jridgewell There were many more `:` alignments already in the codebase than `=` alignments. I just tried to make the rules enforce current convention.

jashkenas:
We shouldn't be aligning those colons. Colons should be visually attached to the label on the left.

michaelficarra:
@jashkenas done: https://github.com/jashkenas/underscore/pull/2196

PR #2199, opened 2015-05-27
Remove nativeBind usage

jridgewell:
We cover almost all of nativeBind's cases:
- Binding function to context
- Binding function to context with args
- Binding constructor with args

The only downside is supported environments will no longer get the correct `length` property of the bound method. Usually, you subtract the number of bound args from `func`'s length. But, now Underscore is consistent between all environments.

http://jsperf.com/native-bind-vs-underscore/2

Re: #2197

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -17,7 +17,7 @@
   var previousUnderscore = root._;
 
   // Save bytes in the minified (but not gzipped) version:
-  var ArrayProto = Array.prototype, ObjProto = Object.prototype, FuncProto = Function.prototype;
+  var ArrayProto = Array.prototype, ObjProto = Object.prototype;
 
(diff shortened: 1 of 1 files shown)


michaelficarra:
I'm okay with this.

unwitting:
Alternative in #2198 which still uses native bind and therefore would preserve `arguments.length`, but at massive cost compared to this approach.

See: http://jsperf.com/underscore-js-2198-vs-2199 - 2198 was twice as fast as current, but this lodash-style approach blows it out of the water.

megawac:
This is likely breaking for constructor cases. Consider

```js
var D = _.bind(Date, null, 2015, 05)
new D(10);
```

Sure this doesn't work in old environments but it may break some peoples code

/cc @jdalton

PR #2202, opened 2015-05-29
Noticed that flatten would copy the array in every step and fixed it.

erikbern:
If you have a n=1000 deeply nested array with k=1000 elements in the leaf node then it would be copied n times leading to O(n*k). After this fix, it will take O(n+k)

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -490,17 +490,19 @@
   };
 
   // Internal implementation of a recursive `flatten` function.
-  var flatten = function(input, shallow, strict) {
-    var output = [], idx = 0;
+  var flatten = function(input, shallow, strict, _output) {
+    var output = _output ? _output : [];
+    var idx = output.length;
     for (var i = 0, length = getLength(input); i < length; i++) {
       var value = input[i];
       if (isArrayLike(value) && (_.isArray(value) || _.isArguments(value))) {
         //flatten current level of array or arguments object
-        if (!shallow) value = flatten(value, shallow, strict);
-        var j = 0, len = value.length;
-        output.length += len;
-        while (j < len) {
-          output[idx++] = value[j++];
+        if (shallow) {
+          var j = 0, len = value.length;
+          while (j < len) output[idx++] = value[j++];
+        } else {
+          flatten(value, shallow, strict, output);
+          idx = output.length;
         }
       } else if (!strict) {
         output[idx++] = value;
--- a/test/arrays.js
+++ b/test/arrays.js
@@ -98,6 +98,11 @@
     equal(_.flatten([_.range(10), _.range(10), 5, 1, 3]).length, 23);
     equal(_.flatten([new Array(1000000), _.range(56000), 5, 1, 3]).length, 1056003, 'Flatten can handle massive collections');
     equal(_.flatten([new Array(1000000), _.range(56000), 5, 1, 3], true).length, 1056003, 'Flatten can handle massive collections');
+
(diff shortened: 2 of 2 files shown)


erikbern:
Test took 2.5s before fix, 250ms after fix

underscore.js
```diff
@@ -490,17 +490,21 @@
   };
 
   // Internal implementation of a recursive `flatten` function.
-  var flatten = function(input, shallow, strict) {
-    var output = [], idx = 0;
+  var flatten = function(input, shallow, strict, _output, _idx) {
+    var output = _output ? _output : [], idx = _idx ? _idx : 0;
```
megawac: Instead of passing `_idx` can you just make `idx = output.length`. Otherwise LGTM

underscore.js
```diff
@@ -490,17 +490,21 @@
...
     for (var i = 0, length = getLength(input); i < length; i++) {
       var value = input[i];
       if (isArrayLike(value) && (_.isArray(value) || _.isArguments(value))) {
         //flatten current level of array or arguments object
-        if (!shallow) value = flatten(value, shallow, strict);
-        var j = 0, len = value.length;
-        output.length += len;
-        while (j < len) {
-          output[idx++] = value[j++];
+        if (!shallow) {
```
jridgewell: Do you mind swapping the two conditional paths? Reading a `!condition` with an `else` always makes my head hurt.

```patch
if (shallow) {
  var j = 0, len = value.length
  //...
} else {
  flatten(value, shallow, strict, output, idx);
  idx = output.length;
}
```

underscore.js
```diff
@@ -490,17 +490,21 @@
...
+          flatten(value, shallow, strict, output, idx);
+          idx = output.length;
+        } else {
+          var j = 0, len = value.length;
+          output.length += len;
```
jridgewell: Mind dropping this line? It seems we [added](https://github.com/megawac/underscore/commit/63fc82284021b7f7d916b7561d71168935dd60fb#diff-0f36b362a0b81d6f4d4bfd8a7413c75dR476) it a bit prematurely, it's actually a [slowdown](https://github.com/lodash/lodash/pull/1173) in recent node / iojs

erikbern:
Addressed comments by @megawac and @jridgewell – see updated commit

underscore.js
```diff
@@ -490,17 +490,20 @@
   };
 
   // Internal implementation of a recursive `flatten` function.
-  var flatten = function(input, shallow, strict) {
-    var output = [], idx = 0;
+  var flatten = function(input, shallow, strict, _output) {
+    var output = _output ? _output : [];
```
megawac: can you rename the param output and do

output = output || [];

I'll merge this in a bit

underscore.js
```diff
@@ -490,17 +490,20 @@
...
         //flatten current level of array or arguments object
-        if (!shallow) value = flatten(value, shallow, strict);
-        var j = 0, len = value.length;
-        output.length += len;
-        while (j < len) {
-          output[idx++] = value[j++];
+        if (shallow) {
+          var j = 0, len = value.length;
+          while (j < len)
+            output[idx++] = value[j++];
```
jridgewell: One last nitpick: Can you join these lines? `while(j < len) output[idx++] = value[j++]`.

erikbern: fixed in latest

megawac:
Thanks!

Issue #2147, opened 2015-04-10
Supporting ES6 iterables in collection methods


tgriesser:
So this sounds like something lodash isn't going to support yet because it will add quite a bit of code https://github.com/lodash/lodash/issues/737 - but I think it'd be cool to have a ticket here to explore the possibility/implications (code, performance, and otherwise) of supporting ES6 iterables in at least the collection functions of underscore.

As ES6 `Map` / `Set` are available in more places (latest browsers, babel polyfill) it'd be great to be able to iterate in the same way as objects, especially since `Map` it avoids the weird edge cases of treating an object as a Map (`__proto__`, ownProperty checks etc.). Also, the ES6 iterator API provides a standard convention for iterating arbitrary objects and it'd be great if underscore could maybe act as a functional api `for-of` in the same way it does for `for`/`for-in`

I feel like it should be possible to come up with a way that does this in most reasonable cases with out too much overhead, but I admittedly haven't explored it too in-depth - just wanted to see if anyone else would be interested in discussion.

jashkenas:
Would certainly be worth exploring in a pull request. Would it really be so much code? You couldn't put the iteration logic in a single function? ES6 `for of` is atrociously slow as is, so it seems like it would be hard to do worse?

jdalton:
Related https://github.com/lodash/lodash/pull/1420.

jdalton:
`_#next` could be a way to get the ball rolling.

jgonggrijp:
See also discussion in #2808.

PR #2208, opened 2015-06-06
Make _.each/_.forEach support ES6 Set/Map

benjycui:
IMHO, with the approaching of ES6, it is time to support new features. But I am not sure if you want it :-)
See also: #2147.

### DIFF
--- a/test/collections.js
+++ b/test/collections.js
@@ -23,6 +23,20 @@
     deepEqual(answers, ['one', 'two', 'three'], 'iterating over objects works, and ignores the object prototype.');
     delete obj.constructor.prototype.four;
 
+    if (typeof Set == 'function') {
+      answers = [];
+      var set = new Set([1, 2, 3]);
+      _.each(set, function(value){ answers.push(value); });
+      deepEqual(answers, [1, 2, 3], 'iterating over Set works');
+    }
+
+    if (typeof Map == 'function') {
+      answers = [];
+      var map = new Map([['one', 1], ['two', 2], ['three', 3]]);
+      _.each(map, function(value, key){ answers.push(key); });
+      deepEqual(answers, ['one', 'two', 'three'], 'iterating over Map works');
+    }
+
     // ensure the each function is JITed
     _(1000).times(function() { _.each([], function(){}); });
     var count = 0;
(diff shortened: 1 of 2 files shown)


jdalton:
@megawac can you make the coveralls less chatty (there's an option in its config on the coveralls site).

@benjycui I don't think it's time to support new forms of iteration just yet. For example IE 11 doesn't support `new Set([1, 2, 3])` for assigning values to a `Set`. I'm also not a fan of duck typing for iteration or adding more overloaded functionality to methods.

The easiest way to deal with this is in your ES6 code convert your sets/maps to an array with `Array.from`.

jdalton:
Or even:
```js
_.each([...set], function() {});
```

benjycui:
@jdalton yep, you are right. We can convert `Set`/`Map` with `...`.

But for `Map`, this solution is inconsistent with the default behaviour of `_.each`:
> Each invocation of iteratee is called with three arguments: `(element, index, list)`. If list is a JavaScript object, iteratee's arguments will be `(value, key, list)`.

```js
var map = new Map([[1, 'one'], [2, 'two'], [3, 'three']]);
_.each([...map], function (kv) {  // I think this arguments list should be: `(value, key, list)`
  console.log(kv[0], ': ', kv[1]);
});
```
Do you think that consistency is worthy of 2 lines of source code? :-)

jdalton:
> But for `Map`, this solution is inconsistent with the default behaviour of `_.each`:

If you want the familiar arg sig then you can do
`_.each(_.object([...map]), function() {})`

> Do you think that consistency is worthy of 2 lines of source code? :-)

It's not just 2 lines, you're essentially buying into the whole duck type thing which then requires modifying all other methods. Duck typing adds a bit of a wild card to the API too as you don't know how the duck typed method will really act. In this case you probably want a specialized method instead of ones designed for arrays and objects.

jridgewell:
> Do you think that consistency is worthy of 2 lines of source code? :-)

I agree with @jdalton here. Adding this means adding it to all collection-iteration functions. Additionally, this doesn't fix the arg signature for `Set`, since it passes `(value, value, set)` to the `forEach` callback.

jdalton:
Closing for now.

benjycui:
Thank you for your explanation :-)

PR #2211, opened 2015-06-11
Extend max, min to return max/min of multple arrays

tpgmartin:
This is in relation to the comment https://github.com/jashkenas/underscore/issues/2037

### DIFF
--- a/test/collections.js
+++ b/test/collections.js
@@ -564,6 +564,8 @@
     equal(3, _.max([1, 2, 3, 'test']), 'Finds correct max in array starting with num and containing a NaN');
     equal(3, _.max(['test', 1, 2, 3]), 'Finds correct max in array starting with NaN');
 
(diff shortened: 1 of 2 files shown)


jdalton:
I think this is leading to a bigger concept of more thorough guards. See [isIterateeCall](https://github.com/lodash/lodash/blob/npm/internal/isIterateeCall.js) and [example max/min](https://github.com/lodash/lodash/blob/npm/internal/createExtremum.js#L17-L19) usage.

tpgmartin:
Will this be implemented in the proposed merge?

jdalton:
> Will this be implemented in the proposed merge with lodash?

I'd hope so :)

megawac:
Going to merge for now and submit a follow up pr to implement guard checks on other methods

megawac:
Thanks!

PR #2212, opened 2015-06-12
Alias restArgs as restParam

### LINKED ISSUE #2209: Alias `_.restArgs` as `_.restParam`
For lib interop the edge `_.restArgs` would need an alias as `_.restParam`.

megawac:
Closes #2209 

Related issue (where name was discussed) #2140 

In the name of lodash/underdash compatibility I think we should probably drop `restArgs` in favour of  `restParam` which is closer to what is described in the JavaScript spec (`rest` parameter). The counter argument is we call them args elsewhere in the library.

Otherwise an alias should be sufficient.

### DIFF
--- a/test/functions.js
+++ b/test/functions.js
@@ -620,4 +620,8 @@
     }, 0)(1, 2, 3, 4);
   });
 
+  test('restParam', function() {
+    strictEqual(_.restParam, _.restArgs, 'Should be aliased');
+  });
+
 }());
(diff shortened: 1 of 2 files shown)


jdalton:
> The counter argument is we call them args elsewhere in the library.

Except a rest parameter isn't plural its a single param so `args` (plural) is incorrect no matter how you slice it. Also a function can accept several args but may only have 1 or 2 params. For example [_.without](https://github.com/jashkenas/underscore/blob/1.7.0/underscore.js#L493-L495) from v1.7.0 has 1 param, `array`, but accepts multiple arguments, values to filter out.

PR #2215, opened 2015-06-16
added _.coalesce() + tests

jgrygowski:
Proposed new feature: coalesce

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -876,25 +876,23 @@
       return result;
     };
   };
-  
+
   // Returns the first non-null, defined value
   _.coalesce = function() {
     var args = arguments;
     return function() {
       for (var i = 0; i < args.length; i++) {
-        if (typeof(args[i]) === 'function') {
-          if (args[i].apply() !== null && args[i].apply() !== undefined) {
+        if (typeof args[i] === 'function') {
+          if (args[i].apply() !== null && typeof args[i].apply() !== 'undefined') {
             return args[i].apply();
           }
-        } else {
-          if (args[i] !== null && args[i] !== undefined) {
-            return args[i];
-          }
+        } else if (args[i] !== null && typeof args[i] !== 'undefined') {
+          return args[i];
         }
       }
       return null;
     };
-  }
+  };
 
   // Returns a function that will only be executed on and after the Nth call.
   _.after = function(times, func) {
--- a/underscore.js
+++ b/underscore.js
@@ -877,6 +877,24 @@
     };
   };
 
+  _.coalesce = function() {
+    var args = arguments;
+    return function() {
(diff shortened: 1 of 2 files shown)


megawac:
Is this based on a feature in another programming language/library?

jgrygowski:
@megawac I was inspired by COALESCE in SQL.

akre54:
This could be achieved much simpler with `_.any`:

```js
_.coalesce = function() {
  var result;
  var found = _.any(arguments, function(arg) {
    result = _.isFunction(arg) ? arg() : arg;
    return result != null;
  });
  return found ? result : null;
};
```

That said, what's the real-world use case for this? I'm struggling to think of one.

jashkenas:
> That said, what's the real-world use case for this in JavaScript?

Quite. Got one?

jgrygowski:
@jashkenas @akre54 For example: when parsing text from various sources and trying to extract data, some patterns will match and others will not, this could be a quick way to apply various filters and get the matched data. Useful for parsing OCR, detecting patterns, etc..

akre54:
It's probably best to do that kind of processing with a `while` loop and a `return` (see [how CoffeeScript does it](https://github.com/jashkenas/coffeescript/blob/master/src/lexer.coffee), for example). This is a little too special-case for Underscore to add to core though.

jgrygowski:
@akre54 Makes sense. By the way, thank you for your sample implementation of this using _.any(), learned something new!

underscore.js
```diff
@@ -877,6 +877,23 @@
     };
   };
 
+  // Returns the first non-null, defined value
```
michaelficarra: Unless that value is a function, in which case it invokes it a few times and checks the result instead. That's some surprising behaviour right there. What you've done is lose parametricity. A nice `coalesce` function would treat all values the same -- an array of strings, an array of numbers, and, yes, an array of functions. This is why everyone hates `Array.prototype.concat`: concat creates a new array with the given value on the end *unless* that value is an array (technically @@isConcatSpreadable as of ES6), in which case it will instead have each of the array's values on the end. I would write this function like this:

```js
_.coalesce = _.partial(_.find, _, x => x != null)
```

underscore.js
```diff
@@ -877,6 +877,23 @@
...
+  _.coalesce = function() {
+    var args = arguments;
+    return function() {
+      for (var i = 0; i < args.length; i++) {
+        if (typeof args[i] === 'function') {
+          if (args[i].apply() !== null && typeof args[i].apply() !== 'undefined') {
+            return args[i].apply();
+          }
+        } else if (args[i] !== null && typeof args[i] !== 'undefined') {
```
megawac: ps, you can write this `args[i] != null` in JS

PR #2217, opened 2015-06-19
Added object merges

joeyism:
I added ability to merge two objects together


megawac:
See `_.extend` and `_.assign`

PR #2219, opened 2015-06-19
Adds paragraph for bdelespierre/underscore.php lib in #links

bdelespierre:
Hey,

It's been a while since I've been developping, testing and using my own port of Underscore in PHP. This is an updated version of [brianhaveri/Underscore.php](http://brianhaveri.github.io/Underscore.php/) which includes features that weren't possible at the time Brian made his version (mostly some new PHP stuff like closures contexts and late static binding).

I invite you to discover it [here](https://github.com/bdelespierre/underscore.php) and to let me know if you have any comment or suggestion :+1:

### DIFF
--- a/index.html
+++ b/index.html
@@ -2145,10 +2145,9 @@ _([1, 2, 3]).value();
       </p>
 
       <p>
-        <a href="http://brianhaveri.github.io/Underscore.php/">Underscore.php</a>,
-        a PHP port of the functions that are applicable in both languages.
-        Includes OOP-wrapping and chaining.
-        (<a href="https://github.com/brianhaveri/Underscore.php">source</a>)
+        <a href="https://github.com/bdelespierre/underscore.php">Underscore.php</a>,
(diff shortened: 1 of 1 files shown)


jashkenas:
If @brianhaveri agrees, should we just simple change where the Underscore.php link points to?

bdelespierre:
I think @brianhaveri will not respond to this; I proposed him to step in several weeks ago and got no response. I will modify the PR as @jashkenas stated.

captbaritone:
Agreed that the version we currently link to is pretty dead. I've noticed http://anahkiasen.github.io/underscore-php/ as well, which seems more poplar than @bdelespierre's version. However, it looks like it diverges significantly from the API.

I'm not sure exactly what our criteria is for which libraries should be included in this list, or why we even have the list in the first place. I see that it provides some validation of the popularity and influence that Underscore has had on the broader developer community, but I would be surprised if a port of Underscore's API would ever really be successful or even a good idea in another language.

All that philosophizing aside, I'm going to merge this since this version seems more active than the one to which we are currently linking. I'll also tag @anahkiasen to get his input.

I'll followup with another commit to clean up the word-wrapping.

PR #2221, opened 2015-06-24
Ensure underscore can be run in a Node VM

megawac:
Discovered during https://github.com/caolan/async/pull/804

### DIFF
--- a/test/utility.js
+++ b/test/utility.js
@@ -14,6 +14,37 @@
 
   });
 
+  if (typeof this == 'object') {
+    test('noConflict', function() {
+      var underscore = _.noConflict();
+      equal(underscore.identity(1), 1);
+      if (typeof require != 'function') {
+        equal(this._, void 0, 'global underscore is removed');
+        this._ = underscore;
+      }
+    });
+  }
+
+  if (typeof require == 'function') {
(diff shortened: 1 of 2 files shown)


akre54:
I'm not sure I understand. Wouldn't you be using CommonJS (and thus `module.exports` instead of `root`) in a node environment? Why would you be using `_.noConflict()` at all?

test/utility.js
```diff
@@ -14,6 +14,34 @@
 
   });
 
+  test('noConflict', function() {
```
akre54: :+1:  this test

megawac:
> Wouldn't you be using CommonJS (and thus module.exports instead of root) in a node environment? 
No, this is for the node [vm module](https://nodejs.org/api/vm.html) (sandboxes)

> Why would you use _.noConflict() at all?

You probably would't; the bigger issue is we're not correctly detecting the global, and thus the export for the vm, `root._ = _;` doesn't actually work.

test/utility.js
```diff
@@ -14,6 +14,34 @@
...
+    var underscore = _.noConflict();
+    equal(this._, void 0);
```
michaelficarra: To avoid strict mode problems, test `typeof _` instead.

PR #2228, opened 2015-07-06
Remove unused optimizeCb case

megawac:
Currently unneeded

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -69,9 +69,6 @@
       case 1: return function(value) {
         return func.call(context, value);
       };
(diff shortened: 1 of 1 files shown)


michaelficarra:
:pensive: yeah, I guess it is unused, but making this function partial feels like a bug waiting to happen.

michaelficarra:
I've added a clarifying comment about this intentional omission: https://github.com/jashkenas/underscore/commit/acf826fa42ba65d42c7e5cccd78ea3a7876c8032

Issue #1865, opened 2014-09-22
Expand support for methods used as iteratees.


jdalton:
Methods like `_.sample`, `_.first`, `_.initial`,  `_.last`, & `_.rest` have guards to support being used as an iteratee for methods like `_.map`. I think it'd be great if this was expanded to other methods that can operate with just 1 param, like `_.template`, `_.sortBy`, `_.uniq`, `_.flatten`, `_.min`, `_.max`, & so on. Support can be added to some by adding a simple `guard` check while others may require a [helper](https://github.com/lodash/lodash/blob/f0fb02074d41c728df28eef19fc5823c7b6bd8a6/dist/lodash.js#L702-L708).

amZotti:
I'll look into this

dotpao:
Any update here in order to allow _.uniq as _.map iteratee?

jgonggrijp:
I suggest adding this only to the functions for which a simple `guard` is sufficient. That helper looks like a pessimization as well as a potentially serious breaking change.

PR #2231, opened 2015-07-09
Add isIterationCall

jridgewell:
Stolen from lodash, used to determine if one of our underscore methods was called by `_.each`, `_.map`, or friends.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -291,7 +291,9 @@
   // Aliased as `includes` and `include`.
   _.contains = _.includes = _.include = function(obj, item, fromIndex, guard) {
     if (!isArrayLike(obj)) obj = _.values(obj);
-    if (typeof fromIndex != 'number' || guard) fromIndex = 0;
+    if (typeof fromIndex != 'number' || (guard && isIterationCall(item, fromIndex, guard))) {
+      fromIndex = 0;
+    }
     return _.indexOf(obj, item, fromIndex) >= 0;
   };
 
@@ -325,7 +327,8 @@
   _.max = function(obj, iteratee, context) {
     var result = -Infinity, lastComputed = -Infinity,
         value, computed;
-    if (iteratee == null || (typeof iteratee == 'number' && typeof obj[0] != 'object') && obj != null) {
+    if (context && isIterationCall(obj, iteratee, context)) iteratee = null;
+    if (iteratee == null) {
       obj = isArrayLike(obj) ? obj : _.values(obj);
       for (var i = 0, length = obj.length; i < length; i++) {
         value = obj[i];
@@ -350,7 +353,8 @@
   _.min = function(obj, iteratee, context) {
     var result = Infinity, lastComputed = Infinity,
         value, computed;
-    if (iteratee == null || (typeof iteratee == 'number' && typeof obj[0] != 'object') && obj != null) {
+    if (context && isIterationCall(obj, iteratee, context)) iteratee = null;
+    if (iteratee == null) {
       obj = isArrayLike(obj) ? obj : _.values(obj);
       for (var i = 0, length = obj.length; i < length; i++) {
         value = obj[i];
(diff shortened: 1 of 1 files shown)


underscore.js
```diff
@@ -136,15 +136,35 @@
     };
   };
 
+  // Helper to determine if index is within the bounds of an array.
+  var isIndex = function(index, length) {
```
megawac: is this helper necessary?

underscore.js
```diff
@@ -136,15 +136,35 @@
...
+    return typeof length == 'number' && isIndex(length, MAX_ARRAY_INDEX);
+  };
+
+  // Helper to determine if the arguments to a function are from
+  // an iteration of `each`, `map`, and friends.
+  var isIterationCall = function(value, index, collection) {
+    if (!_.isObject(collection)) return false;
+    var contains = typeof index === 'number' ?
+      isArrayLike(collection) && isIndex(index, collection.length) :
+      index in collection;
```
megawac: should we care about sparse array handling? (I.e. it won't be gaurded for sparse array indexes)

megawac: I'd rather just not test `index in collection`

underscore.js
```diff
@@ -136,15 +136,35 @@
...
   var getLength = property('length');
   var isArrayLike = function(collection) {
     var length = getLength(collection);
-    return typeof length == 'number' && length >= 0 && length <= MAX_ARRAY_INDEX;
+    return typeof length == 'number' && isIndex(length, MAX_ARRAY_INDEX);
+  };
+
+  // Helper to determine if the arguments to a function are from
+  // an iteration of `each`, `map`, and friends.
+  var isIterationCall = function(value, index, collection) {
```
megawac: Should this be a wrapper function?

underscore.js
jridgewell: Not necessarily, but there's a bit of duplication between `isArrayLike` and `isIterationCall` that'll get worse once I propose only considering ints and not floats as proper index.

underscore.js
jridgewell: > should we care about sparse array handling?

Sparse arrays will hit the `isArrayLike(collection) && isIndex(index, collection.length)` check.

> I'd rather just not test index in collection

We have to for object iteration.

underscore.js
jridgewell: Hm?

underscore.js
megawac: > > I'd rather just not test index in collection
>We have to for object iteration.

Whoops read the code too quickly

underscore.js
megawac: You know, 

```_.map = gaurdIterationCall(function() {
 ...
});
```

jridgewell: That'd be nifty, but I think it'd slow down common usage (it'll require an `#apply`). And I don't think the other contributors want more wrappers.

megawac: Currently we could get away with a direct invocation as none of them require context, but I hear you

underscore.js
```diff
@@ -330,7 +353,8 @@
   _.min = function(obj, iteratee, context) {
     var result = Infinity, lastComputed = Infinity,
         value, computed;
-    if (iteratee == null || (typeof iteratee == 'number' && typeof obj[0] != 'object') && obj != null) {
+    if (context && isIterationCall(obj, iteratee, context)) iteratee = null;
+    if (iteratee == null) {
```
megawac: Mind adding a test for this change

jashkenas:
Please explain what this is doing, beyond what the previous "guard" flag was already doing?

jdalton:
> Please explain what this is doing, beyond what the previous "guard" flag was already doing?

It's a more thorough check for when a `guard` param isn't applicable (like in cases where it's a `context` param).

jashkenas:
Maybe an example would help clarify? I'm don't see any changed tests in this PR.

jgonggrijp:
For future reference, `isIterationCall` is a throw at the helper function suggested as part of #1865. Which is a bit overused here as you only need to use a `guard` *or* `isIterationCall`, not both. Also it's a breaking change and I'm not in favor of it.

Issue #2115, opened 2015-03-17
_.defaults() does not initialize for undefined/null


balnagy:
Hi,

I'm using _.defaults to initalize the config of my class.

```
var config = {
    port: 10000
};

function Server(options) {
    this.config = _.defaults(options, config);
}
```

I expected it to work even if I don't pass any options, but it doesn't. I need to do this:

```
this.config = _.defaults(options || {}, config);
```

For me it's unexpected. How do you think?

lukeasrodgers:
`_.defaults` could be rewritten to handle this scenario, but I would say that, in general, calling it on `null` or `undefined` probably means you have a bug in your code. I would consider rewriting your code to have `options` actually be a set of defaults which you are (optionally) overriding with whatever's in `config`.

jridgewell:
See https://github.com/jashkenas/underscore/issues/1565 for history. I wouldn't mind having `_.extend` and `_.defaults` default `null` / `undefined` to an empty array instead of returning early.

balnagy:
Thanks for the link to history. Actually empty array still doesn't make sense to me if the defaults was defined.

> _.defaults(object, *defaults) 
Fill in undefined properties in object with the first value present in the following list of defaults objects.

I might interpret this wrong, but my expectation was that
1. it should always return an object
2. that object should contain all the properties from `defaults` if it was not present in `object`

jridgewell:
Lol, typo. Corrected now.

jgonggrijp:
This was adressed in #2232.

PR #2232, opened 2015-07-10
_.defaults() does not initialize for undefined/null

alewiscondev:
Fixes issue https://github.com/jashkenas/underscore/issues/2115. Defaults null and undefined objects correctly.

Paired with @jridgewell and @djrconcepts!

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -1008,9 +1008,10 @@
   };
 
   // An internal function for creating assigner functions.
-  var createAssigner = function(keysFunc, undefinedOnly) {
+  var createAssigner = function(keysFunc, defaults) {
     return function(obj) {
       var length = arguments.length;
+      if (defaults) obj = Object(obj);
       if (length < 2 || obj == null) return obj;
       for (var index = 1; index < length; index++) {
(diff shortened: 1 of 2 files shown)


jdalton:
Instead maybe passing values through `Object(value)`. That way there can be a consistency to it.

alewiscondev:
Good call @jdalton, fixed!

underscore.js
```diff
@@ -1008,9 +1008,10 @@
   };
 
   // An internal function for creating assigner functions.
-  var createAssigner = function(keysFunc, undefinedOnly) {
+  var createAssigner = function(keysFunc, defaults) {
     return function(obj) {
       var length = arguments.length;
+      if (defaults) obj = Object(obj);
```
megawac: Any good reason we avoid this for `assign` and `extend`?

jridgewell: `Object.assign` throws errors for this case. Not if we want to stray too far from that.

It makes sense to me in defaults. If you have nothing, and want to default it to `{a: 1}`, `{a: 1}` should be returned.

PR #2233, opened 2015-07-10
Iterate functions as objects

### LINKED ISSUE #1992: Collections should not iterate functions as array-likes
I brought this up in https://github.com/jashkenas/underscore/pull/1990#issuecomment-68230946 and https://github.com/jashkenas/underscore/issues/1590#issuecomment-42617297

I think all the collection methods should not iterate functions as `arrays` but as `objects`. I would like to address this in #1990

``` js
_.each(_, console.log.bind(console));
```

djrosenbaum:
fixes #1992: Collections should not iterate functions as array-likes

Paired with @jridgewell, @alewiscondev.

### DIFF
--- a/test/collections.js
+++ b/test/collections.js
@@ -148,18 +148,17 @@
   test('Iterate function as object', function() {
     //github issue #1992: collection methods should not iterate functions as arrays but as objects
 
-    var collection_methods = ['each','map','reduce','reduceRight','find','filter','where','findWhere','reject','every','some','contains','pluck','max','min','sortBy','groupBy','indexBy','countBy','shuffle','sample','toArray','size','partition'];
-    collection_methods.map(iterate_function_as_object);
+    var collectionMethods = ['each', 'map', 'reduce', 'reduceRight', 'find', 'filter', 'where', 'findWhere', 'reject', 'every', 'some', 'contains', 'pluck', 'max', 'min', 'sortBy', 'groupBy', 'indexBy', 'countBy', 'shuffle', 'sample', 'toArray', 'size', 'partition'];
+    _.each(collectionMethods, iterateFunctionAsObject);
 
-    function iterate_function_as_object(method) {
-
-      var fn = function(a, b, c) {};
+    function iterateFunctionAsObject(method) {
+      var fn = function(a, b, c) {a = b = c;};
       fn.foo = 'bar';
-      _[method](fn, function(value,key) {
-        strictEqual(key+value, 'foobar', '_.' + method + ' iterates function as object');
+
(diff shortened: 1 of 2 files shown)


megawac:
:+1: just needs a couple tests

jdalton:
:+1:

test/collections.js
```diff
@@ -145,6 +145,23 @@
     });
   });
 
+  test('Iterate function as object', function() {
+    //github issue #1992: collection methods should not iterate functions as arrays but as objects
+
+    var collection_methods = ['each','map','reduce','reduceRight','find','filter','where','findWhere','reject','every','some','contains','pluck','max','min','sortBy','groupBy','indexBy','countBy','shuffle','sample','toArray','size','partition'];
+    collection_methods.map(iterate_function_as_object);
```
megawac: Hi, couple nits here:

- you can't use `methods.map` here as we test in ES3 environments (use `_.each` instead)
- camel case methods
- add a space between the comman and strings

megawac: Run `npm run lint` to get more linting errors

megawac:
Turns out there might be a blocker `typeof node.childNodes == 'function'` in phantomJS... wtf?

/cc @jdalton

djrconcepts:
Node List types fixed in phantomJS 2.0
https://github.com/ariya/phantomjs/issues/11478

michaelficarra:
@djrconcepts Underscore supports many old/obscure environments, and dropping support for one as common/recent as PhantomJS 1.9.x doesn't seem wise.

jdalton:
I'll figure it out on the lodash side and report back.

djrconcepts:
@michaelficarra added support for versions of phantomJS before v2.0

underscore.js
```diff
@@ -143,6 +143,7 @@
   var MAX_ARRAY_INDEX = Math.pow(2, 53) - 1;
   var getLength = property('length');
   var isArrayLike = function(collection) {
+    if ( typeof collection === 'function' && collection.constructor !== root.NodeList ) return false;
```
megawac: Unfortunately this may not be sufficient for cross frame scenarios :/

michaelficarra: The [[Class]] for NodeLists is "NodeList", at least for my modern chrome. Would that test be sufficient?

djrconcepts:
@megawac added cross frame compatibility

jashkenas:
> So for iteratee checks or binds where the expected value is a function then the light check is used.

Can that be amended as a part of this patch?

jgonggrijp:
Closing because of #2830.

PR #2237, opened 2015-07-14
added function findAllIndexes

Eskatrem:
findAllIndexes finds all the indexes of an array-like that pass some test.
For example:
_.findAllIndexes([1,2,3,3,1,2,3,5,1], function(x) {return x >2});
[2,3,6,7]

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -626,7 +626,7 @@
       return -1;
     };
   };
-    
+
 
   // Returns the first index on an array-like that passes a predicate test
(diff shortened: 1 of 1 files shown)


jashkenas:
Thanks, but no thanks.

Eskatrem:
Why not?

jashkenas:
Because it's not needed often enough to justify shipping it with Underscore. Feel free to `_.mixin` it for your own use.

PR #2239, opened 2015-07-16
_.omit maps keys to String, _.pick should as well

mitar:
Making `_.pick` convert keys to `String` as well, to match the behavior of `_.omit`.


mitar:
Ah, invalid. This is not the problem. Ignore.

Issue #2249, opened 2015-07-28
[enhancement] partitionMuti


cabbage89:
function  partitionMuti(array,every){

var  arr=[];
_.each(array,function(n,index){
         var rem = Math.floor(index / every);
         if (index% every == 0)
                 arr[rem] = [];
          arr[rem].push(n);
});
return arr;
}

cabbage89:
_.partitionMuti([1,2,3,4,5,6,7,8,9],3)   ==   [[1,2,3],[4,,5,6],[7,8,9]]

jdalton:
It's [_.chunk](https://github.com/lodash/lodash/blob/npm/array/chunk.js#L29-L44).

megawac:
Duplicate of #1891

Issue #2130, opened 2015-03-31
Chunk function


hhoechtl:
It would be really great to have a *chunk* function which splits an array into n equal parts.

Currently something like this can be done with the following code:

    _.chain(container.parameter).groupBy(function (el, idx) {
        return Math.floor(idx / 4);
    }).toArray().value();

But having this as ```_.chunk(array, numberOfChunks);``` would be really great.

Florian-R:
See #1919 for a currently discussed PR.

PR #2252, opened 2015-07-28
Chunk

megawac:
Resolves #1919 pull request (squash, adapted fork of that branch)

Biggest deviation from pr above (and lodash) is `chunkSize < 1` returns an array of 0 chunks (empty array)

Closes a bunch of issues #2130 #1891 #2249 #696 #998 #714 (others perhaps?)

Ping https://github.com/underdash/underdash/issues/7 https://github.com/gkz/prelude-ls/pull/81

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -699,12 +699,13 @@
   // Split an **array** into several arrays containing **count** or less elements
   // of initial array
   _.chunk = function(array, count) {
+    count = +count || 1;
+    if (count <= 1) return slice.call(array);
+
     var result = [];
-
-    if (array == null || count <= 0) return [];
-
-    for (var i = 0, length = array.length; i < length;) {
-      result.push(array.slice(i, i += count));
+    var i = 0, length = array.length;
+    while (i < length) {
+      result.push(slice.call(array, i, i += count));
     }
 
     return result;
--- a/underscore.js
+++ b/underscore.js
@@ -699,15 +699,13 @@
   // Split an **array** into several arrays containing **count** or less elements
   // of initial array
   _.chunk = function(array, count) {
-    count = +count || 1;
-    if (count <= 1) return slice.call(array);
+    if (count == null || count < 1) return [];
 
     var result = [];
     var i = 0, length = array.length;
     while (i < length) {
       result.push(slice.call(array, i, i += count));
     }
-
     return result;
   };
 
--- a/underscore.js
+++ b/underscore.js
@@ -696,6 +696,20 @@
     return range;
   };
 
+  // Split an **array** into several arrays containing **count** or less elements
+  // of initial array
+  _.chunk = function(array, count) {
+    var result = [];
+
+    if (array == null || count <= 0) return [];
+
+    for (var i = 0, length = array.length; i < length;) {
+      result.push(array.slice(i, i += count));
+    }
+
+    return result;
+  };
+
   // Function (ahem) Functions
   // ------------------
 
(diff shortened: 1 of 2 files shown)


jdalton:
I'll align lodash to this implementation.

michaelficarra:
There are no tests for a default count. I'm still against the idea of having a default count.

underscore.js
```diff
@@ -696,6 +696,20 @@
     return range;
   };
 
+  // Split an **array** into several arrays containing **count** or less elements
+  // of initial array
+  _.chunk = function(array, count) {
+    count = count != null ? count : 1;
```
michaelficarra: When there's two cases, as in all conditional expressions, don't negate the test. Should be `count = count == null ? 1 : count`. Or we could avoid an unnecessary assignment and use `if (count == null) count = 1;`.

jridgewell:
> There are no tests for a default count.

[Line 549](https://github.com/jashkenas/underscore/pull/2252/files#diff-f900544b2245f714b4e01b71b7152cbeR549)?

What's your argument against a default `count`?

underscore.js
jridgewell: I prefer the `if` statement, though the reversed ternary would be good too.

michaelficarra:
Ah, I didn't see that one. My argument against a default count is that I don't feel any particular count is "special" or should be assumed or will be used more frequently than others.

underscore.js
```diff
@@ -696,6 +696,20 @@
...
+    if (count < 1) return [];
```
jridgewell: Should this be an `[[]]`?

jdalton: Welcome to the rabbit hole :rabbit:

jridgewell: I guess leave as it. :stuck_out_tongue:

underscore.js
megawac: Sure

underscore.js
michaelficarra: No, the correct answer is no.

megawac:
> Ah, I didn't see that one. My argument against a default count is that I don't feel any particular count is "special" or should be assumed or will be used more frequently than others.

I sort of agree, however I don't want to deviate from `lodash` too much

jdalton:
The default could be `0` too. Look to built-in methods with similar params, they have default values.

megawac:
I'm fine with `0` or `1`

jdalton:
The one weak win for having it be `1` would be when it's used as an iteratee in `_.map` (assuming call guards are in place) but that's a pretty narrow case.

jdalton:
I went with `1` before as that's the smallest meaningful result.

michaelficarra:
A default of 0 will just return `[]` unconditionally. That's probably a better default. By the way @jdalton are you coming to the TC39 meeting? There's like 20 MSFT employees here and you're not one of them.

jdalton:
> By the way @jdalton are you coming to the TC39 meeting? There's like 20 MSFT employees here and you're not one of them.

Imma be at the dinner. On the agenda items I'm :+1: on `trimLeft` and `trimRight` (no name change) and I'm experimenting with the more thorough/robust `RegExp.escape` implementation being discussed (it escapes leading numbers and things) in lodash as `_.escapeRegExp`.

jdalton:
I'm down with `0` being a default. It means there's little use for it as a `_.map` iteratee which means no need to add a guard for it either. (keeps it simple)

megawac:
Right on, amended.

On Tue, Jul 28, 2015 at 3:47 PM, John-David Dalton <notifications@github.com

PR #2254, opened 2015-07-28
fixes #2253: default export for ES6 imports

### LINKED ISSUE #2253: add default export to allow for usage from ECMAScript 6
This is needed to allow for the `_` in the following program:

```js
import _, {some, every} from "underscore";
```

PR incoming.

michaelficarra:
Fixes #2253.

### DIFF
--- a/test/utility.js
+++ b/test/utility.js
@@ -409,4 +409,8 @@
     strictEqual(template(), '<<\nx\n>>');
   });
 
+  test('#2253 - default export for ES6 imports', 1, function() {
+    strictEqual(_, _['default']);
+  });
+
 }());
--- a/underscore.js
+++ b/underscore.js
@@ -45,6 +45,9 @@
     this._wrapped = obj;
   };
 
+  // Add a "default" field for ES2015 module system compatibility
+  _['default'] = _;
+
   // Export the Underscore object for **Node.js**, with
   // backwards-compatibility for their old module API. If we're in
   // the browser, add `_` as a global object.


jdalton:
This isn't actually an es6 export though. If this is to cater to transpilers there are other ways to accomplish it without adding a property to the `_` object.

jdalton:
\cc @thejameskyle and @sebmck as sanity checks.

thejameskyle:
Because of:

```js
function _interopRequireDefault(obj) {
  return obj && obj.__esModule ? obj : { 'default': obj };
}
```

This isn't going to work the way you think it will. ([See](http://babeljs.io/repl/#?experimental=true&evaluate=false&loose=false&spec=false&code=import%20foo%20from%20'foo'%3B%0A%0Aconsole.log(foo)%3B&playground=false))

Although you can add the following and it will work:

```js
Object.defineProperty(exports, "__esModule", {
  value: true
});
```

thejameskyle:
I should also point out that because of that interop `import _ from 'underscore'` currently works fine.

sebmck:
@thejameskyle Nope, @michaelficarra is using the `commonStrict` module formatter which assumes that all default imports are found via the `default` property.

thejameskyle:
Oh, missed that.

jdalton:
Is it correct to assert that adding a `['default']` property to `_` which isn't exported in any es6 specific way is transpiler-specific and not an ES6 module export route.

michaelficarra:
That's true.

jdalton:
@michaelficarra 
> That's true.

Ah cool. Then I think it's an issue for transpilers to tackle and not Underscore.

PR #2255, opened 2015-07-31
Move isEqual under Collections in index.html

gaborluk:
Since `isEqual` operates on arrays as well as on objects, it belongs under Collections. The language has been changed accordingly.

### DIFF
--- a/index.html
+++ b/index.html
@@ -221,6 +221,7 @@
       <li>- <a href="#toArray">toArray</a></li>
       <li>- <a href="#size">size</a></li>
       <li>- <a href="#partition">partition</a></li>
+      <li>- <a href="#isEqual">isEqual</a></li>
     </ul>
 
     <a class="toc_title" href="#arrays">
@@ -293,7 +294,6 @@
       <li>- <a href="#matcher">matcher</a></li>
       <li>- <a href="#property">property</a></li>
       <li>- <a href="#propertyOf">propertyOf</a></li>
-      <li>- <a href="#isEqual">isEqual</a></li>
       <li>- <a href="#isMatch">isMatch</a></li>
       <li>- <a href="#isEmpty">isEmpty</a></li>
       <li>- <a href="#isElement">isElement</a></li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -809,6 +809,21 @@ <h2 id="collections">Collection Functions (Arrays or Objects)</h2>
 =&gt; [[1, 3, 5], [0, 2, 4]]
 </pre>
 
+      <p id="isEqual">
+        <b class="header">isEqual</b><code>_.isEqual(list, other)</code>
+        <br />
+        Performs an optimized deep comparison between one <b>list</b> and the <b>other</b>, to determine
```
megawac: I think two objects is better as a list is a JS object

gaborluk: The documentation consistently uses **list** for every function in the *Collections* section. It serves as a shorthand for "array or object".

megawac: Actually in this case I think object methods or utility is more correct as
it has multiple behaviors based on type switching

gaborluk: I can see what you mean, it seems that it has some undocumented behaviors. It's a tough one. I pushed the change in the first place because it had taken me some time today to track `isEqual` down, since the Collections-Arrays-Objects distinction suggests that anything under *Objects* will not work on arrays, and vice versa.

Ultimately, *Utility* seems to make the most sense. What would you suggest instead of **list** (that is not **object** to avoid the above ambiguity) as a general term for all the supported types? Maybe **value**?

gaborluk: How about something like this, under *Utility*?
> **isEqual** `_.isEqual(first, second)`
Performs a comparison between **first** and **second** of arbitrary types, without type conversion. Where applicable, performs an optimized deep comparison to determine if **first** and **second** should be considered equal.
```
var stooge = {name: 'moe', luckyNumbers: [13, 27, 34]};
var clone  = {name: 'moe', luckyNumbers: [13, 27, 34]};
stooge == clone;
=> false
_.isEqual(stooge, clone);
=> true
```

akre54:
I agree with @megawac. This could maybe be a utilitys method, but it's much more an object (not necessarily Object) method.

PR #2261, opened 2015-08-04
add findKeyByValue

craigmichaelmartin:


jdalton:
Because it's so composable and I could see folks differing on strict vs deep equality I think this would be better suited as a mixin than core.

PR #2262, opened 2015-08-05
Added collection creation utility function

kohesion:
A collection is a list of objects with the same keys.

Now they can be created by sending an array of the keys and an array of
arrays of values.

_.collection(['name', 'score'], [['Anastasia', 'Jim', 'Anthony', 'Edmund'], [20, 40, 90, 80, 100]]);

=> [{"name":"Anastasia","score":20},{"name":"Jim","score":40},{"name":"Anthony","score":90},{"name":"Edmund","score":80},{"name": undefined, "score":100}]

### DIFF
--- a/test/utility.js
+++ b/test/utility.js
@@ -51,6 +51,13 @@
     ok(new _(instance) === instance);
   });
 
+  test('collection', function () {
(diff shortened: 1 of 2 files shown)


kohesion:
I want to do this from a branch, not master.

I will reopen this with a different source.

PR #2265, opened 2015-08-06
Collection creation utility

kohesion:
Creates a collection from a list of keys and a list of lists of values.

### DIFF
--- a/test/utility.js
+++ b/test/utility.js
@@ -52,7 +52,7 @@
   });
 
   test('collection', function() {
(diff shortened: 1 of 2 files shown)


captbaritone:
First, thanks for the suggestion! I appreciate the work you put into this. I hope PhantomJS didn't give you too much trouble!

However, I don't personally think this is a common enough use-case to merit inclusion in the library.

Also, I think it should be pretty easy to do this idiomatically in Underscore:

    var values = [['Anastasia', 'Suzanne', 'Edmund'], [90, 95, 100, 50]];
    var keys = ['name', 'score'];

    _.chain(values).unzip().map(function(objValues) {
        return _.object(keys, objValues);
    }).value();

If you use it frequently, you could wrap that up as a mix-in.

If you think that I'm missing some common use-case or pattern where this would be widely needed, feel free to let us know.

PR #2266, opened 2015-08-06
Compact allowZero flag

kohesion:
_.compact([0, 1, false, undefined], true)
=> [0, 1]


javadev:
It may be better to create a new method not just add boolean parameter for existing. Or may be add iterator which can filter null values for compact.

PR #2267, opened 2015-08-06
Simplify - reduces nested array complexity to depth 1

kohesion:
_.simplify([[1, 2, 3], [2, 3, 4], [[[3, 4, 5], [6, 7, 8]]]])

=> [[1, 2, 3], [2, 3, 4], [3, 4, 5], [6, 7, 8]]

### DIFF
--- a/test/arrays.js
+++ b/test/arrays.js
@@ -79,6 +79,10 @@
     equal(result, 3, 'works on an arguments object');
   });
 
+  test('simplify', function() {
+    deepEqual(_.simplify([[1, 2, 3], [2, 3, 4], [[[3, 4, 5], [6, 7, 8]]]]), [[1, 2, 3], [2, 3, 4], [3, 4, 5], [6, 7, 8]], 'should simplify array as expected');
+  });
+
   test('flatten', function() {
     deepEqual(_.flatten(null), [], 'Flattens supports null');
     deepEqual(_.flatten(void 0), [], 'Flattens supports undefined');
--- a/underscore.js
+++ b/underscore.js
@@ -225,6 +225,22 @@
     if (key !== void 0 && key !== -1) return obj[key];
   };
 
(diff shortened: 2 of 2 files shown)


jdalton:
That's what `_.flatten` does.

kohesion:
Flatten does something different:

.flatten([[1,2,3], [2,3,4], [[[3,4,5], [6,7,8]]]])
=> [1, 2, 3, 2, 3, 4, 3, 4, 5, 6, 7, 8]

.flatten([[1,2,3], [2,3,4], [[[3,4,5], [6,7,8]]]], true)
=> [1, 2, 3, 2, 3, 4, Array[2]]

kohesion:
Flatten with shallow flag does this:
_.flatten([[1,2,3], [2,3,4], [[[3,4,5], [6,7,8]]]], true)
[1, 2, 3, 2, 3, 4, Array[2]]

Simplify does this:

_.simplify([[1,2,3], [2,3,4], [[[3,4,5], [6,7,8]]]])
[Array[3], Array[3], Array[3], Array[3]]

What do you mean by "mixin"?

Sorry, I'm a newbie.

jdalton:
> Flatten does something different:

Correct. But it's close enough to where you could create something similar mapping over the array.

> What do you mean by "mixin"?

My mixin I mean as a separate chunk of code that you [_.mixin](http://underscorejs.org/#mixin) to Underscore.

kohesion:
Hmmm... Okay.

I just thought it would be super useful when used with .apply . Sometimes
.apply can be tricky because you have to do .apply(null, [a].concat(b)
where b is a nested array. I thought it would be easier to do .apply(null,
_.simplify(a, b, c))

I don't suppose you could leave it up for discussion?

If not, I understand.

Either way, thank you for your time.

PR #2275, opened 2015-08-11
Created _.perform function.

briantracy:
Similar function to Objective-C's [`-[NSArray makeObjectsPerformSelector:]`](https://developer.apple.com/library/prerelease/ios/documentation/Cocoa/Reference/Foundation/Classes/NSArray_Class/index.html#//apple_ref/occ/instm/NSArray/makeObjectsPerformSelector:), but will build an array composed of the return values.


jridgewell:
That's `_.invoke`. :wink:

PR #2276, opened 2015-08-18
OptimizeCb Documentation Update

bkillenit:
updated optimizeCb function documentation to include the comments that explains the exclusion of case 2.


megawac:
That file is generated by docco from https://github.com/jashkenas/underscore/blob/master/underscore.js

akre54:
Please open just one pull per change.

PR #2277, opened 2015-08-18
Function Bind Doc Update

bkillenit:
removed functionBind variable instantiation from documentation.


akre54:
These docs are auto-generated. We regenerate them on every version bump.

bkillenit:
got it, thank you for explaining!

PR #2278, opened 2015-08-18
Cb Comments Fix

bkillenit:
fixed English for comments of cb function


akre54:
"identity" refers to the [`_.identity`](http://underscorejs.org/#identity) function. This could be marked a little clearer though.

PR #2279, opened 2015-08-19
Made Reducer Function Comment More Explicit

bkillenit:
Increased comprehensibility of the comments for the reducer function to make the optimizations around the arguments.length value more understandable to contribs.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -185,8 +185,8 @@
 
   // Create a reducing function iterating left or right.
   var createReduce = function(dir) {
-    // Optimized iterator function as using arguments.length
-    // in the main function will deoptimize the, see #1991.
+    // Optimize iterator function to use arguments.length
+    // in the main function. This will deoptimize the length lookup time.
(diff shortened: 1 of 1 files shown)


underscore.js
```diff
@@ -185,8 +185,8 @@
 
   // Create a reducing function iterating left or right.
   var createReduce = function(dir) {
-    // Optimized iterator function as using arguments.length
-    // in the main function will deoptimize the, see #1991.
+    // Optimize iterator function to use arguments.length
```
megawac: to avoid using arguments.length

underscore.js
```diff
@@ -185,8 +185,8 @@
...
+    // in the main function. This will deoptimize the length lookup time.
```
megawac: , as this would deoptimize reduce, see #1991.

megawac:
You're currently mixing up functions, that functions purpose is to avoid looking up arguments.length in the main reduce funciton

jashkenas:
Yeah, this is backwards.

PR #2284, opened 2015-08-24
Remove outliers

kohesion:
// reduce a collection by removing outliers for a given property
  // Use it like this:
  // var col = [{a:20}, {a:30}, {a: '5'}, {a:20}, {a: 10}, {a: '5'}, {a: '5'}, {a: '5'}, {a: 20}, {a: 30}, {a: 50}]

  // for the collection examine property 'a' and take the most common 10% for values above 6
  // _.removeOutliers(col, 'a', 0.1, 6);


akre54:
This is a little too specific for core underscore. Would make a great mixin though.

PR #2285, opened 2015-08-25
Change string equality checks from '!=' or '==' to '!==' or '===' respectively.

wgpsutherland:
There were some strings being compared with '==', and others being compared with '===', so I have made it uniform across the code to '==='. 

Please let me know if I have formatted this pull request incorrectly, or if there are any changes I need to make. Today is the first time I have committed to a large, open source project, so if there is anything I should have done differently please let me know. Thanks!

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -145,7 +145,7 @@
   var getLength = property('length');
   var isArrayLike = function(collection) {
     var length = getLength(collection);
-    return typeof length == 'number' && length >= 0 && length <= MAX_ARRAY_INDEX;
+    return typeof length === 'number' && length >= 0 && length <= MAX_ARRAY_INDEX;
   };
 
   // Collection Functions
@@ -306,7 +306,7 @@
   _.max = function(obj, iteratee, context) {
     var result = -Infinity, lastComputed = -Infinity,
         value, computed;
-    if (iteratee == null || (typeof iteratee == 'number' && typeof obj[0] != 'object') && obj != null) {
+    if (iteratee == null || (typeof iteratee === 'number' && typeof obj[0] != 'object') && obj != null) {
       obj = isArrayLike(obj) ? obj : _.values(obj);
       for (var i = 0, length = obj.length; i < length; i++) {
         value = obj[i];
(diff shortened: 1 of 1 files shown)


akre54:
Hi @wgpsutherland and welcome to open source :)

Strings are fine to compare with `==` because we know they will always be the same type. It saves us a few characters and isn't really hurting anything.

wgpsutherland:
Would it be worth changing the strings that are currently being compared with `===` to `==`, so that a consistent style is used throughout? Thanks for your feedback!

akre54:
I'd say lets leave 'em be. When you start worrying on a case-by-case basis which equality operator to use, the bikeshedders win. It can potentially open up some ugly bugs if we meant to use one over the other and didn't consider an edge case.

jdalton:
Seems related to linting so \cc @michaelficarra.

michaelficarra:
Yeah, I'd be up for at least making it consistent. I didn't realise we were mixing `==` and `===` for comparisons of values with the same type.

megawac:
Same, either one is fine by me. This has bugged me in the past but I never bothered to patch it

wgpsutherland:
@megawac @michaelficarra I have a pull request ready for string comparisons all using `==` if you would prefer them over `===`

akre54:
Simple rules:
* Anything guaranteed to be a string gets `==`
  * The return value from `typeof`
* Anything guaranteed to be a number gets `==`
  * `arr.length`
* Nullish (`undefined` or `null`) checks use `==`
* Anything user-provided gets `===`

This is where things get hairy. I'd personally prefer `===` over `==` where possible (`== null` being the one exception). I'd rather we not spend too many more brain cycles on this if we can.

michaelficarra:
In my own personal projects, the only time I use `==` is for `== null`, but I'd also be fine with @akre54's proposal.

akre54:
Agreed with @michaelficarra. Personally I only ever use `==` for null and `===` for everything else, but I realize it's cargo-culting in the same vein as automatic semicolon insertion.

If it really irks someone, open a pull and let's be done with this.

PR #2287, opened 2015-08-26
Change equality comparisons of definite strings and numbers

wgpsutherland:
As discussed in #2285, this pull request changes the code to match the guidelines proposed by @akre54.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -11,8 +11,8 @@
   // Establish the root object, `window` (`self`) in the browser, `global`
   // on the server, or `this` in some virtual machines. We use `self`
   // instead of `window` for `WebWorker` support.
-  var root = typeof self === 'object' && self.self === self && self ||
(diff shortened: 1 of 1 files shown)


underscore.js
```diff
@@ -1149,7 +1149,7 @@
     if (a !== a) return b !== b;
     // Exhaust primitive checks
     var type = typeof a;
-    if (type !== 'function' && type !== 'object' && typeof b !== 'object') return false;
+    if (type !== 'function' && type !== 'object' && typeof b != 'object') return false;
```
megawac: `type` is definitely a string

wgpsutherland: @megawac yes it definitely is, however changing it to `type != 'function'` causes the travis build to fail as it doesn't pass the linter.

jashkenas:
This PR is a pretty good example of Peak Bikeshed ;)

Issue #2288, opened 2015-08-26
Add an implementation for lists as applicative functors for great non-determinism.


jfarid27:
My proposal is to add a function that takes a list of functions of the same arity, say n, and n arrays. For each combination of elements from those n arrays, apply the combination to each of the given functions. For example...

```javascript

//Case 1
_.nMap([function (a, b) { return a + b }], [1, 2, 3], [10, 11, 12]) //  -> [11, 12, 13, 12, 13, 14, 13, 14, 15]


//Case 2
var a = function (f, j) {
    return f + j
}

var b = function(f, j) {
    return f * j
}

_.nMap([a, b], [2, 4], [1, 2]) // -> [3, 4, 5, 6, 2, 4, 4, 8]

```

In case 1, I want all the ways of adding elements from the given lists. In case 2, I want all the ways of not only adding elements, but multiplying them from the given lists. 

Essentially this solves an issue I have in javascript because there is no concept of list comprehensions until [ES7](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Array_comprehensions). If I want to do a computation for each combination of choices from multiple lists, I have to [use embedded maps](http://jsfiddle.net/vqtz3qdz/) or some wonky for loops. This is nicer and has the added benefit of multiple computations as well as the multiple combinations of arguments.

Also, it should not enforce order. Lists can be viewed as a value that "can't decide what it is", so these computations should be seen as "applying a function that can't decide what it is, to values that can't decide what they are". Order doesn't really matter, and the user can implement his own sort if needed. This is described in [learnyouahaskell](http://learnyouahaskell.com/functors-applicative-functors-and-monoids#applicative-functors) and is a very powerful concept. 

It would be a cool addition.

Issue #452, opened 2012-01-30
still recommend flatMap or concatMap


rebcabin:
with reference to issue 436, I agree that documenting the second argument to flatten is a good start, but it didn't quite close the issue for me.  I would still recommend adding a "flatMap" or "concatMap" primitive to the base library so that we get proper chain composability. Adding this new primitive would not break anything, but would make chains of array-valued maps composable. For instance, imagine the following expression, in pseudo-C#, that represents the line items for high-value orders from a list of customers:
Customers
    .Where(customer => customer.Orders
        .SelectMany(order => order.LineItems)
        .Sum() > 1000)
    .SelectMany(customer => customer.Orders)
    .SelectMany(order => order.LineItems) ;

to write this in underscore with a new flatMap primitive, the "SelectMany"s become "flatMaps."  

I implemented a proposal for flatmap in a fork of underscore's source, along with a few unit tests for your consideration. I don't have docco running and I didn't produce a "min" version of the javascript, so it's just the javascript and the unit tests. https://github.com/rebcabin/underscore/commit/4f90bab75d8f69df8d8d0317354f9f5981e2fedb

AlexGalays:
@inca, Or use ramda:   http://ramdajs.com/docs/R.html#chain

eirirlar:
+1

pfultz2:
Well, `flatMap` is very fundamental, more so than `filter` and `map`(especially since they can be implemented using `flatMap`). I just don't understand the rejection of basic list comprehension.

> Needing a recursive flatten is what is uncommon. 

Exactly. List comprehension(ie `flatMap`) is much more common than recursive flatten.

inca:
I'm with you, my friends :smile:

AlexGalays:
flatMap is a very basic tool in one FP programmer's tool belt. It wouldn't hurt to peek at what other languages have.

arcseldon:
Would recommend readers interested in flatMap also take a look at Ramda - some of the core contributors were originally involved with Underscore.js. Very well thought out JavaScript library for functional programmers, with auto-currying of much of the API methods which makes composing functions easy to do. 
Here is an implementation of [Chain](http://ramdajs.com/ramdocs/docs/R.html#chain) - equivalent to flatMap).

QuotableWater7:
+1 to `flatMap`

PR #2289, opened 2015-08-26
Adding a list implementation of applicative functors.

jfarid27:
PR for #2288

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -926,11 +926,11 @@
     var args = Array.prototype.slice.call(arguments, 1);
 
     if (args.length === 1) {
-      var appls = _.map(funcs, function(func) {
+      var appl = _.map(funcs, function(func) {
         return _.map(args[0], func);
       });
 
-      return _.reduce(appls, function(agg, result) {
+      return _.reduce(appl, function(agg, result) {
         return agg.concat(result);
       }, []);
 
--- a/underscore.js
+++ b/underscore.js
@@ -919,6 +919,31 @@
 
   _.restArgs = restArgs;
 
+  // Returns non-deterministic map computation of list of functions applied to combinations of
+  // arguments from given lists of possible arguments. An implementation of lists as applicative functors.
+  _.nMap = function(funcs) {
+    var self = this;
+    var args = Array.prototype.slice.call(arguments, 1);
+
+    if (args.length === 1) {
+      return _.map(funcs, function(func) {
+        return _.map(args[0], func);
+      }).reduce(function(agg, result) {
+        return agg.concat(result);
+      }, []);
+    } else if (args.length > 1) {
+      var partials = _.map(funcs, function(func) {
+        return _.map(args[0], function(arg){
+          return _.partial(func, arg);
+        });
+      }).reduce(function(agg, result) {
+        return agg.concat(result);
+      }, []);
+      var rest = Array.prototype.slice.call(args, 1);
+      return _.nMap.apply(self, [partials].concat(rest));
+    }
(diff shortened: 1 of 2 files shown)


akre54:
Can you give a more real-world example for this?

underscore.js
```diff
@@ -919,6 +919,31 @@
...
+  // Returns non-deterministic map computation of list of functions applied to combinations of
+  // arguments from given lists of possible arguments. An implementation of lists as applicative functors.
+  _.nMap = function(funcs) {
+    var self = this;
+    var args = Array.prototype.slice.call(arguments, 1);
+
+    if (args.length === 1) {
+      return _.map(funcs, function(func) {
+        return _.map(args[0], func);
+      }).reduce(function(agg, result) {
```
akre54: Keep in mind the return value for `_.map` is an array. This `reduce` call here relies on `Array.prototype.reduce` which isn't present in pre-es5 environs. Try using `_.chain`.

underscore.js
```diff
@@ -919,6 +919,31 @@
...
+        return agg.concat(result);
+      }, []);
+    } else if (args.length > 1) {
+      var partials = _.map(funcs, function(func) {
+        return _.map(args[0], function(arg){
+          return _.partial(func, arg);
+        });
+      }).reduce(function(agg, result) {
```
akre54: ditto

jfarid27:
@akre54 I just added using _.reduce. Check out my write-up in #2288. The real world example is in the jsFiddle mentioned, but again, it's just a nice way to implement list comprehensions. Also, since a list comprehension can only "compute" one function, this is a beefed up list comprehension.

akre54:
I saw #2288. Got something more-real world than adding and multiplying numbers? List comprehensions are great, but this looks really complicated.

Philosophically, Underscore tries to be a set of generally-applicable composable utilities and I'm struggling to see how this fits.

jfarid27:
@akre54 That's a hard request. It's abstract, like map, reduce, partial application, etc. Specifically it's a way to do this in javascript, without ES7 support...

```python
x = [f(x, y) for f in fs for x in xs for y in ys]
```

In fact, map and this nMap are related. They look like this.
```haskell
map :: (a -> b) -> [a] -> [b]
nMap :: [(a -> b)] -> [a] -> [b]
```

It just allows for a list of functions to apply to a given list of a's.

Imagine asking what's a real world application of map? It just helps in certain situations. Map can be implemented with for loops, but I hate rolling my own when I need to do things like get combinations of values out of hash tables, or make lots of combinations of values to test functions on. [For me, adding and multiplying numbers is real world since I work on data, mainly doing sums on pivots of a hash table.](http://jsfiddle.net/e2jrbtor/1/) Which real world application would you like?

akre54:
> For me, adding and multiplying numbers is real world since I work on data, mainly doing sums on pivots of a hash table.

Totally. But Underscore tries to give you the building blocks for function composition, not become the end-all be-all of functional libraries. It doesn't have a ton of other data-manipulation or statistics utilities, and adding this function without adding others wouldn't make a ton of sense without a few broader applicable use cases (which `map` and `partial` have).

You can do these examples a bit simpler (and IMO clearer) with nested `_.map` and `_.flatten` (see also the discussion on `flatMap` and `concatMap` on #452).

jfarid27:
@akre54 I wish I saw #452 before sending this PR. Had I known you already had people clamoring at the gates to implement *every* functional pattern that Haskell and Scala, I probably wouldn't have worked on this...

Issue #2296, opened 2015-09-14
Support astral symbols in _.toArray


jdalton:
```js
var string = 'I 💕 the 🍂';
Array.from(string);
// => [ "I", " ", "💕", " ", "t", "h", "e", " ", "🍂" ]

[...string]
// => [ "I", " ", "💕", " ", "t", "h", "e", " ", "🍂" ]

_.toArray(string);
// => [ "I", " ", "", "", " ", "t", "h", "e", " ", "", "" ]
```

lucasfanning:
:+1:

lucasfanning:
![image](https://cloud.githubusercontent.com/assets/5374165/9857050/b422ddde-5aee-11e5-9066-591b9c7f0828.png)

for me this works.

jdalton:
@lucasfanning Yes an array to an array would work. That's not the issue.

lucasfanning:
@jdalton , I'm sorry

jdalton:
@lucasfanning no worries!

akre54:
That's kind of cheating, `[...string]` and `Array.from(string)` [are equivalent](http://babeljs.io/repl/#?experimental=true&evaluate=true&loose=false&spec=false&playground=false&code=var%20string%20%3D%20'I%20%F0%9F%92%95%20the%20%F0%9F%8D%82'%3B%0A%5B...string%5D). 

Also, `_.toArray`, like most Underscore methods, isn't really meant to be used on strings. Might be better for Underscore.string.

jdalton:
May want to change the language of the code comment for `_.toArray` as
```js
// Safely convert anything iterable into a real, live array.
```
as `iterable` has a meaning in ES6.

michaelficarra:
Agreed with @jdalton, this needs to be fixed.

PR #2297, opened 2015-09-14
Support astral symbols in _.toArray

lucasfanning:
https://github.com/jashkenas/underscore/issues/2296

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -435,6 +435,7 @@
   // Safely create a real, live array from anything iterable.
   _.toArray = function(obj) {
     if (!obj) return [];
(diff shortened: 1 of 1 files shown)


akre54:
Underscore targets ES3+, `Array.from` is ES6.

jdalton:
It's doable in pre- ES6 enviros without relying on `Array.from`. Could have waited a bit before the insta-close.

michaelficarra:
@jdalton The real solution would be nothing like this, though. No reason in working off of this.

jdalton:
You all do what you all do.

PR #2298, opened 2015-09-15
Maintain astral characters in _.toArray

### LINKED ISSUE #2296: Support astral symbols in _.toArray

JonAbrams:
Fixes #2296

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -432,10 +432,15 @@
     if (_.has(result, key)) result[key]++; else result[key] = 1;
   });
 
+  var reStrSymbol = /[^\ud800-\udfff]|[\ud800-\udbff][\udc00-\udfff]|[\ud800-\udfff]/g;
   // Safely create a real, live array from anything iterable.
   _.toArray = function(obj) {
     if (!obj) return [];
     if (_.isArray(obj)) return slice.call(obj);
+    if (_.isString(obj)) {
+      // Keep surrogate pair characters together
+      return obj ? obj.match(reStrSymbol) : [];
+    }
     if (isArrayLike(obj)) return _.map(obj, _.identity);
     return _.values(obj);
   };
--- a/test/collections.js
+++ b/test/collections.js
@@ -789,6 +789,10 @@
     var numbers = _.toArray({one: 1, two: 2, three: 3});
     assert.deepEqual(numbers, [1, 2, 3], 'object flattened into array');
 
+    var astral = _.toArray('\udfff💩poop💩\ud800');
+    assert.deepEqual(astral, ['\udfff', '💩', 'p', 'o', 'o', 'p', '💩', '\ud800'], 'maintains astral characters');
+    assert.deepEqual(_.toArray(''), [], 'empty string into empty array');
+
     if (typeof document != 'undefined') {
       // test in IE < 9
       var actual;


underscore.js
```diff
@@ -436,6 +436,18 @@
   _.toArray = function(obj) {
     if (!obj) return [];
     if (_.isArray(obj)) return slice.call(obj);
+    if (_.isString(obj)) {
+      // Keep surrogate pair characters together
+      // https://mathiasbynens.be/notes/javascript-unicode
+      return _.reduce(obj, function (output, char, idx) {
+        if (idx > 0 && obj[idx-1].match(/[\uD800-\uDBFF]/) && char.match(/[\uDC00-\uDFFF]/)) {
+          output.push(output.pop() + char);
```
michaelficarra:
```js
output[output.length - 1] += char;
```

underscore.js
```diff
@@ -436,6 +436,18 @@
   _.toArray = function(obj) {
     if (!obj) return [];
     if (_.isArray(obj)) return slice.call(obj);
+    if (_.isString(obj)) {
+      // Keep surrogate pair characters together
+      // https://mathiasbynens.be/notes/javascript-unicode
+      return _.reduce(obj, function (output, char, idx) {
+        if (idx > 0 && obj[idx-1].match(/[\uD800-\uDBFF]/) && char.match(/[\uDC00-\uDFFF]/)) {
```
michaelficarra: I'd prefer a lookahead to a lookbehind here, but that would require you to switch this up from a reduce to a simple for loop. At the very least, I'd use `<` comparison with the actual boundary characters (`x < "\uD800"`, etc.) instead of using a regexp.

JonAbrams: Isn't the regex more concise? Also, isn't lookbehind + reduce more concise?

Apologies, but I'm new to contributing here, so am not sure yet what styles/techniques are preferred.

underscore.js
JonAbrams: Still requires the pop though.

michaelficarra: Why?

JonAbrams: Derp, because I'm dumb. nvm!

underscore.js
JonAbrams: Also, would _.each be preferable to a simple for loop?

JonAbrams: Hmm, maybe _.each is less safe than a for loop. Astral characters might (should?) be treated as a single item when iterating in the future.

michaelficarra: We may just want to make this fix in `_.each` an piggy back on that support in this position and possibly others.

michaelficarra:
I like it. This LGTM, so :+1:, though we should look into pulling this support into `_.each`.

JonAbrams:
Looks like _.each, _.map, and other similar iterating functions all use for loops. They'd all need to be changed in the same way to be consistent.

jdalton:
`_.each`, `_.map` are consistent with their built-in counterparts:
```js
Array.prototype.forEach.call("💕", function(v) { console.log(v); });
// logs "", ""
```
Underscore won't tackle strings for older environments either. I think as long as there's a way to create an array from a string users can pass that to the other methods. Similar to handling iterators and iterables.

michaelficarra:
@jdalton Then can I get a :+1:?

jdalton:
Here's lodash's implementation:
```js
/** Used to match [string symbols](https://mathiasbynens.be/notes/javascript-unicode). */
var reStrSymbol = /[^\ud800-\udbff\udc00-\udfff]|[\ud800-\udbff][\udc00-\udfff]|[\ud800-\udbff\udc00-\udfff]/g,
    reStrSurrogate = /[\ud800-\udbff\udc00-\udfff]/;

function stringToArray(string) {
  return string ? string.match(reStrSymbol) : [];
}

function stringSize(string) {
  return (string && reStrSurrogate.test(string))
    ? stringToArray(string).length
    : string.length;
}
```

jdalton:
Other than that I'm for it or I wouldn't have implemented it and opened the issue. My :+1: is implied.

JonAbrams:
@jdalton does `reStrSymbol` contain an extra `|[\ud800-\udbff\udc00-\udfff]`?

underscore.js
```diff
@@ -436,6 +436,11 @@
   _.toArray = function(obj) {
     if (!obj) return [];
     if (_.isArray(obj)) return slice.call(obj);
+    if (_.isString(obj)) {
+      // Keep surrogate pair characters together
+      // https://mathiasbynens.be/notes/javascript-unicode
+      return obj.match(/[^\ud800-\udbff\udc00-\udfff]|[\ud800-\udbff][\udc00-\udfff]/g);
```
jdalton: Snipping the regexp means you fail to match lone surrogates.

jdalton: From the lodash unit tests:
```js
var hearts = '\uD83D\uDC95',
    pairs = hearts.split(''),
    loneSurrogates = pairs[0] + ' ' + pairs[1];

assert.deepEqual(_.toArray(loneSurrogates), [pairs[0], ' ', pairs[1]]);
```

jdalton:
Nope, I commented on it.

underscore.js
jdalton: I'll break down the regexp:
```js
/[^\ud800-\udbff\udc00-\udfff]|[\ud800-\udbff][\udc00-\udfff]|[\ud800-\udbff\udc00-\udfff]/g
```
I went with a negated character class to keep things simple and reuse the same unicode ranges in the regexp.

`[^\ud800-\udbff\udc00-\udfff]` means match anything that isn't a high or low surrogate, so regular characters like `a` or `@`.

`[\ud800-\udbff][\udc00-\udfff]` these two character classes match a high and low surrogate pair.

`[\ud800-\udbff\udc00-\udfff]` this last one looks similar but matches lone surrogates that are not paired.

JonAbrams:
I see, makes sense. I've restored it and updated the spec as well.

underscore.js
```diff
@@ -436,6 +436,11 @@
   _.toArray = function(obj) {
     if (!obj) return [];
     if (_.isArray(obj)) return slice.call(obj);
+    if (_.isString(obj)) {
+      // Keep surrogate pair characters together
+      // https://mathiasbynens.be/notes/javascript-unicode
+      return obj.match(/[^\ud800-\udbff\udc00-\udfff]|[\ud800-\udbff][\udc00-\udfff]|[\ud800-\udbff\udc00-\udfff]/g);
```
jdalton: Don't forget the empty string case I commented on earlier.

JonAbrams: Good catch. Fixed, spec'd.

jdalton:
:sparkles:

underscore.js
```diff
@@ -436,6 +436,12 @@
   _.toArray = function(obj) {
     if (!obj) return [];
     if (_.isArray(obj)) return slice.call(obj);
+    if (_.isString(obj)) {
+      // Keep surrogate pair characters together
+      // https://mathiasbynens.be/notes/javascript-unicode
+      var reStrSymbol = /[^\ud800-\udbff\udc00-\udfff]|[\ud800-\udbff][\udc00-\udfff]|[\ud800-\udbff\udc00-\udfff]/g;
```
michaelficarra: `\ud800-\udbff\udc00-\udfff` can be replaced with `\ud800-\udfff` in both instances, right?

michaelficarra: If this is constant, it should be pulled out of the function.

jdalton: > `\ud800-\udbff\udc00-\udfff` can be replaced with `\ud800-\udfff` in both instances, right?

Those are the ranges for high and low surrogates. I don't think wiping out the range into one is a good idea. But @mathiasbynens would know more.

jdalton: Hey @JonAbrams can you chill on the rebase while I try to string a convo together. You're rebasing before we have a verdict on a course of action.

michaelficarra: They're adjacent ranges, and thus may be combined. I'm just making regexp semantics-preserving transformations.

mathiasbynens:
:+1:

underscore.js
JonAbrams: @jdalton apologies, I'll leave it as is until there's a consensus or a merge.

underscore.js
```diff
@@ -433,9 +433,15 @@
...
 
   // Safely create a real, live array from anything iterable.
+  var reStrSymbol = /[^\ud800-\udfff]|[\ud800-\udbff][\udc00-\udfff]|[\ud800-\udfff]/g;
   _.toArray = function(obj) {
     if (!obj) return [];
     if (_.isArray(obj)) return slice.call(obj);
+    if (_.isString(obj)) {
+      // Keep surrogate pair characters together
+      // https://mathiasbynens.be/notes/javascript-unicode
+      return (obj) ? obj.match(reStrSymbol) : [];
```
michaelficarra: Unnecessary parens around `obj`.

underscore.js
```diff
@@ -433,9 +433,15 @@
   });
 
   // Safely create a real, live array from anything iterable.
+  var reStrSymbol = /[^\ud800-\udfff]|[\ud800-\udbff][\udc00-\udfff]|[\ud800-\udfff]/g;
   _.toArray = function(obj) {
     if (!obj) return [];
     if (_.isArray(obj)) return slice.call(obj);
+    if (_.isString(obj)) {
+      // Keep surrogate pair characters together
+      // https://mathiasbynens.be/notes/javascript-unicode
```
michaelficarra: What exactly are you attempting to reference with this URL?

underscore.js
jdalton: So the unicode character after `\udbff` is `\udc00`? My concern is that there are gaps between the two ranges and so collapsing it into one globbers the gaps.

mathiasbynens: Yeah, `[\ud800-\udbff\udc00-\udfff]` is equivalent to `[\ud800-\udfff]`.

underscore.js
```diff
@@ -433,9 +433,15 @@
   });
 
   // Safely create a real, live array from anything iterable.
```
michaelficarra: This comment should remain adjacent to the `_.toArray` definition.

underscore.js
jdalton: :rocket!

underscore.js
jdalton: @michaelficarra 
> What exactly are you attempting to reference with this URL?

It's taken from the lodash implementation comment above the regexp.
```js
/** Used to match [string symbols](https://mathiasbynens.be/notes/javascript-unicode). */
```

michaelficarra: I don't think it's necessary. The regexp itself is not taken from there. The "keep surrogate pairs together" comment is sufficient.

JonAbrams: I actually found it via google, it's pretty definitive though, so no wonder we both used it.

I put it there because it does a great job of explaining the crazy regex.

jdalton: Psst @mathiasbynens is lodash core ;P

JonAbrams: Good job catching him!

michaelficarra:
Okay let's get this commit cleaned up so we can merge this. @JonAbrams please apply your remaining changes.

underscore.js
jdalton: @michaelficarra 
> I don't think it's necessary. The regexp itself is not taken from there. The "keep surrogate pairs together" comment is sufficient.

I like bits like that to help explain the source. Considering the implementation is based on the info from that post and others interested the feature would have a clear sign of where to go. But you're right Underscore has traditionally washed their source of such things opting to leave devs in the dark.

michaelficarra:
Thanks, @JonAbrams. :snowman:

PR #2301, opened 2015-09-16
Created _.update _.replace with unit tests to validate behavior for i…

snowmac:
…ssue 2300


snowmac:
Was reporting incorrectly locally. :( :-1:

PR #2306, opened 2015-09-27
Adding comb function. fixes #2305

### LINKED ISSUE #2305: Add pattern matching combinators

jfarid27:
Add Haskell like function definitions using pattern matching.

In Haskell a function can be defined by pattern matching over
argument values.

```haskell
myFunc :: Int -> Int -> String
myFunc 0 0 = "Hello"
myFunc 1 0 = "Goodbye"
```

We can define pattern matches to arbitrary arguments.

```haskell
myFunc _ 1 = "Good Afternoon"

(myFunc 10 1) -- "Good Afternoon"
(myFunc 1 0) -- "Goodbye"
(myFunc 0 0) -- "Hello"
```

Last, we can define standard functions simply by defining functions
with non-value arguments, similar to javascript function definitions.

```haskell
myFunc x y = "The sum is: " + (show (x+y)) 
```

Using standard function definitions in javascript, we have to use nested
if statements for each argument case. The same definitions as above are...

```javascript
function myFunc(a, b) {
    if (b == 1){
        return "Good Afternoon"
    } 
    if (a == 0 && b == 0){
        return "Hello"
    }
    if (a == 1 && b == 0){
        return "Goodbye"
    }
    return "The sum is: " + (a+b)
}
```

_.comb allows concise pattern matching using haskell-like syntax

```javascript
//A default function if no patterns are matched
var myFunc = _.comb(function(a, b){
    return "The sum is: " + (a+b)
})

//Cases to match on and values to return
myFunc.link(_, 1, "Good Afternoon")
myFunc.link(0, 0, "Hello")
myFunc.link(1, 0, "Goodbye")
```

The arguments to link sets patterns to match over in a first link,
first match basis. Lastly, we can pattern match on values and supply 
abstract functions to apply arguments to.

```javascript
myFunc.link(2, _, function(a, b){
    return "Product of 2:" + (a * b)
})
```

Sum outputs with the definitions...
```javascript
myFunc(2, 3) // 6
myFunc(10000, 1) // "Good Afternoon"
myFunc(100, 100) // "The sum is: 200"
```

A full example of definitions in javascript...
```javascript
var myFunc = _.comb(function(a, b){
    return "The sum is: " + (a+b)
})
myFunc.link(0, 0, "Hello")
myFunc.link(1, 0, "Goodbye")
myFunc.link(_, 1, "Good Afternoon")
myFunc.link(2, _, function(a, b){
    return "Product of 2:" + (a * b)
})

myFunc(2, 3) // 6
myFunc(10000, 1) // "Good Afternoon"
myFunc(100, 100) // "The sum is: 200"
```

### DIFF
--- a/test/functions.js
+++ b/test/functions.js
@@ -89,6 +89,66 @@
     _.partial.placeholder = _;
   });
 
+  test('comb', function(assert){
+    var testFunc1 = function(a, b, c) {
+      return 'Func1 ' + b;
+    }, testFunc2 = function(a, b, c) {
+      return 'Func2 ' + b;
+    }, testFunc3 = function(a, b, c) {
+      return 'Func3 ' + b;
+    }, defFunc = function(a, b, c) {
+      return 'Default';
+    };
+
+    var testFunc = _.comb(defFunc);
+
+    var setDefault = testFunc._default();
+    assert.equal(setDefault(), 'Default', 'Properly loads default function when initialized');
+
+    //Single function tests
+    testFunc.link(0, _, 0, testFunc1);
+    var setCase1 = testFunc._funcs()[0],
+    setPattern1 = testFunc._patterns()[0];
+
+    assert.equal(setCase1.apply(null, ['foo', 'bar', 'zed']), 'Func1 bar', 'Properly stores functions with case method');
+    assert.deepEqual([0, _, 0], setPattern1, 'Properly stores patterns with case method');
+
+    //Tests for default matcher
+    var defMatch = testFunc._matcher();
+    var val1 = [0, 'foo', _],
+        val2 = [0, 'foo', 'bar'],
+        val3 = [1, 'bar', 'zed'];
+
+    assert.equal(defMatch(val1, val2), true, 'Properly matches given values with given array to match to: 1');
+    assert.equal(defMatch(val1, val3), false, 'Properly matches given values with given array to match to: 2');
+    assert.equal(defMatch(val2, val2), true, 'Properly matches given values with given array to match to: 3');
+    assert.equal(defMatch(val2, val3), false, 'Properly matches given values with given array to match to: 4');
+
+    //Tests using main caller using matcher
+    testFunc.link(0, _, 1, testFunc2);
+    testFunc.link(1, _, _, testFunc3);
+
+    assert.equal(testFunc(0, 'Case 1', 0), 'Func1 Case 1', 'Properly calls function based on case: 1');
+    assert.equal(testFunc(0, 'Case 2', 1), 'Func2 Case 2', 'Properly calls function based on case: 2');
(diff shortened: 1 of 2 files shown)


akre54:
Same deal as #2289. Can you give us something more concrete for an example? Javascript isn't Haskell.

jfarid27:
@akre54 just this

```javascript
var myFunc = function(a, b){
   if (a == "foo" && b == "foo"){
     return 0
   } else if (a == "foo" && b == "bar") {
     return 1
   } else if (a == "bar" && b == "foo") {
     return 2
   } else if (a == "bar" && b == "bar") {
     return 3
   }
   return 4
}
```
vs this 

```javascript
var myFunc = _.comb(function(){
    return 4
})
myFunc.link("foo", "foo", 0)
myFunc.link("foo", "bar", 1)
myFunc.link("bar", "foo", 2)
myFunc.link("bar", "bar", 3)
```

The second case is nice to read.

akre54:
foos and bars aren't exactly concrete :)

It's a slightly prettier syntax but it's certainly not as clear or debuggable. And it doesn't fit our generally-applicable composable utilities philosophy. It could make a cool standalone library though.

akre54:
It looks like you could also just use an object for this case:

```js
var defaultVal = 4;

var outputs = {
  foo: {
    foo: 0,
    bar: 1
  },
  bar: {
    foo: 2,
    bar: 3
  }
}

var myFunc = function(a,b) {
  return outputs[a][b] || defaultVal;
};

```

jfarid27:
@akre54 only somewhat. _.comb supports arbitrary values in the pattern match. A case for (_, "zed") can't be supported using objects since calling with an undefined property, for example calling with ("blah", "zed") will result in a TypeError. You could do a lot of the things _.comb does with standard JS, but it would get messy really fast.

megawac:
I see your use but I agree with Adam, this is better for a standalone library (this ticket could be a good start to a blog post) as it doesn't mix well with the current offerings of underscore

akre54:
You could always guard on undefined. Or use a regular ol conditional. I
don't see this way as any clearer

On Mon, Sep 28, 2015, 11:35 Jalil F. <notifications@github.com> wrote:

> @akre54 <https://github.com/akre54> only somewhat. *.comb supports
> arbitrary values in the pattern match. A case for (*, "zed") can't be
> supported using objects since calling with an undefined property, for
> example calling with ("blah", "zed") will result in a TypeError. You could
> do a lot of the things _.comb does with standard JS, but it would get messy
> really fast.
>
> —
> <https://github.com/jashkenas/underscore/pull/2306#issuecomment-143780920>
> .
>
-- 

Adam K (mobile)

jfarid27:
@megawac I'm starting to think so too. @akre54 is right. Adding more functional-like patterns to underscore might be too much. There are patterns that are useful to 90% of programmers like map and mapObject, but other patterns like #2288 are interesting but probably won't be used by anyone.

PR #2307, opened 2015-09-29
syntax fix

modernlegend:
change example syntax fom `bind` to `on`, since they are functionally equivalent, and the use of `bind` was causing fellow readers some confusion when used as part of the example for the `_.bindAll` function


michaelficarra:
Works for me.

PR #2309, opened 2015-10-01
add _.combinations to array utils

jaeschliman:
Hello, this adds `_.combinations` to the array utils, which fits nicely with `_.chunk`.
The call `_.combinations(array, n)` returns all combinations of length n from array,
or the empty array if n is not provided, or if n is too large or too small.

Cheers,
Jason


michaelficarra:
When would you use something like this?

jaeschliman:
It can be very handy for e.g. finding the closest pair of points in a list or similar 'best pair' or 'best triplet' problems.

akre54:
We don't include combination or permutation functions because they're not common enough to fit with Underscore's m.o.. A dedicated library for this kind of thing would be better.

PR #2312, opened 2015-10-04
Exclude null from _.max() function

### LINKED ISSUE #2218: _.max should ignore null values
_.max currently has strange behavior if the dataset you're evaluating contains nulls.

Basically you can run into some indeterminate behavior depending on the order of your data
sooo...
```
_.max([null, 0]) => null
```
but
```
_.max([0, null]) => 0
```

While you _could_ just go ahead and filter your data, I don't really see any advantages to leaving it this way.

Cause: line 310
```
for (var i = 0, length = obj.length; i < length; i++) {
    value = obj[i];
    if (value > result) {
        result = value;
    }
}
```

Both `(null > 0)` and `(0 > null)` evaluate to false.

I'm willing to push a PR with some unit tests but I just wanted to get some other opinions first

kutyel:
Fixes #2218

### DIFF
--- a/test/collections.js
+++ b/test/collections.js
@@ -564,6 +564,18 @@
     assert.equal(3, _.max([1, 2, 3, 'test']), 'Finds correct max in array starting with num and containing a NaN');
     assert.equal(3, _.max(['test', 1, 2, 3]), 'Finds correct max in array starting with NaN');
 
+    assert.equal(3, _.max([1, 2, 3, null]), 'Finds correct max in array starting with num and containing a `null`');
+    assert.equal(3, _.max([null, 1, 2, 3]), 'Finds correct max in array starting with a `null`');
+
+    assert.equal(3, _.max([1, 2, 3, '']), 'Finds correct max in array starting with num and containing an empty string');
+    assert.equal(3, _.max(['', 1, 2, 3]), 'Finds correct max in array starting with an empty string');
+
+    assert.equal(3, _.max([1, 2, 3, false]), 'Finds correct max in array starting with num and containing a false');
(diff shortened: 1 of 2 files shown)


akre54:
This would remove all falsy values, not just `null`.

kutyel:
Done in commit 00a9291 @akre54 is this not the intended behaviour? A value must be **truthy** in order to be evaluated and compared in the `_.max()` function.

jdalton:
What about `NaN` values?

jdalton:
or `undefined` values?

akre54:
This would also fail for 0 and negative values.

kutyel:
@jdalton wouldn't those be included in b5b66f1? There was already a spec checking for `NaN`.

kutyel:
@akre54 @jdalton  Would be OK to check for a `typeof value === 'number'`?

kutyel:
Now it also **works with 0 and negative values**. For reference, check the specs in f4365a3

akre54:
FWIW, [`d3.max` uses `!= null`](https://github.com/mbostock/d3/blob/v3.5.6/src/arrays/max.js). Fix this, squash down and I'll merge. Thanks for the pull.

kutyel:
@akre54 Done, your welcome! :smile:

akre54:
Can you squash down to a single commit (`git rebase -i HEAD~5`) for a clean history?

jridgewell:
Needs to be in `_.min`, too.

jdalton:
I dig the d3 version. Avoids the `-Infinity` and `Infinity` thing too.

kutyel:
@akre54 Done, clean history and corrected also `_.min()` function.

PR #2314, opened 2015-10-06
Fix IE prototype-less hasOwn shenanigans

### LINKED ISSUE #2308: _.has and Object.create(null)
There's a bug in IE9 and IE10 where:

``` js
var o = Object.create(null);
o[1] = 'a';

Object.prototype.hasOwnProperty.call(o, '1');
// => false
Object.keys(o);
// => ['1']
```

jridgewell:
Fixes #2308.

### DIFF
--- a/test/objects.js
+++ b/test/objects.js
@@ -775,12 +775,21 @@
     assert.ok(_.has(obj, 'func'), 'has() works for functions too.');
     obj.hasOwnProperty = null;
     assert.ok(_.has(obj, 'foo'), 'has() works even when the hasOwnProperty method is deleted.');
-    var child = {};
-    child.prototype = obj;
+    var Child = function() {};
+    Child.prototype = obj;
+    var child = new Child();
     assert.ok(!_.has(child, 'foo'), 'has() does not check the prototype chain for a property.');
     assert.strictEqual(_.has(null, 'foo'), false, 'has() returns false for null');
     assert.strictEqual(_.has(void 0, 'foo'), false, 'has() returns false for undefined');
   });
+
+  if (Object.create) {
+      test('has: #2308', 1, function(assert) {
+          var child = Object.create(null);
+          child[1] = 'foo';
+          assert.ok(_.has(child, '1'));
+      });
+  }
 
   test('isMatch', function(assert) {
     var moe = {name: 'Moe Howard', hair: true};
--- a/underscore.js
+++ b/underscore.js
@@ -33,7 +33,8 @@
   var
     nativeIsArray = Array.isArray,
     nativeKeys = Object.keys,
-    nativeCreate = Object.create;
+    nativeCreate = Object.create,
+    nativeGetPrototypeOf = Object.getPrototypeOf;
 
   // Naked function reference for surrogate-prototype-swapping.
   var Ctor = function(){};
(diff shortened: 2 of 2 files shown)


test/objects.js
```diff
@@ -775,11 +775,12 @@
     assert.ok(_.has(obj, 'func'), 'has() works for functions too.');
     obj.hasOwnProperty = null;
     assert.ok(_.has(obj, 'foo'), 'has() works even when the hasOwnProperty method is deleted.');
-    var child = {};
-    child.prototype = obj;
+    var child = _.create(obj);
```
akre54: Doesn't this rely on implementation details of `_.create` though?

jridgewell: `child.prototype = obj` never did anything, so I think we're better of relying on `create` to properly setup a prototype chain... :wink:

akre54: Ha, fair point. You know what I mean though, shouldn't we set the proto chain manually here?

jridgewell: I was being cheeky, I'll revise tonight.

test/objects.js
```diff
@@ -775,11 +775,12 @@
...
     assert.ok(!_.has(child, 'foo'), 'has() does not check the prototype chain for a property.');
     assert.strictEqual(_.has(null, 'foo'), false, 'has() returns false for null');
     assert.strictEqual(_.has(void 0, 'foo'), false, 'has() returns false for undefined');
+    child = _.create(null, {1: 'foo'});
```
jdalton: Not actually creating an object with a `[[Prototype]]` of `null`.

jridgewell: How so?

jdalton: Test it out. `_.create` has an `isObject` check. The `null` behavior is not shimmable so not supported in Underscore's `_.create` method.

jgonggrijp:
I closed #2308 but since this fix was already contributed and it looks lean enough, I plan to look into this more carefully later.

underscore.js
```diff
@@ -131,6 +132,10 @@
     return result;
   };
 
+  var baseGetPrototype = nativeGetPrototypeOf || function() {
+    return null;
+  };
```
jgonggrijp, line 137: By itself, I find this fallback function a bit questionable. I would either expect it to try harder to return a prototype (e.g. by trying to access the `__proto__` property of the argument) or have it named differently in order to signal that it doesn't attempt to replace `Object.getPrototypeOf`, for example `getPrototypeOrNull`.

I realize this function was introduced specifically to facilitate the change in `_.has`, but given the great distance, there is no evident link between the two.

underscore.js
```diff
@@ -1328,7 +1333,10 @@
   // Shortcut function for checking if an object has a given property directly
   // on itself (in other words, not on a prototype).
   _.has = function(obj, key) {
-    return obj != null && hasOwnProperty.call(obj, key);
+    if (!_.isObject(obj)) return false;
```
jgonggrijp, line 1336: `_.isObject(obj)` is not exactly equivalent to `obj != null`, so this is a subtle but potentially very nasty breaking change.

underscore.js
```diff
@@ -1328,7 +1333,10 @@
...
+    if (hasOwnProperty.call(obj, key)) return true;
+    // Fix an IE prototype-less object bug. #2308
+    return key in obj && baseGetPrototype(obj) === null;
```
jgonggrijp, line 1339: I'm willing to believe that this fixes the issue for old IE, but it seems to me that it might also cause false positives in any engine that doesn't have `Object.getPrototypeOf` (including but not limited to old IE).

jgonggrijp (commented):
Upon closer inspection, I decided not to adopt this change. Review comments below for future reference.

PR #2317, opened 2015-10-09
Add "whereNot"

CurosMJ:


jdalton:
Hi @CurosMJ! You could just do:
```js
_.reject(obj, attrs)
```

CurosMJ:
@jdalton 
yea, but similar argument could be made for `_.where` also. We can use `_.find` instead of `_.where`

CurosMJ:
This adds convenience, nothing more.

jdalton:
> yea, but similar argument could be made for `_.where` also. We can use `_.find` instead of `_.where`

You're right and I've dropped `_.where` in my own code.

PR #2321, opened 2015-10-13
Ensure that `window._` is assigned when elements with IDs of `exports` and `module` exist.

### LINKED ISSUE #2292: Underscore fails to register window._ if there is an element with id=exports
Create a document this like:
```
<body>
<div id="exports"></div>
<script src="https://cdnjs.cloudflare.com/ajax/libs/underscore.js/1.8.3/underscore.js"></script>
<script>
console.log(_)
</script>
```

Result in Chrome and Firefox:
`Uncaught ReferenceError: _ is not defined`

I was bamboozled when adding an element with the id of "exports" caused underscore to stop working, but it turns out thanks to some obscure IE functionality you can get an element by id by doing window[id], so when this underscore code runs:
```
if (typeof exports !== 'undefined') {
    if (typeof module !== 'undefined' && module.exports) {
      exports = module.exports = _;
    }
    exports._ = _;
  } else {
    root._ = _;
  }
```
It attaches _ to the exports *element*, causing window._ to no longer be available.

This problem doesn't occur in lodash, which checks `exports.nodeType` to detect if the exports object is actually a node.

d10:
Closes #2292.

**Note:** I did not add tests as I’m not too sure how to go about doing it.

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -7,6 +7,7 @@
 </head>
 <body>
   <div id="qunit"></div>
+  <div id="exports" style="display: none"></div>
   <script src="vendor/qunit.js"></script>
   <script src="vendor/qunit-extras.js"></script>
   <script src="../underscore.js"></script>
--- a/underscore.js
+++ b/underscore.js
@@ -48,8 +48,10 @@
   // Export the Underscore object for **Node.js**, with
   // backwards-compatibility for their old module API. If we're in
   // the browser, add `_` as a global object.
-  if (typeof exports != 'undefined') {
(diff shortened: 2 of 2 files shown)


jdalton:
For a similar test see https://github.com/lodash/lodash/blob/3.10.1/test/index.html#L32.

michaelficarra:
LGTM. Add @jdalton's test and I'll merge.

d10:
Sure, I’ve added the `exports` element, with `style="display: none"`.

michaelficarra:
Thanks.

PR #2323, opened 2015-10-17
Fix bug when iterating object literal with 'length' key

mdiebolt:
This fixes an issue when using `_.each` to iterate over an object that contains the key "length"

Before:

```JavaScript
_.each({"length": 3}, function(value, key){
  console.log(key, value)
})

// => 0 undefined
// => 1 undefined
// => 2 undefined
```

After: 

```JavaScript
_.each({"length": 3}, function(value, key){
  console.log(key, value)
})

// => "length" 3
```

I'd appreciate feedback from anyone who knows a more elegant way to determine if an object contains the key `length` and not the property length since JavaScript doesn't differentiate between property access and accessing an object key.

### DIFF
--- a/test/collections.js
+++ b/test/collections.js
@@ -30,6 +30,12 @@
     _.each(obj, function(){ count++; });
     assert.equal(count, 3, 'the fun should be called only 3 times');
 
+    obj = {'length': 1};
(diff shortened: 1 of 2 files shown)


michaelficarra:
See the many duplicates.

PR #2327, opened 2015-10-25
Handle null or undefined templates

vikikamath:
```js
_.template = function(text, settings, oldSettings) {

```
Handles if `text` parameter was set to `undefined` or `null`

### DIFF
--- a/test/utility.js
+++ b/test/utility.js
@@ -188,6 +188,11 @@
   });
 
   test('template', function(assert) {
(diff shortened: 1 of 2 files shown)


megawac:
I think I'd rather provide an error as this case is probably erroneous, unless you have a counter example?

jdalton:
FWIW lodash treats nullish strings as empty.

PR #2333, opened 2015-11-02
Make defaults an immutable copy

jridgewell:
A little bit of discussion here https://github.com/jashkenas/backbone/pull/3843#discussion_r43514938.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -1030,10 +1030,9 @@
   };
 
   // An internal function for creating assigner functions.
-  var createAssigner = function(keysFunc, defaults) {
+  var createAssigner = function(keysFunc) {
     return function(obj) {
       var length = arguments.length;
-      if (defaults) obj = Object(obj);
       if (length < 2 || obj == null) return obj;
       for (var index = 1; index < length; index++) {
         var source = arguments[index],
@@ -1041,7 +1040,7 @@
             l = keys.length;
         for (var i = 0; i < l; i++) {
           var key = keys[i];
-          if (!defaults || obj[key] === void 0) obj[key] = source[key];
+          obj[key] = source[key];
         }
       }
       return obj;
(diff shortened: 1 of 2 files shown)


jdalton:
Not tackleable with `_.defaults({}, obj, other)` ?

jridgewell:
> Not tackleable with _.defaults({}, obj, other) ?

We were hitting the `watch` case from https://github.com/jashkenas/backbone/issues/3842 with that code. We could also let it be mutable if we check if the `dest`'s property is enumerable, overriding if not.

jdalton:
I don't see how not mutating would avoid the issue? You're still performing a check `value !== void 0 || !(key in dest)`.

jridgewell:
Because it's testing from the source, not the destination. `source[key] !== void 0 || !(key in dest)`.

jdalton:
> Because it's testing from the source, not the destination. 

Can you slow walk me through it.

jridgewell:
If the `value` is present (not `undefined`) in `source`, we assign it to `destination`. If it is `undefined`, we only assign it if the key isn't already in `source`. So:

```js
newDefaults({ hasOwnProperty: true }); // => { hasOwnProperty: true }
newDefaults({ hasOwnProperty: undefined }); // => { }

oldDefaults({ }, { hasOwnProperty: true }); // => { }
```

jdalton:
Cool, the more I think on this the more I dig it. The common case for `_.defaults` is options objects which should at least be shallow cloned by methods anyways.

underscore.js
```diff
@@ -1105,8 +1104,22 @@
     return _.pick(obj, iteratee, context);
   });
 
-  // Fill in a given object with default properties.
-  _.defaults = createAssigner(_.allKeys, true);
+  // Fill in an object with default properties.
+  _.defaults = function() {
+    var dest = {};
+    for (var index = arguments.length - 1; index >= 0; index--) {
```
jdalton: changing the iteration order here is weird.

jdalton: what this does is pave over already set values. See
```js
_.defaults({ 'a': 1 }, { 'a': 2 })
// key: a; value: 2; value !== void 0: true; !(key in dest): true
// key: a; value: 1; value !== void 0: true; !(key in dest): false
// => { 'a': 1 }

jdalton:
I decided to simply check if the value is strict equal to those on `Object.prototype[key]`. This solves the issue while allowing `_.defaults` to act as it has in other respects. https://github.com/lodash/lodash/commit/8a3842b788e920e905827bebc82447c9ce97b29d

underscore.js
jridgewell: That's the intention, since I can't test for presence on the `destination` object (that was the old `destination[prop] === void 0`). This way, we can ensure all the proper values are set, using only the `source`'s `value` (`value !== void 0`) during the iteration.

jdalton: That's less a `_.defaults` and more an `_.extendRight`.

jridgewell: Except for the `!== void 0` check.

jdalton: Right, so it's not really `_.defaults` and not really `_.extendRight`. I think the way I went is the lesser of worse ways to go in this situation.

jgonggrijp:
@jashkenas I feel like closing this one for solving the wrong problem. Would you agree?

jashkenas:
> @jashkenas I feel like closing this one for solving the wrong problem. Would you agree?

I agree.

PR #2335, opened 2015-11-03
Update links in README.md

masterT:
Use markdown link tags. It's more neat and now we can click on it on npm page.


akre54:
Thanks but I think the bare URLs read cleaner.

masterT:
Ok, but then why not jut wrap the URLs in markdown tag `[URL](URL)`?

akre54:
Because Github renders that for us automatically. See the main project page.

masterT:
I know, but what if its not interprested by GitHub? Check the npm page https://www.npmjs.com/package/underscore

PR #2337, opened 2015-11-05
Clear the debounce

jridgewell:
Supersedes #1542.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -857,7 +857,7 @@
       }
     };
 
-    return function() {
+    var func = function() {
       context = this;
       args = arguments;
       timestamp = _.now();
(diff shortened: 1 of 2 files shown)


underscore.js
```diff
@@ -870,6 +870,13 @@
 
       return result;
     };
+
+    debounced.clear = function () {
```
akre54: Looks like this is failing lint.

akre54:
:+1:

underscore.js
jridgewell: Fixed.

jdalton:
This would be good to add to `_.throttle` too.
Lodash gets this as `throttle` is implemented by way of `_.debounce`.

jridgewell:
Already working on it.

PR #2338, opened 2015-11-05
Clear the throttle

jridgewell:


underscore.js
```diff
@@ -805,35 +805,40 @@
...
       if (!timeout) context = args = null;
     };
-    return function() {
+
+    var throttled = function() {
       var now = _.now();
       if (!previous && options.leading === false) previous = now;
       var remaining = wait - (now - previous);
-      context = this;
-      args = arguments;
```
megawac: I'd rather not change this here

jridgewell: Removed.

PR #2339, opened 2015-11-05
Cleanup the debounce function

jridgewell:


underscore.js
```diff
@@ -841,35 +841,25 @@
...
       var callNow = immediate && !timeout;
-      if (!timeout) timeout = setTimeout(later, wait);
+      if (timeout) clearTimeout(timeout);
       if (callNow) {
-        result = func.apply(context, args);
-        context = args = null;
+        timeout = setTimeout(later, wait);
+        result = func.apply(this, args);
+      } else if (!immediate) {
+        timeout = setTimeout(later, wait, this, args);
```
megawac: This won't work in pre-es5

megawac: This is the reason `_.delay` exists

jridgewell: Well that's lame. :disappointed:

megawac: You can replace `setTimeout` with delay here and see how it performs. May be worth it, this changeset is pretty nice

PR #2340, opened 2015-11-05
Cleanup the debounce function

jridgewell:
Don't know why I didn't think of `_.delay`...

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -790,7 +790,7 @@
   // Delays a function for the given number of milliseconds, and then calls
   // it with the arguments supplied.
   _.delay = restArgs(function(func, wait, args) {
-    return setTimeout(function(){
+    return setTimeout(function() {
       return func.apply(null, args);
     }, wait);
   });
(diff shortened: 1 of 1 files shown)


megawac:
+1 also Justin if you create the branches on the main repo rather than your fork sauce can run on the pull

akre54:
Nah let's leave 'em on our own repos. Little easier administratively.

megawac:
For things like this is nice to prevent regressions as would of happened if #2339 went through

akre54:
But did sauce catch it eventually? The problem with this idea is that it wouldn't catch commits by non-"contributors". I'd rather have better error alerting for both core and casual contributors than preference one over the other and potentially miss an error like this.

dcorb:
With these changes, the improvements achieved by http://modernjavascript.blogspot.co.uk/2013/08/building-better-debounce.html, where thrown away.
I don't see any discussion about it, pros/cons about the decision

leeleo26:
With this "clean up", the improvements achieved by http://modernjavascript.blogspot.co.uk/2013/08/building-better-debounce.html, where thrown away.
I don't see any discussion about it, pros/cons about the decision

megawac:
How do you mean, please justify your statement

jashkenas:
FWIW, I think he's saying that the old and crusty debounce was something that was arrived at through *a lot* of tweaks/input — this cleanup may have reverted to an even earlier, more naive, implementation that has some different behavior that isn't caught by the tests. (Testing debounce edge cases are very, very tricky)

dcorb:
Thank you Jeremy. That's what I meant. 

The improvements that were introduced here https://github.com/jashkenas/underscore/pull/1269 (based on the blog post) were basically reverted. There was no discussion about it, so I wanted to raise awareness of the regression.

The current version is cleaner and more maintainable probably. So maybe it's ok.

I don't have any problems with the current implementation. I just research different implementations at the time I was writing this blog post, some months ago https://css-tricks.com/debouncing-throttling-explained-examples/

jgonggrijp:
For future reference: the optimization was restored in #2892. CC @dcorb

dcorb:
Thanks @jgonggrijp for pinging me :) 5 years later was finally fixed 💪

PR #2341, opened 2015-11-05
Fix debounce clear error

jridgewell:
Just spotted this, re: #2340.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -872,7 +872,7 @@
 
     debounced.clear = function() {
       clearTimeout(timeout);
-      timeout = context = args = null;
+      timeout = null;
     };
 
     return debounced;


michaelficarra:
The test framework (and linter) should ensure we don't ever accidentally create unintended globals like this. What happened?

jridgewell:
Did we set that up for underscore? I can never remember which setup went in Backbone...

michaelficarra:
Well we have [a .eslintrc](https://github.com/jashkenas/underscore/blob/master/.eslintrc), but I can't find any eslint rules that disallow this. I can't imagine nobody has created this rule.

Also, mocha has a `--check-leaks` flag, but I'm not sure if an equivalent exists for qunit.

jdalton:
It exists in qunit, lodash tests use it.

megawac:
Yea we used to test for globals with qunit. not sure how we could do that with the current setup as qunitjs and karma-qunit don't support it afaik

megawac:
@michaelficarra I thought globals errored by default w/ eslint

akre54:
qunit has the "check for globals" checkbox and the "?noglobals" url param. It looks like we should be able to pass that through karma.conf.js as `client.qunit.noglobals` (see [here](https://github.com/karma-runner/karma-qunit#configuration)). 

FYI `throttle` tests are failing for me locally.

jdalton:
You can also set it via JS:
```js
QUnit.config.noglobals = true;
```

akre54:
They're equivalent. The `client` config just passes it through from Karma, though I can't see why we wouldn't always want this to run.

michaelficarra:
@megawac Found it: http://eslint.org/docs/rules/no-undef

PR #2342, opened 2015-11-11
enable eslint no-undef rule

michaelficarra:
See #2341.


jridgewell:
Is this all that's needed? If so, merge away.

michaelficarra:
@akre54 It'd be great if you could also send a PR to enable the QUnit global check you mentioned in #2341.

akre54:
Sure. It only operates on tests though (basically diffs `Object.keys(window)` before and after each test).

michaelficarra:
That's exactly what I'd like.

PR #2345, opened 2015-11-12
enable QUnit.config.noglobals in tests

akre54:
Ref jashkenas/underscore#2341

### DIFF
--- a/test/utility.js
+++ b/test/utility.js
@@ -21,6 +21,8 @@
       if (typeof require != 'function') {
         assert.equal(this._, void 0, 'global underscore is removed');
         this._ = underscore;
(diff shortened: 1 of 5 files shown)


test/utility.js
```diff
@@ -21,6 +21,8 @@
       if (typeof require != 'function') {
         assert.equal(this._, void 0, 'global underscore is removed');
         this._ = underscore;
+      } else if (typeof global !== 'undefined') {
+        delete global._;
```
akre54: `_.noConflict` actually *creates* a global, but it's undefined.

jdalton:
:rocket:

PR #2347, opened 2015-11-14
ease

kyo4311:
According to _.reduce and the _.template transform data into a string


akre54:
This is too specific for core underscore, but it would make a great mixin or plugin for your own copy.

kyo4311:
ok, thanks

PR #2348, opened 2015-11-17
debounce doc: which invocation's args are used

chris-martin:
I couldn't figure out from reading the docs whether `const f = _.debounce(x => { console.log(x); }, 1000); f(1); f(2);` would print `1` or `2`.

Discovered experimentally that the answer is `2`. This PR drops a hint about that behavior into the documentation.


akre54:
Thanks

PR #2349, opened 2015-11-18
Remove redundant empty string handing in _.toArray()

captbaritone:
As part of the review for pull request #2298, @jdalton requested that we add
this check to handle the empty string case. However, I'm guessing he didn't
realize that the first line of the function, `if (!obj) return [];`, already
handles that case.

As part of that pull request, @JonAbrams wisely added a test for the empty
string case, so as long as tests are passing this change should be safe.

See: https://github.com/jashkenas/underscore/pull/2298#discussion_r39480698

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -441,7 +441,7 @@
     if (_.isArray(obj)) return slice.call(obj);
     if (_.isString(obj)) {
       // Keep surrogate pair characters together
-      return obj ? obj.match(reStrSymbol) : [];
+      return obj.match(reStrSymbol);
     }
     if (isArrayLike(obj)) return _.map(obj, _.identity);
     return _.values(obj);


captbaritone:
Tests are failing for an unrelated reason: 

    $ nvm install 0.10
   
    Version '0.10' not found - try `nvm ls-remote` to browse available versions.
    
    The command "nvm install 0.10" failed and exited with 3 during .

captbaritone:
Closed and reopened the pull request to trigger a Travis rebuild. Tests are now green.

JonAbrams:
Nice catch!

michaelficarra:
Thanks.

captbaritone:
No problem. Thanks for the great library.

PR #2354, opened 2015-11-24
Remove redundant _.first() test

captbaritone:
This test was originally added in pull request #500 as a way to affirm that the
`_.take()` alias was working. In pull request #1663 we moved to testing aliases
explicitly, but left this test an additional `_.first()` test.

It is redundant because it is functionally identical to the test on line 10.


jdalton:
It does appear to be redundant. Thanks @captbaritone!

PR #2360, opened 2015-11-24
Clean up assertions for _.last()

captbaritone:
Makes assertions consistent with `_.first()`'s assertions. This involved
rewording the descriptions, adding a test for the object oriented version of
`_.last()`, and making the `-n` assertion explicitly check for an empty array.

### DIFF
--- a/test/arrays.js
+++ b/test/arrays.js
@@ -56,16 +56,16 @@
 
   test('last', function(assert) {
     assert.equal(_.last([1, 2, 3]), 3, 'can pull out the last element of an array');
-    assert.deepEqual(_.last([1, 2, 3], 0), [], 'can pass an index to last');
-    assert.deepEqual(_.last([1, 2, 3], 2), [2, 3], 'can pass an index to last');
-    assert.deepEqual(_.last([1, 2, 3], 5), [1, 2, 3], 'can pass an index to last');
+    assert.equal(_([1, 2, 3]).last(), 3, 'can perform OO-style "last()"');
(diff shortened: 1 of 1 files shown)


test/arrays.js
```diff
@@ -56,16 +56,16 @@
...
+    assert.deepEqual(_.last([1, 2, 3], 5), [1, 2, 3], 'returns the whole array if n > length');
     var result = (function(){ return _(arguments).last(); }(1, 2, 3, 4));
     assert.equal(result, 4, 'works on an arguments object');
     result = _.map([[1, 2, 3], [1, 2, 3]], _.last);
     assert.deepEqual(result, [3, 3], 'works well with _.map');
-
-    assert.equal(_.last(null), void 0, 'handles nulls');
-    assert.strictEqual(_.last([1, 2, 3], -1).length, 0);
+    assert.equal(_.last(null), void 0, 'returns undefined when called on null');
+    assert.deepEqual(_.last([1, 2, 3], -1), [], 'returns an empty array when asked for -n elements');
```
michaelficarra: <= 0

captbaritone: The `== 0` case is already covered in the second assertion. This assertion is intended to cover the `< 0` case, so I think it's okay for its description to only refer to the case it covers.

Maybe moving the assertions next to each other would help improve clarity?

michaelficarra: There's one property with multiple assertions to ensure that it holds. The message should describe the property, not the particular assertion, in my opinion.

captbaritone:
@michaelficarra Updated.

michaelficarra:
Thank you.

PR #2361, opened 2015-11-24
Improve assertions for _.compact()

captbaritone:
Explicitly test all falsy values, and assert it works with map.


megawac:
Thanks!

PR #2367, opened 2015-11-30
added iterate OBJ parameter to keep the consistency

adictovirtual:
pick iteratee function is called with `iteratee(value, key, obj)` but omit iteratee is defined with `iteratee(value, key)`


jdalton:
Hi @adictovirtual!

In this case the extra param isn't used so has no functional consistency to worry about.

Issue #2311, opened 2015-10-02
_.union doesn't work with arrays of objects


wilhen01:
_.union will always produce duplicates when passed arrays of objects. 

e.g. `_.union( [ { a:1 } ], [ { a:1 } ])` will return `[ { a:1 }, { a:1 } ]`

Perversely, underscore's own isEqual function will tell you that the objects in question are equal. Maybe we could have a flag/option which dictates the equality comparison to use, or the option to pass in a comparator?

michaelficarra:
I'm surprised it doesn't already accept the comparison function. :+1:

wilhen01:
It's worth pointing out that this applies to all the other array computation functions like difference, intersection, unique, etc.

amiral84:
Wouldn't it be easier if we would have an option that compares equality based on boolean parameter value that could be passed to _.union() function? If it is true then it would automatically compare all objects in that array.

E.g. <code>_.union([1, 2, 3, 10, [{a:1}, {a:1}]], true)</code>, would output <code>[1,2,3,10, {a:1}]</code>

michaelficarra:
@amiral84 No. That is unrelated. If you want that behaviour, compose union with flatten.

amiral84:
@michaelficarra Then did I miss the point of this topic? :D

michaelficarra:
@amiral84 It appears so. The feature request is explained fully and succinctly in the first comment.

lucasfanning:
because ({a:1} == {a:1}) === false;

lucasfanning:
Because {a:1} != {a:1}

dperrymorrow:
the underlying problem seems to be in ```_.unique``` as ```_.union``` is merely a wrapper function for unique and flatten.

```javascript
_.union = restArgs(function(arrays) {
  return _.uniq(flatten(arrays, true, true));
});
```

jdalton:
This thread inspired me to add `_.intersectionWith`, `_.differenceWith`, `_.unionWith`, and `_.uniqWith` to handle comparison customization.

```js
_.uniqWith([ { 'a': 1, 'b': 2 }, { 'a': 1, 'b': 3 }, { 'a': 1, 'b': 2 } ], _.isEqual);
// => [{ 'a': 1,'b': 2 }, { 'a': 1,'b': 3 }]
```

dperrymorrow:
[...]

PR #2368, opened 2015-11-30
optional parameter for predicate on _.unique

dperrymorrow:
this allows for using custom matcher function for comparison
defaults to `_.isEqual` for comparison #2311 

added a `_.uniqueWith` function as discussed, also added `_.includeWith` as a method to detect if an Array or Collection has an item passing a method. This will be useful if we implement the other methods discussed in the thread as well.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -278,15 +278,6 @@
     return _.indexOf(obj, item, fromIndex) >= 0;
   };
 
-  _.containsWith = _.includesWith = _.includeWith = function(array, item, predicate) {
-    predicate = _.isFunction(predicate) ? predicate : _.isEqual;
-    for (var i = 0, length = getLength(array); i < length; i++) {
-      if (predicate(array[i], item))
-        return true;
-    }
-    return false;
-  };
-
   // Invoke a method (with arguments) on every item in a collection.
   _.invoke = restArgs(function(obj, method, args) {
     var isFunc = _.isFunction(method);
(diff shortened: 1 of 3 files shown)


jdalton:
I think this would have a better chance of landing without bundling it with `includesWith` at the moment because that raises questions about how to handle `fromIndex` and things which are a side track to `_.uniqWith`.

dperrymorrow:
Ive updated this to use a predicate parameter #2311
let me know how you think this works. 

```javascript
_.uniq = _.unique = function(array, isSorted, iteratee, context, predicate) {
//...
```

underscore.js
```diff
@@ -533,12 +533,20 @@
   // Produce a duplicate-free version of the array. If the array has already
   // been sorted, you have the option of using a faster algorithm.
   // Aliased as `unique`.
-  _.uniq = _.unique = function(array, isSorted, iteratee, context) {
+  _.uniq = _.unique = function(array, isSorted, iteratee, context, predicate) {
     if (!_.isBoolean(isSorted)) {
       context = iteratee;
       iteratee = isSorted;
       isSorted = false;
+      predicate = _.isNull(predicate) ? context : predicate;
```
megawac: predicate == null && _.isFunction(context) ? ...

underscore.js
```diff
@@ -533,12 +533,20 @@
...
     }
+    // cannot use sort with predicate matcher
+    if (_.isFunction(predicate)) {
+      isSorted = false;
+      iteratee = null;
+      predicate = _.bind(predicate, context);
```
megawac: createCallback

underscore.js
```diff
@@ -533,12 +533,20 @@
   // Produce a duplicate-free version of the array. If the array has already
   // been sorted, you have the option of using a faster algorithm.
   // Aliased as `unique`.
-  _.uniq = _.unique = function(array, isSorted, iteratee, context) {
+  _.uniq = _.unique = function(array, isSorted, iteratee, context, predicate) {
```
megawac: I would rather split this out into another method and make this implementation internal:
1) leave `_.uniq` alone
2) add `function uniqWith(array, predicate, context)`

underscore.js
dperrymorrow: yes, that would be better to ensure that context is a function before assigning it to predicate. thanks.

underscore.js
dperrymorrow: @jashkenas felt strongly otherwise, this is what the original pull request did. 
#2311

dperrymorrow:
made the suggested changes, thanks @megawac

PR #2373, opened 2015-12-04
chore(arrays.js): clarify comment

henryw4k:
Made the comment clear by fixing the grammar.


test/arrays.js
```diff
@@ -251,7 +251,7 @@
     var stooges = _.zip(['moe', 30, 'stooge 1'], ['larry', 40, 'stooge 2'], ['curly', 50, 'stooge 3']);
     assert.deepEqual(stooges, [['moe', 'larry', 'curly'], [30, 40, 50], ['stooge 1', 'stooge 2', 'stooge 3']], 'zipped pairs');
 
-    // In the case of difference lengths of the tuples undefineds
+    // In the case of different length(s) in the tuple(s), undefine(s)
```
michaelficarra: This would be better:

> In the case of different lengths of the tuples, undefined values

Issue #2372, opened 2015-12-04
non-string programming form of _.get


jdalton:
I figured I'd float this by.

One of the reasons cited for not adding deep property getter and setter methods is wanting to avoid string programming.

Ruby has added [Array#dig](http://ruby-doc.org/core-2.3.0_preview1/Array.html#method-i-dig) and [Hash#dig](http://ruby-doc.org/core-2.3.0_preview1/Hash.html#method-i-dig) (thanks @braddunbar for the heads up).

Underscore could take a note from Ruby (as it's done with other methods) and have `_.get` and `_.set` methods which accept an array of properties to walk. No string magic needed. The use of an array, instead of individual arguments like Ruby's `Hash#dig`, allows for adding a `defaultValue` param after too.

```js
var object = {};
_.get(object, ['a', 'b', 'c'], 'defaultValue');
_.set(object, ['a', 'b', 'c'], 'newValue');
```

This deep property getter/setter support can also easily make its way into `_.pluck`, `_.result`, and other methods that access properties in the future if so desired too.

Related to:
#2370, #2268, #1169, #1266

akre54:
I can dig. Sort of like Immutable's [`getIn`](https://facebook.github.io/immutable-js/docs/#/Iterable/getIn) / `keyPath` methods.

Are we already overloading these APIs too much though?

jdalton:
The `_.get` and `_.set` methods don't exist in Underscore at the moment. The `_.result` method could be expanded to accept an array as the `property` param and others can be addressed on a case by case.

jridgewell:
I'd love to see these in the collection functions.

> Are we already overloading these APIs too much though?

I don't think so. I see this as translating primitives into the correct function for collection functions, with a few special cases for `_.result` and friends.

PR #2374, opened 2015-12-05
Implement deepFilter function

dchacke:
I needed something like this myself and though I'd share it. Based on https://github.com/IndigoUnited/js-deep-filter, but with a few tweaks.

`deepFilter` is pretty much the same as `filter`, except it goes through nested objects/arrays.

### DIFF
--- a/index.html
+++ b/index.html
@@ -202,6 +202,7 @@
       <li>- <a href="#reduceRight">reduceRight</a></li>
       <li>- <a href="#find">find</a></li>
       <li>- <a href="#filter">filter</a></li>
+      <li>- <a href="#deepFilter">deepFilter</a></li>
       <li>- <a href="#where">where</a></li>
       <li>- <a href="#findWhere">findWhere</a></li>
       <li>- <a href="#reject">reject</a></li>
@@ -551,6 +552,47 @@ var even = _.find([1, 2, 3, 4, 5, 6], function(num){ return num % 2 == 0; });
 var evens = _.filter([1, 2, 3, 4, 5, 6], function(num){ return num % 2 == 0; });
 =&gt; [2, 4, 6]
 </pre>
+
+    <p id="deepFilter">
+      <b class="header">deepFilter</b><code>_.deepFilter(list, predicate, [context])</code>
+      <br />
+      Performs a deep check through <b>list</b>, which can be an array or an
+      object, and returns a new array or object with only the values that pass
+      a truth test (<b>predicate</b>).
+    </p>
+    <pre>
+_.deepFilter([false, 1, [0, true, '']], function(value) {
+  return !!value;
+});
+=&gt; [1, [true]]
+
+_.deepFilter({ foo: 'bar', removed: 0, nested: { foo: 'bar', bar: 0 }, array: [0, 1, 2] }, function(value) {
+  return !!value
+});
+=&gt; { foo: 'bar', nested: { foo: 'bar' }, array: [1, 2] }
+</pre>
+
+    <p>
+      In addition to the current value, the <b>predicate</b> is also passed the
+      current path and the current parent array/object, also known as
+      <b>subject</b>.
+    </p>
+
+    <p>
+      The path can be used to find the current value's location in <b>list</b>.
+    </p>
+
+    <pre>
+_.deepfilter({ first: 'foo', second: 'foo', third: { nested: true, array: [1] } }, function(value, path, subject) {
+  // path will be 'first', 'second', 'third.nested', 'third.nested.array[0]',
+  // 'third.nested.array', and 'third' in each iteration respectively
+});
+</pre>
+
+    <p>
+      As can be seen in the paths' order, the <b>list</b> is traversed depth-
+      first.
+    </p>
 
       <p id="where">
         <b class="header">where</b><code>_.where(list, properties)</code>
--- a/index.html
+++ b/index.html
@@ -555,6 +555,7 @@ var evens = _.filter([1, 2, 3, 4, 5, 6], function(num){ return num % 2 == 0; });
 
     <p id="deepFilter">
       <b class="header">deepFilter</b><code>_.deepFilter(list, predicate, [context])</code>
+      <span class="alias">Alias: <b>deepSelect</b></span>
       <br />
       Performs a deep check through <b>list</b>, which can be an array or an
       object, and returns a new array or object with only the values that pass
--- a/test/collections.js
+++ b/test/collections.js
@@ -323,6 +323,78 @@
     assert.deepEqual(_(list).filter({}), list, 'OO-filter');
   });
 
+  test('deepFilter', function(assert) {
+    var object = {
+      stay: {
+        // Values that should stay
+        '0': true,
+        '1': 1,
+        '2': 'foo',
+        '3': {
+          foo: 'bar',
+          array: [
+            1,
+            {partOf: 'array'}
+          ]
+        }
+      },
+      remove: {
+        // Values that will be removed
+        '4': false,
+        '5': 0,
+        '6': null,
+        '7': {
+          nested: {
+            value: null
+          }
+        },
+        '8': [0, null, false, '']
+      }
+    };
+
+    var paths = [];
+
+    var result = _.deepFilter(object, function(value, path) {
+      paths.push(path);
+
+      return !!value;
+    });
+
+    assert.deepEqual(result.remove, {
+      '7': {
+        nested: {}
+      },
+      '8': []
+    });
+
+    assert.deepEqual(result.stay, object.stay);
+
+    assert.deepEqual(paths, [
+      'stay.0',
+      'stay.1',
+      'stay.2',
+      'stay.3.foo',
+      'stay.3.array[0]',
+      'stay.3.array[1].partOf',
+      'stay.3.array[1]',
+      'stay.3.array',
+      'stay.3',
+      'stay',
+      'remove.4',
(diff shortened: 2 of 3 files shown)


akre54:
Thanks for the pull but we don't do nested or deep methods yet. Let's leave this as a mixin for the time being.

jdalton:
Yet, we don't do them yet... #2372

akre54:
Right, that's what I said.

dchacke:
Why don't you folks do recursive methods?

akre54:
As far as I can tell, mostly for performance, size and correctness issues. There are all sorts of edge cases and undefined semantics associated with handling deeply nested objects / arrays. What happens if a property exists in one object but not the other? Do you want it to fail silently? Throw? Return an empty object? Default value? Should it use own keys or iterate prototype properties? And on and on.

It's often easier for you to handle this in your own code, especially since you know the structure of the object ahead of time.

jdalton:
I think many of these questions are similar to those answered in the `_.isEqual` method implementation. Now that there are reference implementations for deep cloning and merging and their customization of I think it takes much of the uneasiness away from recursive methods.

akre54:
Let's try the converse. What do you need a deep filter for that you couldn't do better and more clearly by hand?

jdalton:
> Let's try the converse. What do you need a deep filter for that you couldn't do better and more clearly by hand?

Me or @dchacke? We went from discussing nested or deep methods in general to recursive methods in general to `deepFilter` specifically? If you're looking for opinions on `deepFilter` specifically I think we should pull in the WordPress calypso folks (@ockham) since they use a `deepMapValues` and maybe @marklagendijk the dev behind [lodash-deep](https://github.com/marklagendijk/lodash-deep).

My gut is as with other util methods, that could be implemented by hand, it's for convenience. Not having to worry about circular references, tracking objects, low level object iteration scaffolding, etc.

jdalton:
For me I could see a low level object crawling/walking util method in my future which could be leveraged by methods like a `deepFilter` because of the number of requests for deep object iteration.

megawac:
I've bounced back and forth on opinion on this topic (deep method support in general), from wanting its inclusion to being against string programming to supporting it again. Anyway, I think its currently a killer feature in lodash and can make life much easier.
I was against string programming when most of the deep implementations were overly naive (i.e. when they were being considered for lodash); however, @jdalton , @alphagit, and others have shown it can be made quite robust. I would be very much in favour of their inclusion, starting with [`_.get`](https://lodash.com/docs#get) and [`_.set`](https://lodash.com/docs#set).

akre54:
> currently a killer feature in lodash

How are you using it that makes it so useful?

megawac:
Even the most trivial case is pretty useful

`_.has(obj, 'a["b"].c')` over `_.has(obj, 'a') && _.has(obj.a, 'b') && _.has(obj.a.b, 'c')`. 

What makes these features so useful where I work is we've monkey patched some BB methods (slightly) to be implemented through `_.has`, `_.set`, and `_.get` so they support nesting out of the box. Allowing things like `let namelist = UserCollection.map('user.contact.firstname')` 

Also on behaviour, I think lodash has gotten it right: returning undefined when a deep lookup does not match an item is what a user would expect. If they wanted an error they would do the lookup using javascript directly (`a.b.c`)

jridgewell:
The only part I dislike about deep properties is the `'a["b"].c'` syntax. I'd much rather the array style `['a', 'b', 'c']` like https://github.com/jashkenas/underscore/issues/2372.

akre54:
I dunno. I think if you need more than two levels of checks you're doing something wrong. Code like that is horrible to read, and destructuring makes it much nicer to deal with.

Also, what about an example with a deep filter? 

> I dislike about deep properties is the 'a["b"].c' syntax.

Agreed, it's pretty horrible. The array syntax is a bit nicer, but it's still magic over what is a relatively simple language feature. What's a good use for it in filter?

jdalton:
> I dunno. I think if you need more than two levels of checks you're doing something wrong. Code like that is horrible to read, and destructuring makes it much nicer to deal with.

Deep property look ups are a thing and normally they can look gross which is why util methods to handle it are a good thing to have.

> What's a good use for it in filter?

I've had requests for an object compact methods (Filter + `Boolean`) and could see that expanded to a deep form too. I've had co-workers deal with large JS data objects and have to clean and sanitize them and I could see a deep filter as useful there too.

jdalton:
@megawac @jridgewell 
> The only part I dislike about deep properties is the 'a["b"].c' syntax. I'd much rather the array style ['a', 'b', 'c'] like #2372.

I think your dot/bracket string example is a bit over the top but
for what it's worth lodash supports the array style too.
```js
 _.has(obj, [ 'a', 'b', 'c'])
```

AlphaGit:
Thanks @megawac for the nice words. Definitely appreciated. =)

I'm going to weigh in if you guys don't mind.

There are a few topics being discussed in here. 

**1.a. Inclusion of string-path based functions: why?**

Mostly metadata or dynamic data usage. Given the flexible nature of objects in JavaScript, it is quite likely that you'll end up inspecting or working with a structure to which you don't know exactly what shape it has. It is either is given to you in a different format, or in a way of a pointer -- that may be not other than a string.

My case was exactly that: I had a dynamic structure that came from another service in JSON formt and I had to extract certain properties coming from configuration. This configuration would have the shape of a string path that looked natural to a javascript developer. `myObject.deepProperty.somethingElse[2]`. Alternatively, the array syntax can definitely be used `['myObject', 'deepProperty', 'somethingElse', '2']`, but it is not so readable and that `2` just looks awkward on its own. Mixing them up to have `somethingElse[2]` is doable but just calls for trouble when the first option was available.

Using evals is out of the question. String parsing it would become a util library on its own and there's possibly some other library functions which you would be using with the data, in which underscore/lodash/whatevs is used. It _feels_ like the feature should be part of the library too at that point.

**1.b. Inclusion of string-path based functions: implementation**

The implementations can be leaky if they don't properly manage properly the situations which you can find with an object you can't trust. This should be pretty much standard for any of these libraries, except that they usually delegate the behavior to the consumer by passing a callback.

There's no need for recursion as you are just going deep down, property by property. It can definitely be an iterative process. And even better, it can be aborted prematurely as soon as you reached an unexpected situation.

If you use the array notation, this is not a problem. You just consume each property one by one and go down.

JavaScript notation is a little bit more difficult but not complicated really. In JavaScript you may access a property in each of the following:

- dot notation (`object.property`)
- indexing (`object["property"]`, `object['property']`)

The challenging part for us ([we did some of this in underscore-contrib](https://github.com/documentcloud/underscore-contrib/pull/175)) was to cleanly get the list of properties and still support came-from-hell-property names. When we realized that it's just the delimiters of the property that play into it, we ended up with a -simple- regex that got all that we needed.

**2. Behavior and consumer expectation**

- "Delegating to the prototype vs. getting from the actual object". I don't see any reason why you'd expect `_.get('property1.property2')` to _not_ delegate to the prototype, as just calling `object.property1.property2` would do. In the same fashion that JavaScript objects include a special method for checking own properties, I'd go with the same approach and provide an option, argument or another function that behaves in that way.

- "throwing vs. returning undefined", again, I would lean on how JavaScript behaves by default. If you try to read a property that does not exist, you'd get undefined. I'd lean towards undefined too.

You may argue that if you try to access two-level properties on a non existent one, JavaScript will throw and then you should too, but if I'm doing this with a path that I may not control, sometimes throwing and sometimes undefined would feel really inconsistent, and if you made me check the length of the path (making me parse it myself) and you made me check both against an error and an undefined value, I'd probably hate you in 50 shades of anger.

- Processing inconsistent objects, how to behave?

```js
var obj1 = { level1: { level2: { value: 0 } };
var obj2 = { level1: null };

var myObjs = [obj1, obj2];
// made up function
_.deepFilterByValueEquals(myObjs, 'level1.level2.value', 1);
```

I would bet most would expect to have an empty array back. Not to throw, not to return undefined and not to have 

**3. Deep filter itself** (current implementation in PR)

I am not completely sold on this one, but because I believe the name to be confusing. It took me a little bit of looking at this example to see what it meant:

```js
_.deepFilter([false, 1, [0, true, '']], function(value) {
  return !!value;
});
=> [1, [true]]
```

I'm trying to imagine scenarios were you would find situations like these, and more than 2 levels deep. (Otherwise, it's just a special case and not a generic problem to solve.) It sounds like a great idea, but maybe too specific to the particular domain that requires it. 

Furthermore, it's likely that it's not entirely useful because there's some condition that you may want to apply first to reject/accept the full array regardless of its inner elements, and then you'd have a `_.maybeDeepFilter`, and that's just messy.

ockham:
Thanks for adding me to this discussion, @jdalton.

FWIW,  inside Calypso, we're using lodash-deep's `deepMapValues()` only for [one unit test](https://github.com/Automattic/wp-calypso/blob/master/client/lib/menu-data/test/test-menu-data.js) (actually abusing it to a nested `forEach()` of sorts). We are however using some hand-written functions to process nested structures in other parts, see https://github.com/Automattic/wp-calypso/blob/master/client/lib/tree-convert/tree-traverser. Some of those methods are a bit specific to our use case, but I'd very much appreciate if there were some generic functions available for us to use with those nested structures.

I agree that as soon as more than one nested object are involved -- such as in comparison functions -- things can get messy, so what about starting with functions that act only on one such object? Even some simple, non-modifying functions such as `forEach()`, `find()`/`some()`/etc or `map()`, and, yeah, `filter()` could be quite useful.

PR #2380, opened 2015-12-10
Added _.resultAll

misantronic:
Invokes all methods inside of object and returns a new object


jridgewell:
This is a bit too special purpose. I'd recommend using [`_.mixin`](http://underscorejs.org/#mixin) to add it to your own codebase.

PR #2383, opened 2015-12-10
Dubious Improvement: Combine _.every and _.some logic

captbaritone:
These two functions were nearly identical. Inspired by #2060, I thought I would
try to combine the two to save some bytes. Ironically, repetitive code gzips
very well, so this refactor actually ends up _adding_ 3 bytes to the
minified-gzipped size. Oh well.

Never the less, I figured I'd open the PR, since it does reduce repetition in
the source code, and maybe it will serve as a guide for somebody else who looks
into combing these.

Pros:

* Less code in underscore.js
* DRYer code (https://en.wikipedia.org/wiki/Don't_repeat_yourself)

Cons:

* Larger gzipped file size
* Possibly a _tiny_ bit slower (I'm no perf expert)
* Code is harder to understand

Feel free to close this PR if, as I suspect, these trade-offs are not worthwhile.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -245,31 +245,28 @@
     return _.filter(obj, _.negate(cb(predicate)), context);
   };
 
+  // Generator function to create the some and every functions. `breakValue` is
+  // the boolean predicate response that should cause an early return.
+  var createSomeOrEvery = function(breakValue) {
+    return function(obj, predicate, context) {
+      predicate = cb(predicate, context);
+      var keys = !isArrayLike(obj) && _.keys(obj),
+          length = (keys || obj).length;
+      for (var index = 0; index < length; index++) {
+        var key = keys ? keys[index] : index;
+        if (!!predicate(obj[key], key, obj) === breakValue) return breakValue;
+      }
+      return !breakValue;
+    };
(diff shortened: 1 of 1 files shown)


megawac:
Something similar was proposed a while ago https://github.com/jashkenas/underscore/pull/1637, which got accepted and then reverted. We will need some benchmarks before considering this change

captbaritone:
@megawac My initial approach was similar to #1637's but `_negate` adds an extra function call in the iteration loop, which I assume is part of perf issues it encountered. My approach should avoid that cost.

I agree perf tests would be a good idea. I'll try to create a jsperf using the one in #1637 as a guide.

captbaritone:
http://jsperf.com/combine-every-and-some-logic

I get the following operations/second for `_.some()`:

* Stable: 500,139
* Edge: 457,829
* This PR: 444,623 
* Lodash: 609,778
* #1637: 13,321 

I assume `_.every()` should be about the same.

captbaritone:
Closing this. My change managed to be slower, larger and more confusing. That's too steep a price to pay for "slightly less DRY".

PR #2385, opened 2015-12-11
trust npm to set up $PATH

craigmichaelmartin:
@megawac Same change as https://github.com/jashkenas/backbone/pull/3885


akre54:
Yup, good catch.

PR #2388, opened 2015-12-14
Use assert.raises instead of .throws

captbaritone:
In some environments `.throws()` is a reserved word, so QUnit offers an alias `.raises()`.

Using `.raises` seems safer, and also allows us to remove an eslint exception.

### DIFF
--- a/test/.eslintrc
+++ b/test/.eslintrc
@@ -21,7 +21,6 @@
   },
   "rules": {
     "brace-style": 0,
-    "dot-notation": 1,
     "no-new-wrappers": 0,
     "no-sparse-arrays": 0,
     "no-extend-native": 0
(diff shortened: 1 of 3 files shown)


captbaritone:
See the blue box about half way down this page for the documentation of this alias: http://api.qunitjs.com/throws/

captbaritone:
We originally made the exact inverse of this change here, because QUnit deprecated `raises`: https://github.com/jashkenas/underscore/pull/1828

Looks like they've since brought it back: https://github.com/jquery/qunit/issues/663

michaelficarra:
Can you name a particular browser/version that has a problem with this?

captbaritone:
The QUnit docs mention Closure Compiler, which is presumably not a problem for our tests. I think the main advantage here, is that is allows us to re-enable the eslint setting as a rule instead of a warning.

PR #2389, opened 2015-12-14
Use same version of QUnit in browser as CLI

captbaritone:
In our `package.json` we list `"qunitjs": "^1.18.0"`, but the version in `tests/vendor` was only 1.17.1. This brings our browser test runner up to date.

From the QUnit changelog:

> ## 1.18.0
>
> - Assert: throws uses push method only
> - Assert: Fix missing test on exported throws
> - Assert: Implements notOk to assert falsy values
> - Core: More graceful handling of AMD
> - Core: Simplify stack trace methods
> - Core: Expose Dump maxDepth property
> - Core: Expose QUnit version as QUnit.version property
> - Core: Handle multiple testId parameters
> - Dump: Fix .name/.property doublettes
> - HTML Reporter: New diff using Google's Diff-Patch-Match Library
> - HTML Reporter: Make it more obvious why diff is suppressed.
> - HTML Reporter: Change display text for bad tests
> - HTML Reporter: Fix checkbox and select handling in IE <9
> - HTML Reporter: Fix test filter without any module
> - HTML Reporter: Retain failed tests numbers
> - Test: lowercase the valid test filter before using it

https://github.com/jquery/qunit/releases/tag/1.18.0

### DIFF
--- a/test/vendor/qunit.css
+++ b/test/vendor/qunit.css
@@ -1,12 +1,12 @@
 /*!
- * QUnit 1.17.1
+ * QUnit 1.18.0
  * http://qunitjs.com/
  *
  * Copyright jQuery Foundation and other contributors
  * Released under the MIT license
  * http://jquery.org/license
  *
- * Date: 2015-01-20T19:39Z
+ * Date: 2015-04-03T10:23Z
  */
 
 /** Font Family and Sizes */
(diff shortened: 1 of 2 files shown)


akre54:
Thanks :)

PR #2390, opened 2015-12-15
Upgrade ESLint

captbaritone:
The new version of ESLint has all rules off by default, which matches our existing usage of the tool.

Some rules have been renamed, or their behavior has changed slightly. Most obviously, the `indent` rule is no-longer as lax about the indention of variable declarations.

I think upgrading the tool is a net win, and the changes (while arguably trivial to the point of annoyance) do at least represent an increase in consistency.


michaelficarra:
Cool, thanks!

akre54:
Can either one of you, who know a bit more about eslint than I do, port this over to Backbone? The rules should be similar.

craigmichaelmartin:
:+1: Backbone supposedly lints with `jsl` which I've never heard of and isn't even listed as a dependency.

akre54:
[jsl](http://www.javascriptlint.com) is an (old old) linter for JS, from back when you used to download js in `.tar.gz` archives (hence the global install). A lot of the rules for Underscore should be pretty similar to the ones for Backbone.

captbaritone:
I can take a look at setting up eslint for Backbone using this `.eslintrc` as a starting point.

akre54:
Thanks!

PR #2392, opened 2015-12-16
Add reversed option for sortBy

uruz:
It would be nice to have an option to sort the array with `_.sortBy` in a descending order, like it's done in Python's sorting function for lists.


akre54:
I'm afraid we can't just break argument order like that, so this is a no-go out of the gate. The simplest solution is to do `_.sortBy(arr, key).reverse()` if you need it.

This issue has been brought up a couple times in the past. Please do a search through the issues for some background.

jdalton:
There's been some discussion in [underdash](https://github.com/underdash/underdash/) about a `sortByOrder` method which can accept "asc" and "desc". I wonder if a better name is `orderBy`.

PR #2393, opened 2015-12-17
Improve example for _.every()

captbaritone:
Using `_.identity` (the default predicate) as an example predicate is confusing. It implies that `predicate` is actually a required argument and if you want to evaluate the values untransformed, you must pass `_.identity` explicitly.


akre54:
Looks good, thanks.

PR #2394, opened 2015-12-17
lift, unlift, and orElse, for use of map/orElse in functional style.

zakpatterson:
Compare:

```
const doStuff = arr => {
  let res;
  if (arr.length > 10) {
    res = "long array";
  } else {
    res = "short array";
  }
  return res;
}
```
to:

```
const doStuff2 = arr => (
  _.lift(arr.length)
  .filter(len => len > 10)
  .map(() => "long array")
  .orElse(() => "short array")
  .unlift()
)
```

```
doStuff([1,2,3,4,5,6,7,8]);
> "short array"
doStuff2([1,2,3,4,5,6,7,8]);
> "short array"
```

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -186,16 +186,12 @@
 
   _.orElse = function(obj, iteratee, context) {
     var results;
-    console.log("CHECKING OR ELSE...");
-    console.log(obj);
     iteratee = cb(iteratee, context);
     if (_.isEmpty(obj)) {
       results = [iteratee()];
     } else {
       results = obj;
     }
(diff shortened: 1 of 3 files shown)


underscore.js
```diff
@@ -184,6 +184,21 @@
     return results;
   };
 
+  _.orElse = function(obj, iteratee, context) {
+    var results;
+    console.log("CHECKING OR ELSE...");
+    console.log(obj);
+    iteratee = cb(iteratee, context);
```
jdalton: console logs in the method?

PR #2395, opened 2015-12-17
lift, unlift, and orElse, for use of map/orElse in functional style.

zakpatterson:
Compare:

```
const doStuff = arr => {
  let res;
  if (arr.length > 10) {
    res = "long array";
  } else {
    res = "short array";
  }
  return res;
};
```
to:

```
const doStuff2 = arr =>
  _.lift(arr.length)
  .filter(len => len > 10)
  .map(() => "long array")
  .orElse(() => "short array")
  .unlift();
```

```
doStuff([1,2,3,4,5,6,7,8]);
> "short array"
doStuff2([1,2,3,4,5,6,7,8]);
> "short array"
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -331,6 +331,7 @@
       <li>- <a href="#result">result</a></li>
       <li>- <a href="#now">now</a></li>
       <li>- <a href="#template">template</a></li>
+      <li>- <a href="#orElse">orElse</a></li>
     </ul>
 
     <a class="toc_title" href="#chaining">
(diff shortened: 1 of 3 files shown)


akre54:
That isn't a particularly compelling example. You could write that as a simple one-line ternary:

```js
const doStuff = arr => arr.length > 10 ? long array" : "short array"
```

Regardless, this is a bit too edge-case for core Underscore but might make a fine mixin.

PR #2398, opened 2015-12-18
Remove _.where from _.matcher tests

captbaritone:
Avoids making assumptions about _.where's implementation.


akre54:
Could we do the same by removing `_.filter` from these tests too? Like maybe actually just test what `matcher` does instead of relying on the behavior of others too?

captbaritone:
@akre54 It's not a bad idea, but to me, the assumption that `_.where` uses `_.matches` under the hood feels an order of magnitude more problematic than assuming that `_.filter` behaves as documented.

akre54:
I guess. Given that we know `filter` has to accept a function to work.

PR #2403, opened 2016-01-03
fixes #2402: symbol support for _.isFinite

### LINKED ISSUE #2402: Hard crash with isFinite and Symbol
Using:
* Node.js `v4.0.0`
* Underscore.js `1.8.3`
* and the code below...
```javascript
"use strict";

var _   = require('underscore');
var sym = Symbol();

console.log('sym',sym);

console.log(_.isFinite(sym));

```
...results with the following:
```
sym Symbol()
/test/node_modules/underscore/underscore.js:1242
    return isFinite(obj) && !isNaN(parseFloat(obj));
           ^

TypeError: Cannot convert a Symbol value to a number
    at isFinite (native)
    at Function._.isFinite (/test/node_modules/underscore/underscore.js:1242:12)
    at Object.<anonymous> (/test/test.js:8:15)
    at Module._compile (module.js:434:26)
    at Object.Module._extensions..js (module.js:452:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Function.Module.runMain (module.js:475:10)
    at startup (node.js:117:18)
    at node.js:951:3
```

michaelficarra:
Fixes #2402.

### DIFF
--- a/test/objects.js
+++ b/test/objects.js
@@ -728,6 +728,11 @@
     assert.ok(_.isFinite(0), '0 is finite');
     assert.ok(_.isFinite(123), 'Ints are finite');
     assert.ok(_.isFinite(-12.44), 'Floats are finite');
+    if (typeof Symbol === 'function') {
+      assert.ok(!_.isFinite(Symbol()), 'symbols are not numbers');
+      assert.ok(!_.isFinite(Symbol('description')), 'described symbols are not numbers');
+      assert.ok(!_.isFinite(Object(Symbol())), 'boxed symbols are not numbers');
+    }
   });
 
   QUnit.test('isNaN', function(assert) {
--- a/underscore.js
+++ b/underscore.js
@@ -1311,7 +1311,7 @@
 
   // Is a given object a finite number?
   _.isFinite = function(obj) {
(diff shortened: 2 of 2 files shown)


jdalton:
This would still fail `_.isFinite(Object(Symbol()))`. It be better long term to move to [Number.isFinite](http://www.ecma-international.org/ecma-262/6.0/#sec-number.isfinite) behavior (which avoids the issue).

michaelficarra:
Unfortunately, this function treats numeric strings as numbers, so we can't really do that. I was about to propose adding an `isSymbol` helper along the lines of `isNumber`, `isError`, etc., so if that is accepted I will change this PR to make use of that.

jdalton:
>  Unfortunately, this function treats numeric strings as numbers, so we can't really do that.

Oh I know. That's the nice thing about major bumps every once in a while.

jdalton:
>  I was about to propose adding an `isSymbol`

I added it to Lodash so :+1:

jridgewell:
@michaelficarra: `_.isSymbol` merged, want to update this code?

michaelficarra:
Thanks. I will update soon.

michaelficarra:
@jridgewell Updated.

PR #2406, opened 2016-01-04
Optimize differently-sized array equality

### LINKED ISSUE #2400: _.isEqual optimization: compare array length ASAP
`_.isEqual`'s helper function [checks the length of the arrays it's comparing](https://github.com/jashkenas/underscore/blob/a8c29420dc9e5decbbea2dff7e50527af70c8c03/underscore.js#L1234) when it's figured out it's comparing two arrays together.

This step, however, happens after [aStack/bStack has been initialized (and aStack has been filled)](https://github.com/jashkenas/underscore/blob/a8c29420dc9e5decbbea2dff7e50527af70c8c03/underscore.js#L1218-L1229).

Is there a specific reason it's this way, or would a PR (..that doesn't break the tests) changing it be welcome?

vendethiel:
Fixes #2400

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -1213,6 +1213,11 @@
     // Assume equality for cyclic structures. The algorithm for detecting cyclic
     // structures is adapted from ES 5.1 section 15.12.3, abstract operation `JO`.
 
+    // Compare array lengths to determine if a deep comparison is necessary.
+    if (areArrays && a.length !== b.length) {
+      return false;
+    }
+    
     // Initializing stack of traversed objects.
     // It's done here since we only need them for objects and arrays comparison.
     aStack = aStack || [];
@@ -1230,10 +1235,8 @@
 
     // Recursively compare objects and arrays.
     if (areArrays) {
-      // Compare array lengths to determine if a deep comparison is necessary.
-      length = a.length;
-      if (length !== b.length) return false;
       // Deep compare the contents, ignoring non-numeric properties.
+      length = a.length;
       while (length--) {
         if (!eq(a[length], b[length], aStack, bStack)) return false;
       }


underscore.js
```diff
@@ -1213,6 +1213,11 @@
     // Assume equality for cyclic structures. The algorithm for detecting cyclic
     // structures is adapted from ES 5.1 section 15.12.3, abstract operation `JO`.
 
+    // Compare array lengths to determine if a deep comparison is necessary.
+    if (areArrays && a.length !== b.length) {
+      return false;
+    }
+
```
michaelficarra: You have trailing whitespace here, causing CI to fail.

vendethiel: seems like github isn't taking my force-push into account?

michaelficarra: The latest I see is 7204d5b2. Is that the commit you're trying to force push? Make sure you have the correct remote.

vendethiel: The problem seems to be "from unknown repository" at the top of the PR. Guess I need to resubmit the PR, as that's not something I can change... Sigh.

michaelficarra: Oh, did you delete your fork?

megawac: Note, the `areArrays` check is not strictly necessary as `a.length !== b.length`, `a` and `b` are neq.

jdalton: > Note, the `areArrays` check is not strictly necessary as `a.length !== b.length`, `a` and `b` are neq.

It depends if `length` is non enum as comparisons are on enumerable own props.

megawac: I haven't reviewed the `eq` implementation in a while but I thought one of the constraints was the objects have the same constructor?

vendethiel: @megac what did you mean by "are neq" in your first comment?

megawac: not equal

vendethiel: ...which translates to which part of the code?

akre54:
Can you push again and we'll merge? Thanks.

vendethiel:
I'd like to know what @megawac meant in his comments – it seems he's saying mr PR doesn't change anything?

underscore.js
megawac: @vendethiel I simply meant that the condition can be simplified to 

```js
if (a.length !== b.length) {
      return false;
}
```

vendethiel: Well, I need to re-read the code (am on mobile) but there are "areArrays" check below-I'm afraid your code would make `{length: 3, a: 1}` count as an array... But it might already be the case

megawac: The thing is @vendethiel  `{length: 3, a: 1}` should be not equal to `{length: 4, a: 1}` or similar

PR #2409, opened 2016-01-09
Use canonical name for `_.uniq` in documentation

captbaritone:

PR #2412, opened 2016-01-13
Make _.unique the canonical name

captbaritone:
As requested in #2409 by @michaelficarra.

### DIFF
--- a/index.html
+++ b/index.html
@@ -237,7 +237,7 @@
       <li>- <a href="#union">union</a></li>
       <li>- <a href="#intersection">intersection</a></li>
       <li>- <a href="#difference">difference</a></li>
-      <li>- <a href="#uniq">uniq</a></li>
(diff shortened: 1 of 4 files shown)


michaelficarra:
I'd want to get a :+1: from the other maintainers before merging something like this, but LGTM.

jdalton:
lodash [v4](https://twitter.com/jdalton/status/687052646281158656) dropped the `unique` alias so :-1:

jridgewell:
I'm ambivalent to the change. Ruby uses `#uniq` so it's what I've always used. lodash compatibility is nice, too.

Are there any popular languages using `#unique`?

captbaritone:
> Are there any popular languages using #unique?
> 
The only one I was thinking of was English ;)

While unique seems slightly more intuitive to me, it's only a minor improvement. Increased compatibility with Lodash is probably far more valuable.

--
Jordan Eldredge

PR #2431, opened 2016-01-27
Separate object-oriented style and chaining in docs

captbaritone:
I think the fact that the object-oriented style and chaining share implementation details has lead us to conflate them in the docs. In practice, the two features have almost nothing to do with each other.

The existing documentation could easily lead one to believe that they could start a chain using the OOP style.

This confusion is compounded by a few factors:

1. Lodash supports chains starting with the OOP wrapper.
2. The fact that we share names with native Array methods can lead to manual
   tests that appear to work.

For example:

    _([1,2,3]).reverse().map(function(n){ return n * 2; });

This appears to be an example of successfully starting a chain with the OOP style. The astute observer will note that the lack of a `.value()` call, proves this is not Underscore chaining, but a new user attempting to clarify the behavior for herself may not catch that detail.

### DIFF
--- a/index.html
+++ b/index.html
@@ -334,6 +334,9 @@
       <li>- <a href="#template">template</a></li>
     </ul>
 
+    <a class="toc_title" href="#oop">
+      OOP Style
+    </a>
     <a class="toc_title" href="#chaining">
       Chaining
     </a>
@@ -1861,7 +1864,7 @@ obj.initialize = _.noop;
         <b>iteratee</b> is called with an <tt>index</tt> argument. Produces an
         array of the returned values.
         <br />
-        <i>Note: this example uses the <a href="#chaining">chaining syntax</a></i>.
+        <i>Note: this example uses the <a href="#oop">object-oriented syntax</a></i>.
       </p>
       <pre>
 _(3).times(function(n){ genie.grantWishNumber(n); });</pre>
@@ -2055,7 +2058,7 @@ _.template("Using 'with': <%= data.answer %>", {variable: 'data'})({answer: 'no'
 &lt;/script&gt;</pre>
 
 
-      <h2 id="chaining">Chaining</h2>
+      <h2 id="oop">Object-Oriented Style</h2>
 
       <p>
         You can use Underscore in either an object-oriented or a functional style,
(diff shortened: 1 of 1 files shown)


captbaritone:
If it would aid review, here is a link to my proposed version of the docs: https://rawgit.com/captbaritone/underscore/chain-oop-docs/index.html#oop

michaelficarra:
I'd rather remove all references to OOP and consistently just call it the "chaining style" or "chaining-friendly wrapper".

akre54:
Agreed

captbaritone:
Perhaps I am a bit confused. To me "chaining" means feeding the output of one method into the input of another. The `_([])` syntax allows you to call a single method, but it does not allow you to "chain" multiple methods together.

Calling `_([])` a "chaining-friendly wrapper" seems very misleading, since using it to start a chain with more than one "link" crashes. For example:

    _([1,2,3]).map(function(n){ return n * 2; }).first();

    =>  TypeError: _(...).map(...).first is not a function

captbaritone:
@akre54 Could I get a sanity check regarding my last comment?

akre54:
Yes, they are different uses. `_.chain` requires you to call `.value()` at the end to retrieve the result, while `_()` returns the array directly.

captbaritone:
And we still want to call `_()` "chaining", despite the fact that you can't use it to "chain" together multiple methods?

akre54:
Nah, you're right. I forgot that the OOP-style didn't automatically start a chain for you (I never use it).

michaelficarra:
Any suggestions for a better name though?

captbaritone:
While verbose, I personally think "object-oriented style" captures the idea quite well. Some alternatives:

* Wrapped style
* Method style

I'm not crazy about either of those. Anyone else have any ideas?

captbaritone:
@michaelficarra Do you like either of the above options, or is OOP the best we are going to get?

captbaritone:
Since I think this change is a meaningful improvement and does not introduce any new language. Could we merge this PR as is? If we think it's important to come up with a new name for "Object Oriented Style" we can open a separate issue to try to find a new name.

jridgewell:
LGTM. Maybe @megawac or @akre54 want to thumbs up as well?

michaelficarra:
That's fine by me.

jridgewell:
Whoops, forgot @michaelficarra. That makes two.

captbaritone:
Thanks!

PR #2433, opened 2016-01-29
Update README.md

jchiang7874:
I have added a description to the README.md that is accessible to non-developers, hopefully clarifying jargon. Do you think having this explanation is of value to this project?

(--Assignment for Learn Modern Developer)


akre54:
Thanks but no thanks. We want to leave the README as-is.

I think what we have on http://underscorejs.org/ is good enough. If you think there's something to improve, please suggest changes to index.html (your metaphor is a bit much though).

jchiang7874:
Thanks for the feedback regarding the metaphor.

PR #2435, opened 2016-02-04
Configurable option: on interpolation, ignore missing template variables

marseille:
I propose a configurable option here. The option to have underscore partially interpolate templates. 

The idea here is that all template variables are required to be present on interpolation. This relaxes that restraint and leaves a missing template variable exactly as it was before in the template. Leaving it for future interpolation. 

I use underscore in nodejs, as well as backbone in the same application. I have a centralized configuration file that the application uses. The template I use requires configuration information, and user inputted information. The template then is filled out in two stages, once in nodejs and finished in backbone. 

This really helps with with staged templating and accomplishes my and possibility other's use cases. 

The modification is to the **settings** object. 

The addition is the **ignore_missing_template_variables** key. It accepts a boolean value.

The added usage is as follows:

`// partial replace`
`var example_template = '<%= foo %><%= bar %>;`
`var settings = {'ignore_missing_template_variables': true};`
`var template = _.template(example_template, settings);`
`var example_templated = template({'foo': '123'}); // '123<%= bar%>'`

`// ignore variable`
`var example_template = '<%= foo %>;`
`var settings = {'ignore_missing_template_variables': true};`
`var template = _.template(example_template, settings);`
`var example_templated = template({}); // '<%=foo%>';`

This option can be disabled or omitted for original functionality, and custom interpolation tags are preserved.

### DIFF
--- a/test/utility.js
+++ b/test/utility.js
@@ -428,6 +428,6 @@
     assert.expect(1);
     var settings = {'ignore_missing_template_variables': true};
     var template = _.template('<%=foo%><%=bar%>', settings);
-    assert.deepEqual(template({"foo": "Handle"}), 'Handle<%=bar%>');
+    assert.deepEqual(template({'foo': 'Handle'}), 'Handle<%=bar%>');
   });
 }());
--- a/test/utility.js
+++ b/test/utility.js
@@ -417,4 +417,17 @@
     assert.strictEqual(template(), '<<\nx\n>>');
   });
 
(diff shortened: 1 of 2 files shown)


marseille:
Working through the linting and coding standard issues, pardon the delay; I'm acclimating to this.

marseille:
The build passes now yay~

akre54:
I feel like Underscore templates are the wrong tool for this, and you may want to rethink your design. This is enough of an edge case that you may want to make your own library for this.

PR #2436, opened 2016-02-08
Purpose readme contribution

olegrjumin:
To make Underscore.js more approachable to a larger number of users—from informed developers to curious passersby and new and interested visitors—I have added a new section to the Underscore.js README. The section is entitled What to Use Underscore.js for and When to Use It. As the name suggests, the new section, a brief paragraph, explains in plain and straightforward language what Underscore.js is used for and when one should use it.

Other repositories may begin to use this new informative and catchy section soon, as my colleagues contact open-source technologies in an effort to make them more accessible to the tens of thousands of new users learning to code and thousands joining GitHub every day.

PR #2438, opened 2016-02-08
Add a purpose section to README

jmatembu:
To make Underscore more approachable to a larger number of users—from informed developers to curious passersby and new and interested visitors—I have added a new section to the Underscore README. The section is entitled What to Use Underscore for and When to Use It. As the name suggests, the new section, explains in plain and straightforward language what Underscore is used for and when one should use it. This is in an effort to make Underscore more accessible to the tens of thousands of new users learning to code and thousands joining GitHub every day.


jdalton:
Dup of #2436

PR #2439, opened 2016-02-13
Add _.sort as alias of _.sortBy

captbaritone:
When using the `_.identity` feature of `_.iteratee`, the name "sortBy" does not
make sense.

    // Confusing:
    _.sortBy([5, 4, 3]);
    => [3, 4, 5]

    // Better
    _.sort([5, 4, 3]);
    => [3, 4, 5]

### DIFF
--- a/index.html
+++ b/index.html
@@ -689,6 +689,7 @@ _.min(numbers);
 
       <p id="sortBy">
         <b class="header">sortBy</b><code>_.sortBy(list, iteratee, [context])</code>
+        <span class="alias">Alias: <b>sort</b></span>
         <br />
         Returns a (stably) sorted copy of <b>list</b>, ranked in ascending
         order by the results of running each value through <b>iteratee</b>.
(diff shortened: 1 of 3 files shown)


jdalton:
There's already a `sort` as part of the chaining syntax which defers to the `Array#sort` built-in.

captbaritone:
So the idiomatic way to write this would be to use one of these options?

    [5, 4, 3].sort();

Or:

    _([5, 4, 3]).sort();

captbaritone:
@jdalton For context, I'm working on this rule for my linting plugin: https://github.com/captbaritone/eslint-plugin-underscore/blob/master/docs/rules/identity-shorthand.md and I want to make sure I don't create a rule that suggests something confusing.

Currently it suggests converting:

    _.sortBy([5,4,3], function(n) { return n; });

To: 

        _.sortBy([5,4,3]);

Which seems a bit odd.

jdalton:
Your linting plugin is correct. Because the default iteratee is `identity`.

PR #2440, opened 2016-02-14
Remove superfulous _.identitys

captbaritone:
By removing these we avoid unnecessarily transforming out predicate through `optimizeCb()`.


michaelficarra:
Thanks @captbaritone.

jdalton:
We'll need to be aware of this change if Underscore is updated to use `_.iteratee` internally because folks could customize it to not default to `_.identity` if a function isn't provided.

PR #2452, opened 2016-02-29
Two symbols are equal by valueOf(). Closes #2450

### LINKED ISSUE #2450: Symbol support in isEqual
```js
var symbol = Symbol('x');
_.isEqual(symbol, Object(symbol));
// => should be true, but is false.
```

akre54:


megawac:
:+1:

test/objects.js
```diff
@@ -565,6 +565,14 @@
     assert.equal(_.isEqual({a: 0}, {a: -0}), false);
     assert.equal(_.isEqual([NaN], [NaN]), true);
     assert.equal(_.isEqual({a: NaN}, {a: NaN}), true);
+
+    if (typeof Symbol !== 'undefined') {
+      var symbol = Symbol('x');
+      assert.strictEqual(_.isEqual(symbol, symbol), true, 'A symbol is equal to itself');
+      assert.strictEqual(_.isEqual(symbol, Object(symbol)), true, 'Even when wrapped in Object()');
+      assert.strictEqual(_.isEqual(symbol, null), false, 'Different types are not equal');
```
jdalton: Should add a test for `_.isEqual(Symbol('x'), Symbol('y'))`.

megawac: Yeah probably for `_.isEqual(Symbol('a'), Symbol('a'))` as false as well

jdalton: :clap:

underscore.js
```diff
@@ -20,6 +20,7 @@
 
   // Save bytes in the minified (but not gzipped) version:
   var ArrayProto = Array.prototype, ObjProto = Object.prototype;
+  var SymbolProto = typeof Symbol !== 'undefined' ? Symbol.prototype : null;
```
jdalton: I donno how much of a concern this is but the `undefined` check alone could let a `null` value through and cause a script error at that point.

wwwoodall: Hey @jdalton, curious what you meant by this

> the `undefined` check alone

and what a possible solution might be.

PR #2459, opened 2016-03-05
Clean up text wrapping

captbaritone:
Clean up related to #2219

@jridgewell Not sure if I should do things like this as PRs, or just push them. Your thoughts would be appreciated.


megawac:
Feel free to just push nitty stuff like formatting

captbaritone:
:+1:

jridgewell:
Yah, we don't need at approve formatting/grammar/etc unless it's massive.

PR #2460, opened 2016-03-06
Allow a specified depth for flatten

paulfalgout:
Per a new TC39 proposal http://bterlson.github.io/proposal-flatMap/#sec-Array.prototype.flatten they're suggesting a depth argument.

This was fairly trivial to implement without breaking the current functionality.

Thoughts?

### DIFF
--- a/test/arrays.js
+++ b/test/arrays.js
@@ -90,6 +90,8 @@
     assert.deepEqual(result, [1, 2, 3, 4], 'works on an arguments object');
     list = [[1], [2], [3], [[4]]];
     assert.deepEqual(_.flatten(list, true), [1, 2, 3, [4]], 'can shallowly flatten arrays containing only other arrays');
+    list = [1, [2], [[3]], [[[4]]]];
+    assert.deepEqual(_.flatten(list, 2), [1, 2, 3, [4]], 'can flatten arrays to a given depth');
 
     assert.equal(_.flatten([_.range(10), _.range(10), 5, 1, 3], true).length, 23, 'can flatten medium length arrays');
     assert.equal(_.flatten([_.range(10), _.range(10), 5, 1, 3]).length, 23, 'can shallowly flatten medium length arrays');
--- a/test/arrays.js
+++ b/test/arrays.js
@@ -92,6 +92,8 @@
     assert.deepEqual(_.flatten(list, true), [1, 2, 3, [4]], 'can shallowly flatten arrays containing only other arrays');
     list = [1, [2], [[3]], [[[4]]]];
     assert.deepEqual(_.flatten(list, 2), [1, 2, 3, [4]], 'can flatten arrays to a given depth');
+    assert.deepEqual(_.flatten(list, 0), list, 'can flatten arrays to depth of 0');
+    assert.deepEqual(_.flatten(list, -1), list, 'can flatten arrays to depth of -1');
 
     assert.equal(_.flatten([_.range(10), _.range(10), 5, 1, 3], true).length, 23, 'can flatten medium length arrays');
     assert.equal(_.flatten([_.range(10), _.range(10), 5, 1, 3]).length, 23, 'can shallowly flatten medium length arrays');
--- a/underscore.js
+++ b/underscore.js
@@ -522,7 +522,11 @@
 
   // Flatten out an array, either recursively (by default), or just one level.
   _.flatten = function(array, depth) {
-    if (!depth) depth = Infinity;
+    if (!depth && depth !== 0) {
+      depth = Infinity;
+    } else if (depth <= 0) {
+      return _.clone(array);
+    }
     return flatten(array, depth, false);
   };
 
--- a/underscore.js
+++ b/underscore.js
@@ -499,19 +499,19 @@
   };
 
   // Internal implementation of a recursive `flatten` function.
-  var flatten = function(input, shallow, strict, output) {
+  var flatten = function(input, depth, strict, output) {
     output = output || [];
     var idx = output.length;
     for (var i = 0, length = getLength(input); i < length; i++) {
       var value = input[i];
       if (isArrayLike(value) && (_.isArray(value) || _.isArguments(value))) {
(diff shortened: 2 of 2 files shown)


akre54:
Awesome. :+1:

megawac:
:+1:

underscore.js
```diff
@@ -521,8 +521,9 @@
   };
 
   // Flatten out an array, either recursively (by default), or just one level.
-  _.flatten = function(array, shallow) {
-    return flatten(array, shallow, false);
+  _.flatten = function(array, depth) {
+    if (!depth) depth = Infinity;
```
jdalton: Should a `depth` of `0` be seen as `Infinity`?

paulfalgout: so you think we need a `if (depth === 0) return array;` ?

akre54: :thumbsup:

jdalton: Should return a shallow clone/slice which should shake out naturally if `0` is allow to be used as a depth (no need for a `if` or special branch)

test/arrays.js
```diff
@@ -90,6 +90,8 @@
     assert.deepEqual(result, [1, 2, 3, 4], 'works on an arguments object');
     list = [[1], [2], [3], [[4]]];
     assert.deepEqual(_.flatten(list, true), [1, 2, 3, [4]], 'can shallowly flatten arrays containing only other arrays');
+    list = [1, [2], [[3]], [[[4]]]];
+    assert.deepEqual(_.flatten(list, 2), [1, 2, 3, [4]], 'can flatten arrays to a given depth');
```
jdalton: Should add a test for a `-1` depth (what happens?)

underscore.js
paulfalgout: so then `if (!depth && depth !== 0) return`

jdalton: @paulfalgout To preserve as much back compat ya I think that would do.

paulfalgout:
So looking into both 0 and negative values, it seems much cleaner to handle this in `_.flatten` than it would be to try to get the internal flatten to play nice with these additional scenarios, particularly related to the "strict" argument. I also assumed that negative would act the same as 0?

jdalton:
>  it seems much cleaner to handle this in _.flatten than it would be to try to get the internal flatten to play nice with these additional scenarios

Oh cool!

>  I also assumed that negative would act the same as 0

Yeah.

underscore.js
```diff
@@ -521,8 +521,13 @@
...
   // Flatten out an array, either recursively (by default), or just one level.
-  _.flatten = function(array, shallow) {
-    return flatten(array, shallow, false);
+  _.flatten = function(array, depth) {
+    if (!depth && depth !== 0) {
+      depth = Infinity;
+    } else if (depth <= 0) {
+      return _.clone(array);
+    }
+    return flatten(array, depth, false);
```
michaelficarra: Might want to ToNumber `depth` here so we're not passing `true` to `flatten`.

michaelficarra:
LGTM :+1:

underscore.js
paulfalgout: we're internally calling `flatten` with booleans, so I don't know if we want to enforce that at the moment?

underscore.js
```diff
@@ -521,8 +521,13 @@
   };
 
   // Flatten out an array, either recursively (by default), or just one level.
-  _.flatten = function(array, shallow) {
-    return flatten(array, shallow, false);
+  _.flatten = function(array, depth) {
+    if (!depth && depth !== 0) {
```
jridgewell: This conditional confuses the hell out of me. Maybe:

```js
if (depth === false) {
  depth = Infinity;
} else if (depth === true) {
  depth = 1;
} else if (depth < 0) {
  depth = 0;
}
```

paulfalgout: the internal `flatten` doesn't handle depth 0 at all and will conflict with whatever "strict" is doing.
`depth` as with `shallow` is optional so the default value is `Infinity`, but to keep with the current API `false` should also be assumed as `Infinity` so I suppose it could be:
```js
if(depth === false || _.isUndefined(depth)) {
  depth = Infinity;
} else if (depth <= 0) {
  return _.clone(array);
}
```
We can certainly add the `else if (depth === true) {` but `flatten` is called internally with `true` elsewhere.

paulfalgout:
This was initially easy, but got complicated with edge cases and without refactoring internals a bit.
Any ideas what we should do with this PR? I've answered the two outstanding questions, but I'm not certain they're resolved.

underscore.js
megawac: can we updat ethe internal calls to use a number

jgonggrijp:
@paulfalgout I merged #2849 three weeks ago and I nearly forgot to get back to you. Sorry about that.

So `master` should be stable again for a while and I see no reason not to finish your PR. With some luck, we should be able to include it in the upcoming 1.11 release. In order to do so, we need to update your patch first. Please let me know whether you'd like to do this yourself, or whether you'd prefer me to take care of these final steps:

1. Check out the feature branch and pull in the latest changes from `master`. Disregard the merge conflict for now and postpone the merge commit until step 5.
2. Run a fresh `npm install` to update all the tools and install our commit hooks. These will take care of updating the `underscore.js`.
3. Update `modules/_flatten.js` and `modules/flatten.js` (note the underscore) with the changes that you originally made in the old `underscore.js`.
4. Run `git add underscore.js` to silence the warnings about the merge conflict. The conflict markers will be overwritten by the pre-commit hook.
5. Commit and push.

jgonggrijp:
I'm now looking into merging this by executing the above steps myself. On closer inspection, I noticed that many internal calls to `flatten` use the *internal* flatten with a boolean. I think I'll address this by moving the guard that changes `false` into `Infinity` from the public to the internal `flatten`.

paulfalgout:
@jgonggrijp thanks for hopping on this..  I've been a little out of it as of late.

jgonggrijp:
Welcome.

PR #2461, opened 2016-03-07
Improves `_.memoize` performance

caiogondim:
Creating an object without prototype proves to be faster than a literal
object, since we don't have to go through the prototype chain to check
if a key already exists.

With a memoized recursive fibonacci implementation, we could do on a
MacBook 13 Retina Late 2013 ~9.6M ops/sec, while with the improved
implementation we could get ~26M ops/sec. Almost a 3x improvement.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -779,13 +779,20 @@
 
   // Memoize an expensive function by storing its results.
   _.memoize = function(func, hasher) {
+    function getCache() {
+      if (Object.create) {
+        return Object.create(null);
+      }
+      return {};
+    }
+
     var memoize = function(key) {
       var cache = memoize.cache;
       var address = '' + (hasher ? hasher.apply(this, arguments) : key);
-      if (!_.has(cache, address)) cache[address] = func.apply(this, arguments);
+      if (typeof cache[address] === 'undefined') cache[address] = func.apply(this, arguments);
       return cache[address];
     };
-    memoize.cache = {};
+    memoize.cache = getCache();
     return memoize;
   };
 
--- a/underscore.js
+++ b/underscore.js
@@ -786,23 +786,14 @@
       return {};
     }
 
-    function hasKey(cache, address) {
-      // Objects created with `Object.create(null)` don't have a constructor
-      // property.
-      if (cache.constructor === Object) {
-        return _.has(cache, address);
-      } else {
-        return (typeof cache[address] !== 'undefined');
-      }
-    }
-
     var memoize = function(key) {
       var cache = memoize.cache;
       var address = '' + (hasher ? hasher.apply(this, arguments) : key);
-      if (!hasKey(cache, address)) cache[address] = func.apply(this, arguments);
+      if (!(_.has(cache, address))) cache[address] = func.apply(this, arguments);
       return cache[address];
     };
     memoize.cache = getCache();
+
     return memoize;
   };
 
--- a/underscore.js
+++ b/underscore.js
@@ -786,10 +786,20 @@
       return {};
     }
 
+    function hasKey(cache, address) {
+      // Objects created with `Object.create(null)` don't have a constructor
+      // property.
+      if (cache.constructor === Object) {
+        return _.has(cache, address);
+      } else {
+        return (typeof cache[address] !== 'undefined');
+      }
+    }
+
     var memoize = function(key) {
       var cache = memoize.cache;
       var address = '' + (hasher ? hasher.apply(this, arguments) : key);
-      if (typeof cache[address] === 'undefined') cache[address] = func.apply(this, arguments);
+      if (!hasKey(cache, address)) cache[address] = func.apply(this, arguments);
       return cache[address];
     };
     memoize.cache = getCache();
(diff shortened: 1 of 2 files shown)


caiogondim:
Here the perf results:
screen shot 2016-03-07 at 21 07 26

Here the code with the perf benchmark:
https://gist.github.com/caiogondim/5e8334f7a70974e37590

underscore.js
```diff
@@ -779,13 +779,20 @@
...
+        return Object.create(null);
+      }
+      return {};
+    }
+
     var memoize = function(key) {
       var cache = memoize.cache;
       var address = '' + (hasher ? hasher.apply(this, arguments) : key);
-      if (!_.has(cache, address)) cache[address] = func.apply(this, arguments);
+      if (typeof cache[address] === 'undefined') cache[address] = func.apply(this, arguments);
```
jdalton: Wouldn't this need to be forked a bit more for enviros without `Object.create` which still need the own property checks.

akre54:
This is the kind of thing a generic `_.Map`would be good for.

jdalton:
>  This is the kind of thing a generic `_.Set` would be good for.

Yes! Lodash does this with `_.memoize.Cache` which is a map-like implementation `MapCache`. It can be swapped out with `WeakMap` or `Immutable.Map`. Internally there's also a `Hash`, `Stack` and `SetCache`.

jdalton:
Related to #1862.

akre54:
Sure, perhaps it's time to add it.

caiogondim:
I totally agree a generic `Set` would be better. But that comes with a bigger refactor.

@jdalton Already updated the code to properly check for keys in objects with and without prototype.

caiogondim:
With my last change, still getting something around 3x more performance.

underscore.js
```diff
@@ -779,13 +779,30 @@
...
+      return {};
+    }
+
+    function hasKey(cache, address) {
+      // Objects created with `Object.create(null)` don't have a constructor
+      // property.
+      if (cache.constructor === Object) {
+        return _.has(cache, address);
+      } else {
+        return (typeof cache[address] !== 'undefined');
```
jridgewell: The issue with this is `undefined` is a valid return value that should be memoized. We'll need to use `_.has` in every case.

caiogondim: Makes sense!
Should I add a test for this case?
Right now everything is passing.

jridgewell: Please. :smile:

underscore.js
```diff
@@ -779,13 +779,30 @@
...
+      if (Object.create) {
+        return Object.create(null);
+      }
+      return {};
+    }
+
+    function hasKey(cache, address) {
+      // Objects created with `Object.create(null)` don't have a constructor
+      // property.
+      if (cache.constructor === Object) {
```
jdalton: What if the value `constructor` is cached. That would throw off this check.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,74 @@
+# Contributor Covenant Code of Conduct
```
jdalton: You'll need to rebase this commit out.

caiogondim:
Just added the test case to `return undefined`.
It was indeed not being memoized.

caiogondim:
Did a rebase as well.
Should be good to go.

underscore.js
```diff
@@ -779,13 +779,21 @@
...
+      }
+      return {};
+    }
+
     var memoize = function(key) {
       var cache = memoize.cache;
       var address = '' + (hasher ? hasher.apply(this, arguments) : key);
-      if (!_.has(cache, address)) cache[address] = func.apply(this, arguments);
+      if (!(_.has(cache, address))) cache[address] = func.apply(this, arguments);
       return cache[address];
```
jdalton: Are you still seeing a perf win?

CODE_OF_CONDUCT.md
jdalton: You'll need to do:
```bash
git pull -f https://github.com/jashkenas/underscore
git reset --hard fed2310284ba061c2497e37d447d1d25b2a5af47
git cherry-pick 38ba9104027a8238a6fc8995c47974f85581a71b
git cherry-pick 4f173493f0bf0f26875b47e46b33980788873c10
git cherry-pick 4d518978334b034b83bf7f315deb3ba14b7fe94d
git push origin +master
```

underscore.js
jdalton: If you aren't don't close the PR. I'll walk you through some things to improve it.

caiogondim: No perf win using `_.has` =/
Can you spot something we could improve over here?

CODE_OF_CONDUCT.md
caiogondim: Done!

caiogondim:
Just resolved the conflict.

underscore.js
jdalton: Hey @caiogondim! I created a [screencast](https://www.dropbox.com/s/ykq7cnormkkbc9f/MapCache.mp4?dl=0) to dig into `MapCache` a bit more. You can follow along at [4.6.0-npm-packages/lodash._mapcache/index.js](https://github.com/lodash/lodash/blob/4.6.0-npm-packages/lodash._mapcache/index.js).

jdalton: Note: The `Hash` and `HASH_UNDEFINED` bits is something I learned from Ember.

jgonggrijp:
Closing this as it seems the performance gain has been lost in the process of making this work correctly, and as a performance enhancement based on #1862 would seem preferable.

PR #2464, opened 2016-03-09
Implements isMap, isSet, isWeakMap and isWeakSet #2449

### LINKED ISSUE #2449: isMap, isSet, isWeakSet, isWeakMap
Additional ES6 type methods for isMap, isSet, isWeakSet, isWeakMap as companions to `_.isSymbol`.

gwpmad:
Closes #2449 
This pull request adds the four functions in the title to the library. Relevant tests have been created in the test/objects.js file and all pass (and meet stylistic requirements).

### DIFF
--- a/test/objects.js
+++ b/test/objects.js
@@ -681,6 +681,22 @@
     assert.ok(_.isBoolean(false), 'and so is false');
   });
 
+  QUnit.test('isMap', function(assert) {
+    var keyString = 'a string', keyObj = {}, keyFunc = function() {};
+    var obj = new Map([[keyString, 'value one'], [keyObj, 'value two'],
+      [keyFunc, 'value three']]);
+
+    assert.ok(!_.isMap('string'), 'a string is not a map');
+    assert.ok(!_.isMap(2), 'a number is not a map');
+    assert.ok(!_.isMap({}), 'an object is not a map');
+    assert.ok(!_.isMap(false), 'a boolean is not a map');
+    assert.ok(!_.isMap(void 0), 'undefined is not a map');
+    assert.ok(!_.isMap([1, 2, 3]), 'an array is not a map');
+    assert.ok(!_.isMap(new Set([['key', 'value']])), 'a set is not a map');
+    assert.ok(!_.isMap(new WeakMap([[{x: 1}, 'value']])), 'a weak map is not a map');
+    assert.ok(_.isMap(obj), 'but a map is');
+  });
+
   QUnit.test('isFunction', function(assert) {
     assert.ok(!_.isFunction(void 0), 'undefined vars are not functions');
     assert.ok(!_.isFunction([1, 2, 3]), 'arrays are not functions');
--- a/test/objects.js
+++ b/test/objects.js
@@ -682,53 +682,76 @@
   });
 
   QUnit.test('isMap', function(assert) {
-    var keyString = 'a string', keyObj = {}, keyFunc = function() {};
-    var obj = new Map([[keyString, 'value one'], [keyObj, 'value two'],
-      [keyFunc, 'value three']]);
-
     assert.ok(!_.isMap('string'), 'a string is not a map');
     assert.ok(!_.isMap(2), 'a number is not a map');
     assert.ok(!_.isMap({}), 'an object is not a map');
     assert.ok(!_.isMap(false), 'a boolean is not a map');
     assert.ok(!_.isMap(void 0), 'undefined is not a map');
     assert.ok(!_.isMap([1, 2, 3]), 'an array is not a map');
-    assert.ok(!_.isMap(new Set([['key', 'value']])), 'a set is not a map');
-    assert.ok(!_.isMap(new WeakMap([[{x: 1}, 'value']])), 'a weakmap is not a map');
-    assert.ok(_.isMap(obj), 'but a map is');
+    if (typeof Set === 'function') {
+      assert.ok(!_.isMap(new Set()), 'a set is not a map');
+    }
+    if (typeof WeakSet === 'function') {
+      assert.ok(!_.isMap(new WeakSet()), 'a weakset is not a map');
+    }
+    if (typeof WeakMap === 'function') {
+      assert.ok(!_.isMap(new WeakMap()), 'a weakmap is not a map');
+    }
+    if (typeof Map === 'function') {
+      var keyString = 'a string';
+      var obj = new Map();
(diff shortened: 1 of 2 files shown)


test/objects.js
```diff
@@ -681,6 +681,67 @@
     assert.ok(_.isBoolean(false), 'and so is false');
   });
 
+  QUnit.test('isMap', function(assert) {
+    var keyString = 'a string', keyObj = {}, keyFunc = function() {};
+    var obj = new Map([[keyString, 'value one'], [keyObj, 'value two'],
```
jdalton: You need to ensure `Map`, `Set`, `WeakMap` and `WeakSet` exist. The tests will fail in browsers like IE10 and environments like PhantomJS or Node 0.10.0

test/objects.js
```diff
@@ -681,6 +681,67 @@
...
+      [keyFunc, 'value three']]);
+
```
jdalton: You should also avoid populating the `Map` with the constructor+array syntax as that's not supported in IE11.
Use `map.set(key, value)` instead.

gwpmad: Thanks, appreciate the feedback, will get on with addressing these now.

test/objects.js
```diff
@@ -681,6 +681,67 @@
...
+    var keyObj = {};
+    var obj = new WeakMap([[keyObj, 'value']]);
+
+    assert.ok(!_.isWeakMap('string'), 'a string is not a weakmap');
+    assert.ok(!_.isWeakMap(2), 'a number is not a weakmap');
+    assert.ok(!_.isWeakMap({}), 'an object is not a weakmap');
+    assert.ok(!_.isWeakMap(false), 'a boolean is not a weakmap');
+    assert.ok(!_.isWeakMap(void 0), 'undefined is not a weakmap');
+    assert.ok(!_.isWeakMap([1, 2, 3]), 'an array is not a weakmap');
+    assert.ok(!_.isWeakMap(new Set([['key', 'value']])), 'a set is not a weakmap');
```
jdalton: The [same goes](https://github.com/jashkenas/underscore/pull/2464/files#r55551957) for the `Set` creation.

test/objects.js
jdalton: No prob. There's a bit more it can come after the other things are addressed.
You can use the travis checks indicator to know when issues with tests have been resolved.
image

gwpmad:
@jdalton Have updated code to reflect requested changes.

akre54:
Thanks :+1:

jdalton:
Unit tests still failing.

gwpmad:
@jdalton Travis seemed to be passing when I made the second pull request - could you let us know where the issue is?

jdalton:
The [failing travis tests](https://travis-ci.org/jashkenas/underscore/builds/114860571) kicked in on the merge for the browser runs (browser runs are skipped for PRs).

MahmudH:
@jdalton we've removed the calls to .add(), which some versions of FireFox didn't like. 

However, we are trying to understand why IE 11 is returning false for isMap and isWeakMap. We are wondering if it is how IE 11 treats toString.call(). We are a bit stuck because there are no issues with other types such as String, Number, Date etc.

We were thinking of doing Object.prototype.toString.call() instead of toString.call() after reading http://webreflection.blogspot.co.uk/2007/07/quick-fix-internet-explorer-and.html.

Please could you give us your thoughts?

Thank you.

jdalton:
@MahmudH
Internally Underscore has `toString = ObjProto.toString` so no need to change that.
While IE 11 supports `Map`, `Set`, and `WeakMap` it didn't have a `ToStringTag` for them so they report `[object Object]` when coerced using `Object.prototype.toString.call(...)`.

Here is how I [work around it in Lodash](https://github.com/lodash/lodash/blob/4.6.1-npm/_getTag.js)

* I [grab a reference to their constructors coerced to strings](https://github.com/lodash/lodash/blob/4.6.1-npm/_getTag.js#L23-L26) via `Function.prototype.toString`
* Then I [test for the buggy environment](https://github.com/lodash/lodash/blob/4.6.1-npm/_getTag.js#L40-L42) and fork the `getTag` helper
* Then if the result is `[object Object]` I [check for the constructor](https://github.com/lodash/lodash/blob/4.6.1-npm/_getTag.js#L44-L46) and coerce it to a string
* If the result isn't empty [I compare](https://github.com/lodash/lodash/blob/4.6.1-npm/_getTag.js#L49-L53) to the `Map`, `Set`, and `WeakMap` references

jdalton:
The reason the `add` calls were failing in versions of Chrome/Firefox/IE/etc is because in early implementations the `add()` and `set()` methods didn't return the set or map instance (so no chaining `set.add(1).add('string')`)

akre54:
@jdalton if you were aware this extra `getTag` helper was needed when you opened #2449, it would help us if you made note of these kinds of requirements next time you open an issue.

jdalton:
I [mentioned there was more work to do](https://github.com/jashkenas/underscore/pull/2464#discussion_r55554040). At the time of #2449 I didn't have a PR to add feedback to.

MahmudH:
@jdalton should we basically do the same as the lodash _getTag.js in the underscore.js file

jdalton:
> should we basically do the same as the lodash _getTag.js in the underscore.js file

You could or you could see if there's a better way. Since this PR is merged the fixes should probably come in a new PR.

akre54:
> I mentioned there was more work to do.

Specifically I'm referring to the original issue #2449. Given that this was more complex than simply adding options to the `_.is` helpers, it would have been good if you made note of this.

jdalton:
> Specifically I'm referring to the original issue #2449. Given that this was more complex than simply adding options to the _.is helpers, it would have been good if you made note of this.

I'm the lead of a competing library with a touch and go relationship with some in this core and yet I've opened ~12x more Underscore issues than any other user. API parity, when possible, is a good thing for users. My goal isn't to do all the work for Lodash (adding the feature, testing the feature, research support, and creating workarounds) and then hand it to Underscore. However, I'll open the issue and help when things are blocked. Many times you all come up with completely different and better solutions without having it given to you. 

This issue was great, esp. for students, because it was relatively easy to implement on the surface, but then required digging into a non-obvious test fail, testing multiple environments, and coming up with a work around.

gwpmad:
New pull request now made to address the add() problem and the IE11 bug - #2466

jdalton:
Thank you @gwpmad!

PR #2465, opened 2016-03-10
Added missing isError ToC entry

Khez:


captbaritone:
This change was already in `master` in this PR (https://github.com/jashkenas/underscore/pull/2269), but hadn't been cherry-picked over to `gh-pages`.

The two branches are now slightly out of sync since the ordering is slightly different. Not a problem, but I figured I'd mention it here in case anyone else became confused.

PR #2466, opened 2016-03-10
isMap, isSet and isWeakMap functions implemented, which Fix IE11 and Firefox tests

### LINKED ISSUE #2449: isMap, isSet, isWeakSet, isWeakMap

MahmudH:
This is following our conversation from #2464 to close #2449.

### DIFF
--- a/test/objects.js
+++ b/test/objects.js
@@ -746,7 +746,6 @@
     }
     if (typeof Set === 'function') {
       var obj = new Set();
-      obj.add(1).add('string').add(false).add({});
       assert.ok(_.isSet(obj), 'but a set is');
     }
   });
@@ -770,7 +769,6 @@
     }
     if (typeof WeakSet === 'function') {
       var obj = new WeakSet();
-      obj.add({x: 1}, {y: 'string'}).add({y: 'string'}).add({z: [1, 2, 3]});
       assert.ok(_.isWeakSet(obj), 'but a weakset is');
     }
   });
--- a/underscore.js
+++ b/underscore.js
@@ -1288,25 +1288,39 @@
     return type === 'function' || type === 'object' && !!obj;
   };
 
-  // Add some isType methods: isArguments, isFunction, isString, isNumber, isDate, isRegExp, isError, isWeakSet.
-  _.each(['Arguments', 'Function', 'String', 'Number', 'Date', 'RegExp', 'Error', 'Symbol', 'WeakSet'], function(name) {
+
+  /**
+  Helper function for returning the object tag (e.g. [object Object]); implemented
+  to work around IE11 bug (Object.Prototype.toString with Map, Set and WeakMap
+  doesn't return '[object Map]' etc.)
+  **/
+  var returnTag = function(obj) {
+    var objectTag = Object.prototype.toString.call(obj),
+        objCtor = objectTag === '[object Object]' ? obj.constructor : null,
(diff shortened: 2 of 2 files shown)


underscore.js
```diff
@@ -1288,13 +1288,26 @@
...
 
+  // Work around IE11 bug (Object.Prototype.toString with Map, Set and WeakMap doesn't return '[object Map]' etc.)
+  // Add isType methods for isMap, isWeakMap and isSet.
+  _.each(['Map', 'Set', 'WeakMap'], function(name) {
+    _['is' + name] = function(obj) {
+      if (!obj) {
+        return false;
+      }
+      var objCtor = obj.constructor,
+          objCtorString = Function.prototype.toString.call(objCtor);
```
akre54: In Chrome I get `Uncaught TypeError: Function.prototype.toString is not generic`

underscore.js
```diff
@@ -1288,13 +1288,26 @@
...
+      return objCtorString === 'function ' + name + '() { [native code] }';
```
jdalton: While the string value of the constructor aligns with this it's not guaranteed to be this format which is why I used a reference to the constructors instead of manually generating it.

underscore.js
jdalton: If `objCtor` is not a function it will throw an error when coerced by `Function.prototype.toString.call(...)`.

jdalton: Would make a good unit test.

gwpmad:
Thanks, I'm going to try to fix this. Going to put a helper function in just below it, hopefully that's ok!

gwpmad:
Hi, I've updated the pull request to hopefully address the issues @jdalton mentions. I've basically added a returnTag helper function that brings in some of the functionality from the lodash _getTag.js file. There are a number of ternary statements in there so that older browsers don't get exposed to Maps/Sets/WeakMaps resulting in Travis fails.

MahmudH:
HI @jdalton. Any updates on this? Thanks

test/objects.js
```diff
@@ -746,7 +746,6 @@
     }
     if (typeof Set === 'function') {
       var obj = new Set();
-      obj.add(1).add('string').add(false).add({});
```
jgonggrijp, line 749: If you test code is essentially correct, but it breaks in some environments due to lack of support, then just deleting that test code is never the right solution. In this case it could have been fixed by just not chaining the calls but repeating `obj.add` four times on a separate line.

underscore.js
```diff
@@ -1288,13 +1288,42 @@
...
+          setCtorString = typeof Set === 'function' ? Function.prototype.toString.call(Set) : '';
+
+      switch (objCtorString) {
+        case mapCtorString : return '[object Map]';
+        case weakMapCtorString : return '[object WeakMap]';
+        case setCtorString : return '[object Set]';
+      }
+    }
+    return objectTag;
+  };
```
jgonggrijp, line 1317: This is a very large workaround just to fix three `isType` functions for one version of one browser. I don't like the size of it and I don't like the fact that it is being applied to all the other `isType` functions as well.

jgonggrijp:
This is fix for a real bug but the fix needs to be much more elegant.

jgonggrijp:
Superseded by #2884.

Issue #2457, opened 2016-03-04
_.omit and symbols


jdalton:
ES6 Symbols work out of the box in many Underscore methods like `_.has` and `_.pick`, however `_.omit` needs special attention for support.

``` js
function Foo() {
  this.a = 0;
  this[symbol1] = 1;
}
var symbol1 = Symbol('a');
var symbol2 = Symbol('b');
Foo.prototype[symbol2] = 2;

var foo = new Foo;
var actual = _.omit(foo, 'a');

actual[symbol1] // should be 1 but is `undefined`
actual[symbol2] // should be 2 but is `undefined`
```

edwardkerry:
I'm taking a look a this, hope to submit a PR tomorrow.

jdalton:
Rock! Thank you!

jdalton:
Besides omitting correct symbols it should also not coerce them to strings:
```js
var o = { 'Symbol(a)': 'boo', 'Symbol(b)': 'baz' };
_.omit(o, Symbol('a'));
// => { 'Symbol(b)': 'baz' }
```

harshitharao:
@jdalton @edwardkerry https://github.com/jashkenas/underscore/pull/2789 is the fix PR. Can you check once

jgonggrijp:
Closing for the same reason as #2789.

PR #2468, opened 2016-03-11
WIP adding symbols to _.omit

edwardkerry:
NOT COMPLETED

This is a work in progress, attempting to address
https://github.com/jashkenas/underscore/issues/2457

Any guidance on how best to approach this would be welcome. I have not been successful in passing the objects symbols through the function.


test/objects.js
```diff
@@ -243,6 +243,15 @@
     assert.deepEqual(_.omit(data, function(val, key) {
       return this[key] === 3 && this === instance;
     }, instance), {a: 1, b: 2}, 'function is given context');
+
+    var Foo = function(){
+      this.a = 0;
+      this[symbol1] = 1;
+    }
+    var symbol1 = Symbol('sym1');
+    var foo = new Foo();
```
jdalton: Be sure to add
```js
var symbol2 = Symbol('sym2');
Foo.prototype[symbol2] = 2;
```
before
```js
var var foo = new Foo();
```

underscore.js
```diff
@@ -1100,6 +1100,9 @@
 
    // Return a copy of the object without the blacklisted properties.
   _.omit = restArgs(function(obj, keys) {
+    if(obj != null && Object.getOwnPropertySymbols(obj) != null){
```
akre54: `Object.getOwnPropertySymbols` is a relatively recent addition. IE doesn't support it yet, for instance. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertySymbols#Browser_compatibility

jdalton: IE won't as there's no more IE after 11. Edge 12 and 13 do though.

akre54: Regardless, FF and Chrome both added support only about a year ago, which is typically less than our support window. I think it's less of an issue in evergreen browsers (Edge is evergreen, right?) but probably something we should figure out in another thread.

Regardless, @edwardkerry maybe add a check that `Object.ownPropertySymbols` exists before using it?

jdalton: > Regardless, FF and Chrome both added support only about a year ago, which is typically less than our support window.

The browsers with symbol support have the method so there's not an issue of inconsistent support. If they support symbols then it'll work. Yes they'll need to be method existence checks.

jdalton: > (Edge is evergreen, right?) 

Yep!

PR #2473, opened 2016-03-21
Support to Set and Map with '_.size'

JonDotsoy:
This allows '_.size' for 'Map' and 'Set'.

Example:
```Javascript
let e = new Set();

e.add(1);
e.add(13);
e.add(12);

_.size(e);
```

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -19,9 +19,6 @@
   // Save bytes in the minified (but not gzipped) version:
   var ArrayProto = Array.prototype, ObjProto = Object.prototype;
 
-  // There is support for using 'Set' and 'Map'.
-  var supportToSetAndMap = (typeof Set === "function" && typeof Map === "function" && Set.name == "Set" && Map.name == "Map");
-
   // Create quick reference variables for speed access to core prototypes.
   var
     push = ArrayProto.push,
--- a/underscore.js
+++ b/underscore.js
@@ -441,7 +441,7 @@
   // Return the number of elements in an object.
   _.size = function(obj) {
     if (obj == null) return 0;
-    return (property("size")(obj)) ? obj.size : isArrayLike(obj) ? obj.length : _.keys(obj).length;
+    return (_.isMap(obj) || _.isSet(obj)) ? obj.size : isArrayLike(obj) ? obj.length : _.keys(obj).length;
   };
 
   // Split a collection into two arrays: one whose elements all satisfy the given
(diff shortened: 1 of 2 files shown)


underscore.js
```diff
@@ -441,6 +441,7 @@
   // Return the number of elements in an object.
   _.size = function(obj) {
     if (obj == null) return 0;
+    if (obj instanceof Set || obj instanceof Map) return obj.size;
```
jdalton: `instanceof` checks will fail on maps/sets from other realms.

bjmiller:
Should we be getting emails from coveralls, or other CI services?  Is there a way to restrict (or, perhaps opt out of) automatic CI emails, while still getting the rest?

underscore.js
```diff
@@ -19,6 +19,9 @@
   // Save bytes in the minified (but not gzipped) version:
   var ArrayProto = Array.prototype, ObjProto = Object.prototype;
 
+  // There is support for using 'Set' and 'Map'.
+  var supportToSetAndMap = (typeof Set === "function" && typeof Map === "function" && Set.name == "Set" && Map.name == "Map");
+
```
jdalton: IE11 doesn't have `Function#name`.

underscore.js
```diff
@@ -441,6 +444,7 @@
   // Return the number of elements in an object.
   _.size = function(obj) {
     if (obj == null) return 0;
+    if (supportToSetAndMap && (obj instanceof Set || obj instanceof Map)) return obj.size;
```
jdalton: This is still using `instanceof` which will fail sets and maps from other realms (think iframe / `vm` module in Node).

underscore.js
```diff
@@ -441,7 +441,7 @@
   // Return the number of elements in an object.
   _.size = function(obj) {
     if (obj == null) return 0;
-    return isArrayLike(obj) ? obj.length : _.keys(obj).length;
+    return (property("size")(obj)) ? obj.size : isArrayLike(obj) ? obj.length : _.keys(obj).length;
```
jdalton: This is duck-typing which isn't great.
The master branch version of Underscore has `_.isMap` and `_.isSet`.

akre54:
Yeah where did the coveralls coverage come in? @jridgewell @megawac @michaelficarra did any of yall turn it on?

jridgewell:
Not me.

JonDotsoy:
Move :arrow_right: #2476

Issue #2474, opened 2016-03-22
Is there a reason that the callback function doesn't handle regexes?


olleicua:
It would be great to have:

```javascript
_.select(['abc', 'def'], /ab/) // [ 'abc' ]
```

And it seems like an easy addition:

```javascript
var cb = function(value, context, argCount) {
  if (value == null) return _.identity;
  if (_.isFunction(value)) return optimizeCb(value, context, argCount);
  if (_.isObject(value)) return _.matcher(value);
  if (_.isRegExp(value)) return function(obj) { return !!value.exec(obj) };
  return _.property(value);
};
```

akre54:
Sure, looks great. In the future, please just open one or the other an issue or pull request, no need to open two of them.

jdalton:
Seems like this would be a good reason to use `_.iteratee` to be used internally instead.
That way if folks want to customize shorthands they can just overload `_.iteratee`.

captbaritone:
Closing in favor of the related PR: #2475

FYI: We are discussing allowing users to customize _.iteratee here: https://github.com/jashkenas/underscore/pull/2480

PR #2475, opened 2016-03-22
Issue #2474 allow callback function to handle regexes

olleicua:
https://github.com/jashkenas/underscore/issues/2474

### DIFF
--- a/test/collections.js
+++ b/test/collections.js
@@ -317,6 +317,10 @@
     assert.deepEqual(_.filter(list, {b: 2}), [{a: 1, b: 2}, {a: 2, b: 2}]);
     assert.deepEqual(_.filter(list, {}), list, 'Empty object accepts all items');
     assert.deepEqual(_(list).filter({}), list, 'OO-filter');
+
+    // Works with RegExps
+    var stringArray = ['cat', 'dog', 'catdog'];
+    assert.deepEqual(_.filter(stringArray, /cat/), ['cat', 'catdog']);
   });
 
   QUnit.test('select', function(assert) {
--- a/underscore.js
+++ b/underscore.js
@@ -90,6 +90,7 @@
   var cb = function(value, context, argCount) {
     if (value == null) return _.identity;
     if (_.isFunction(value)) return optimizeCb(value, context, argCount);
+    if (_.isRegExp(value)) return function(obj) { return !!value.exec(obj) };
     if (_.isObject(value)) return _.matcher(value);
     return _.property(value);
   };


akre54:
:thumbsup:

test/collections.js
```diff
@@ -317,6 +317,10 @@
     assert.deepEqual(_.filter(list, {b: 2}), [{a: 1, b: 2}, {a: 2, b: 2}]);
     assert.deepEqual(_.filter(list, {}), list, 'Empty object accepts all items');
     assert.deepEqual(_(list).filter({}), list, 'OO-filter');
+
+    // Works with RegExps
+    var stringArray = ['cat', 'dog', 'catdog'];
+    assert.deepEqual(_.filter(stringArray, /cat/), ['cat', 'catdog']);
```
akre54: Should this method just error on a non-string array?

underscore.js
```diff
@@ -90,6 +90,7 @@
   var cb = function(value, context, argCount) {
     if (value == null) return _.identity;
     if (_.isFunction(value)) return optimizeCb(value, context, argCount);
+    if (_.isRegExp(value)) return function(obj) { return !!value.exec(obj) };
```
jridgewell: Would `#test` work?

jdalton:
:-1: I'd rather `_.iteratee` was overloadable to allow customization without expanding core support (which includes adding another `_.property` or `_.matcher` type method for this case too).

jdalton:
Related to https://github.com/lodash/lodash/issues/971#issuecomment-74441735.

jdalton:
Related to #1965.

test/collections.js
olleicua: In JavaScript arrays can have mixed type.  I'm guessing we want:

```javascript
_.filter([123, "123", {a: 1}, []], /23/); // [123, "123"]
```

Do you agree?

underscore.js
olleicua: @jridgewell I'm not sure I understand the question

olleicua:
@jdalton I'm not sure I understand your concerns here.  The issue in #1965 was with additional layers of function indirection and my solution only adds a single if statement..  am I missing something?

jdalton:
@olleicua Everyone wants just their one thing added. But it has a cost. As I mentioned they'll need to be companion exposed versions of the functionality too (just as with `_.property` and friends). Allowing `_.iteratee` gives Underscore a way to support your use case without having to expand its support or API.

jdalton:
Your use case can be tackled with `_.bind` or the `context` param too.

jdalton:
Allowing `_.iteratee` to be overloaded allows for things beyond iteratee shorthands, like [FP](https://github.com/lodash/lodash/wiki/FP-Guide) wrappers, too.

underscore.js
captbaritone: @olleicua I believe he's asking if we could use this method instead of `.exec()`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test

test/collections.js
captbaritone: It would probably be best to test `_.iteratee` as it's a more direct way of testing `cb()`: https://github.com/jashkenas/underscore/blob/67e5c9eb585124d7a6dda2c1012e4ac240678f6e/test/functions.js#L678-L691

captbaritone:
I agree that overloading _.iteratee/cb() further comes with a cost, and that part of a libraries duty is to resist including everyone's pet functionality, especially for something as magical as `_.iteratee`.

That said, I'm :+1: To this idea in general. The syntax is unambiguous and readable and I can't see any other reasonable interpretation of a RegEx being passed as an iteratee/predicate.

jdalton:
Another alternative is to introduce something like [bindKey](https://github.com/lodash/lodash/blob/4.6.1-npm/bindKey.js#L32-L48) which is like `_.bind` except that it allows bound functions to reference
methods [that may be redefined or don't yet exist](http://peter.michaux.ca/articles/lazy-function-definition-pattern). This would allow both `test` and `exec` forms as:
```js
_.bindKey(/cat/, 'exec')
_.bindKey(/dog/, 'test')
```

olleicua:
I'm not sure what the process is and I'm sensing a lack of consensus in this thread.  Should I make the suggested changes or will this not be merged?

captbaritone:
@olleicua Feel free to wait until we reach a consensus to make the changes. Thanks for your patience!

captbaritone:
I've thought about this a bit more, and I think there is a case where the expected behavior of a regex predicate could be ambiguous. 

In the case of a non-predicate iteratee, it might be reasonable to expect the regex function to return the portion of the string that matched.

```
var urls = ['https://google.com', 'http://example.com'];
var protocols = _.map(urls, /^https?/);

// Might expect
=> ['https', 'http']

// Rather than the proposed
=> [true, true]
```

This additional complexity makes me less enthusiastic about including this syntax.

akre54:
I'm not sure I agree. We could use `exec` or `match` in the callback instead of `test`, which would still work for `filter`. 

```js
> urls.map(d => /^https?/.exec(d))
[ [ 'https', index: 0, input: 'https://google.com' ],
  [ 'http', index: 0, input: 'http://example.com' ] ]
> urls.map(d => /^https/.exec(d))
[ [ 'https', index: 0, input: 'https://google.com' ], null ]
> urls.filter(d => /^https/.exec(d))
[ 'https://google.com' ]
```

captbaritone:
@akre54 Do we care about the falsey `""` string edge case?

akre54:
Blerg. Good point.

akre54:
Wait just kidding (I saw you updated your comment too). That's exactly what I'm suggesting. 

```
> /.*/.exec('')
[ '', index: 0, input: '' ]
```

Still truthy.

Any other edge cases?

jdalton:
What about global matches?
Where will this added functionality live on `_`?
What about guards for the implicit string coercing like arrays containing symbols?
Is the Underscore diet still a concern or can that be put to bed?

captbaritone:
With #2480 merged, and the edge cases mentioned by @jdalton I'm tempted to close this in favor of allowing users to customize `_.iteratee`.

akre54:
Sounds good to me

On Tue, Mar 29, 2016, 12:55 Jordan Eldredge <notifications@github.com>
wrote:

> With #2480 <https://github.com/jashkenas/underscore/pull/2480> merged,
> and the edge cases mentioned by @jdalton <https://github.com/jdalton> I'm
> tempted to close this in favor of allowing users to customize _.iteratee.
>
> —
> <https://github.com/jashkenas/underscore/pull/2475#issuecomment-202996647>
>
-- 

Adam K (mobile)

PR #2476, opened 2016-03-22
Integrates support for such securities "Set" and "Map"

JonDotsoy:


underscore.js
```diff
@@ -449,8 +449,18 @@
...
   // Return the number of elements in an object.
   _.size = function(obj) {
+    var typeObj = toString.call(obj);
     if (obj == null) return 0;
-    return isArrayLike(obj) ? obj.length : _.keys(obj).length;
+    if (isArrayLike(obj)) {
+      return obj.length;
+    } else {
+      switch (typeObj) {
+        case '[object Set]':
```
jdalton: This won't do because IE11 reports them as `[object Object]`, something that `_.isMap` and `_.isSet` can be patched to avoid (out of scope for this PR).

JonDotsoy: This use `toString.call(obj)` to get the type.
It is more optimal than using "_.isMap" or "_.isSet".

jdalton: Yeah, but `_.isMap` and `_.isSet` are fixable for the bug in `toString`.

test/collections.js
```diff
@@ -826,6 +826,8 @@
     assert.equal(_.size({one: 1, two: 2, three: 3}), 3, 'can compute the size of an object');
     assert.equal(_.size([1, 2, 3]), 3, 'can compute the size of an array');
     assert.equal(_.size({length: 3, 0: 0, 1: 0, 2: 0}), 3, 'can compute the size of Array-likes');
+    assert.equal(_.size(new Set([1, 22, 55, 12])), 4, 'can compute the size of an Set');
```
jdalton: IE11 doesn't support map/set initialization through arrays in the constructor so you'll have to create them like
```js
var set = new Set;
set.add(1)
set.add(22);

var map = new Map;
map.set(1, 5);
map.set(3, 4);
```

megawac: Neither do several older iterations of Chrome

PR #2479, opened 2016-03-23
bug fixed

### LINKED ISSUE #2478: there may be some problems with  _.debounce
in this file <https://github.com/jashkenas/underscore/blob/master/underscore.js>

I think there may be some issues in _.debounce function, if we use like this

```
var lazyLayout = _.debounce(function() {
	console.log(123)
	} , 300, true);

window.onresize = lazyLayout;
```

there will be only one '123' in the console after resizing the window several times.

hanzichi:
#2478

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -856,7 +856,7 @@
 
     var later = function(context, args) {
       timeout = null;
-      if (args) result = func.apply(context, args);
+      if (!immediate) result = func.apply(context, args);
     };
 
     var debounced = restArgs(function(args) {
@@ -865,7 +865,7 @@
       if (callNow) {
         timeout = setTimeout(later, wait);
         result = func.apply(this, args);
-      } else if (!immediate) {
+      } else {
         timeout = _.delay(later, wait, this, args);
       }
 


underscore.js
```diff
@@ -865,7 +865,7 @@
       if (callNow) {
         timeout = setTimeout(later, wait);
         result = func.apply(this, args);
-      } else if (!immediate) {
```
jridgewell: This will schedule a new timeout every time the function is called when `immediate = true`.

jridgewell:
After review, this is definitely incorrect behavior. This will call `func` after `wait` every time the debounced function is called when `immediate = true`.

hanzichi:
@jridgewell  thanks for your reply!  I just wonder with the code below, I just find the different results when resizing the window for several times using [underscore.js](https://github.com/jashkenas/underscore/blob/master/underscore.js) and [underscore-min.js](https://github.com/jashkenas/underscore/blob/master/underscore-min.js)

```
  var lazyLayout = _.debounce(function() {
    	console.log(123)
    	} , 300, true);

    window.onresize = lazyLayout;
```




don't you? hope to hear from you soon ! thanks !

megawac:
@hanzichi are you mixing up `debounce` with `throttle`?

hanzichi:
@megawac no, i am not, could you pls have a try as i said above ?

megawac:
I did and I got the results I expected

hanzichi:
i went through the code, and found line 866

`    timeout = setTimeout(later, wait);`

but we will never trigger `later

jridgewell:
> i went through the code, and found line 866
> timeout = setTimeout(later, wait);

That's `_.throttle`'s code.

hanzichi:
I do not think so...

jridgewell:
I don't think this is the right solution, but there's definitely something wrong. I'll spend some more time on this tomorrow.

hanzichi:
@jridgewell thanks for your work !

jridgewell:
Closing this for #2482. Thanks for the bug report!

hanzichi:
that's what i should do, thanks for your quick fix again !

PR #2480, opened 2016-03-23
Allow users to customize _.iteratee

captbaritone:
This fix was originally proposed by @jridgewell in #1965 but it was unable to gain consensus since his approach added an additional layer of abstraction to which @akre54 objected. I've refactored @jridgewell's approach to avoid that abstraction.

This PR was motivated by a renewed interest in customizable shorthand syntaxes such as the RegEx syntax proposed in #2475.

If this approach looks good to people, I can look into more robusts tests and documentation.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -84,18 +84,23 @@
     };
   };
 
+  var builtinIteratee;
+
   // An internal function to generate callbacks that can be applied to each
   // element in a collection, returning the desired result — either `identity`,
   // an arbitrary callback, a property matcher, or a property accessor.
   var cb = function(value, context, argCount) {
+    if (_.iteratee !== builtinIteratee) return _.iteratee(value, context);
     if (value == null) return _.identity;
     if (_.isFunction(value)) return optimizeCb(value, context, argCount);
     if (_.isObject(value)) return _.matcher(value);
     return _.property(value);
   };
 
-  // An external wrapper for the internal callback generator.
-  _.iteratee = function(value, context) {
+  // External wrapper for our callback generator. Users may customize
+  // `_.iteratee` if they want additional predicate/iteratee shorthand styles.
+  // This abstraction hides the internal-only argCount argument.
+  _.iteratee = builtinIteratee = function(value, context) {
     return cb(value, context, Infinity);
   };
 
@@ -443,7 +448,7 @@
       // Keep surrogate pair characters together
       return obj.match(reStrSymbol);
     }
-    if (isArrayLike(obj)) return _.map(obj);
+    if (isArrayLike(obj)) return _.map(obj, _.identity);
     return _.values(obj);
   };
 
@@ -495,7 +500,7 @@
 
   // Trim out all falsy values from an array.
   _.compact = function(array) {
-    return _.filter(array);
+    return _.filter(array, Boolean);
   };
 
   // Internal implementation of a recursive `flatten` function.
--- a/test/collections.js
+++ b/test/collections.js
@@ -165,6 +165,14 @@
     // Passing a property name like _.pluck.
     var people = [{name: 'moe', age: 30}, {name: 'curly', age: 50}];
     assert.deepEqual(_.map(people, 'name'), ['moe', 'curly'], 'predicate string map to object properties');
+
+    var iteratee = _.iteratee;
+    _.iteratee = function(n) { return function(value) { return value * n; }; };
+
+    assert.deepEqual(_.map([1, 2, 3], 2), [2, 4, 6], 'will use a user-defined _.iteratee function');
+
+    // Replace the built-in iteratee so as to not break other tests
+    _.iteratee = iteratee;
   });
 
   QUnit.test('collect', function(assert) {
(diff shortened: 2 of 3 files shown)


underscore.js
```diff
@@ -84,18 +84,23 @@
     };
   };
 
+  var cb, builtinIteratee;
```
captbaritone: Since both functions reference each other, I had do these definitions first avoid triggering ESLint's [no-use-before-define](http://eslint.org/docs/rules/no-use-before-define).

akre54: sigh :)

That's a dumb rule ;)

underscore.js
```diff
@@ -84,18 +84,23 @@
     };
   };
 
+  var builtinIteratee;
```
captbaritone: @akre54, @jridgewell I'd be open to leaving this as is, switching the ESLint rule to a warning, or ignoring the warning with an inline comment. Our previous conversation got destroyed with my force-push.

akre54: I vote for moving the declaration (like this) or changing the rule to a warning. We should also file a bug against ESLint

captbaritone:
FYI tests seem to be failing for an unrelated reason:

> This is most likely a problem with the phantomjs-prebuilt package, not with npm itself.

jridgewell:
Kicked the test build, it should pass. This LGTM.

captbaritone:
@jridgewell Tests now passing. Would you like to wait on improved tests and documentation, or merge now and trust that I'll followup with test/doc PR(s)?

jridgewell:
I'm going to leave that up to @akre54.

captbaritone:
Sounds good. It might be worth merging this on it's own, since the testing/documentation approaches may need their own, narrowly scoped, debate.

underscore.js
```diff
@@ -495,7 +500,7 @@
 
   // Trim out all falsy values from an array.
   _.compact = function(array) {
-    return _.filter(array);
+    return _.filter(array, _.identity);
```
megawac: I'd personally prefer `_.map(obj, Boolean)`

megawac:
:+1:

underscore.js
captbaritone: :+1: Updated.

captbaritone:
I did some work on testing last night, and uncovered something interesting. There are a number of places where we depend upon `cb()` returning a function if you pass it a function.

If the user were to define an `_.iteratee` that did __not__ property handle functions, it would break many Underscore methods.

I see two possible solutions to this problem: 

1. Document this requirement as part of the "you may customize `_.iteratee`" documentation.
2. Move the `_.isFunction` test to the top of `cb()`, before we consider deferring to the user's `_.iteratee`.

Thoughts?

Florian-R:
I'd go for 2, maybe with a test-case added. Should be better perf-wise, as I'm not sure customizing `_.iteratee` will be quite common.

jridgewell:
@captbaritone: Can you change all `optimizeCb` to `cb`? That way everything can be overridden with custom `_.iteratee`. Looks like it's just `_.each`, `_.reduce`/`reduceRight` and `_.pick`.

> There are a number of places where we depend upon cb() returning a function if you pass it a function.

Isn't that the contract for `_.iteratee`? It returns a function. I wouldn't worry about them doing something stupid with it.

captbaritone:
@jridgewell Can we consider expanding the number of methods that transform through `_.iteratee` in a separate issue/PR? There are a few subtleties to discuss which I think might bog us down.

I think this change can stand alone as an improvement.

captbaritone:
@akre54 I think we're waiting on your thoughts.

jridgewell:
Good with me.

akre54:
I think practically not a ton of people are going to use this, so I'm +0. That said, this implementation looks good. Can we add at least a few more baseline tests and then looks good to merge for me.

captbaritone:
@akre54 Added as a second commit with more robust tests. It sets up a custom `_.iteratee` and uses it with each of the methods that transform through `_.iteratee`.

akre54:
lgtm

underscore.js
megawac: So this doesn't work correct?

```js
var builtinIteratee = _.iteratee = function(...
```

akre54: It does. But eslint is being a pain. (See earlier discussion)

On Tue, Mar 29, 2016, 12:20 Graeme Yeates <notifications@github.com> wrote:

> In underscore.js
> <https://github.com/jashkenas/underscore/pull/2480#discussion_r57752919>:
>
> > @@ -84,18 +84,23 @@
> >      };
> >    };
> >
> > +  var builtinIteratee;
>
> So this doesn't work correct?
>
> var builtinIteratee = _.iteratee = function(...
>
> —
> <https://github.com/jashkenas/underscore/pull/2480/files/5efa51ebda047dc4aef8f3680ec31e42159e41d7#r57752919>
>
-- 

Adam K (mobile)

captbaritone:
:clap: Thanks everyone. I'll look at adding some documentation in the next few days.

PR #2482, opened 2016-03-24
Fix debounce immediate

### LINKED ISSUE #2478: there may be some problems with  _.debounce

jridgewell:
Fixes an issue with `_.debounce` when `{ immediate: true }`. When
called twice within `wait` ms, we cleared the `timeout`. Because it is
truthy on the second run, it was cleared but `timeout` remained
truthy (on any later run, `timeout` remains truthy). Because `timeout`
was cleared, the `later` function is never run to null out `timeout`.

Anyways, we should always be setting a fresh timeout function when
`{ immediate: true }`, to prevent further calls to the debounced
function until the `wait` ms after the last call.

Fixes #2478, supersedes #2479. Thanks for the bug report @hanzichi.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -860,12 +860,12 @@
     };
 
     var debounced = restArgs(function(args) {
-      var callNow = immediate && !timeout;
       if (timeout) clearTimeout(timeout);
-      if (callNow) {
+      if (immediate) {
+        var callNow = !timeout;
         timeout = setTimeout(later, wait);
-        result = func.apply(this, args);
-      } else if (!immediate) {
+        if (callNow) result = func.apply(this, args);
+      } else {
         timeout = _.delay(later, wait, this, args);
       }
 
--- a/test/functions.js
+++ b/test/functions.js
@@ -475,9 +475,9 @@
   });
 
   QUnit.test('debounce asap', function(assert) {
-    assert.expect(4);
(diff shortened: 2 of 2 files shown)


megawac:
Fine with this change, but did that not pass before?

jridgewell:
The old tests pass, but nothing tested calling the debounced function again after `wait` is over.

underscore.js
```diff
@@ -862,10 +862,10 @@
     var debounced = restArgs(function(args) {
       var callNow = immediate && !timeout;
```
captbaritone: This is now only used inside the `if(immediate) {` block. Could we move it into that block?

jridgewell: Updated.

bent0b0x:
Is thee a way to tell in which version of underscore this is fixed?

hanzichi:
@bent0b0x it was not in the released versions,  it was fixed in the edge version, so the bug is not in any one of the released ones

PR #2483, opened 2016-03-24
Make no-use-before-define eslint rule a warning.

afahim:
As per https://github.com/jashkenas/underscore/pull/2480#discussion_r57190276, changing this to a warning.


captbaritone:
Thanks @afahim!

Issue #2370, opened 2015-12-02
Adding ability for deep object property lookup (_.get)


patrickml:
This is simular to that of lodash and i believe there is a big use for this functionality to be added to the core. One of the biggest benefits of this is that you no longer have to check for undefined up the object chain like so

Before `this.object && this.object.one && this.object.one.two`
Now `_.get(this.object, 'one.two')`

I found most of this on another post I cannot remember who did it but this is very useful

```js
_.mixin({
  get: function (obj, key) {
      var type = typeof key;

    if (type == 'string' || type == "number") {
      key = ("" + key).replace(/\[(.*?)\]/,/\[(.*?)\]/, function (m, key) { //handle case where [1] may occur
        return '.' + key.replace(/["']/g,/["']/g, ""); //strip quotes
      }).split('.');
    }
    for (var i = 0, l = key.length; i < l; i++) {
      if (typeof obj !== 'undefined' && _.has(obj, key[i])) obj = obj[key[i]];
      else return undefined;
    }
    return obj;
  }
});
```

akre54:
This has been proposed and rejected a few times (search for "string programming" or "deep path" and you'll find #2268, #1169, #1266 and others). Also check out underscore-contrib.

NarHakobyan:
use underscore.get module

`const {get} = require('underscore.get');`

Issue #2268, opened 2015-08-07
[Discussion] Best practices for fetching deep value in an object


paramaggarwal:
What we know till now:
1. Using strings to manipulate keys in JS is bad practice. (ref: ["string programming"](https://github.com/jashkenas/underscore/search?q=%22string+programming%22&type=Issues&utf8=✓))
2. Both `underscore-contrib` and `lodash` implement `getPath` for this.
3. People are confused about this. (ref: #555, #712, #1137, #1169, #1659, #1822, #1889, #1998)

What are the best practices around safely fetching `object.child.child.child.child.name` with purely the use of `_` methods?

jdalton:
Maybe
```js
var o = { 'a': { 'b': { 'c': 3 } } };
_.reduce(['a', 'b', 'c'], _.partial(_.result, _, _, void 0), o);
```

rafde:
> 1. Using strings to manipulate keys in JS is bad practice. (ref: "string programming")
Why? It's convenient and smaller than `['object', 'child', 'child', 'child', 'child', 'name']` vs `object.child.child.child.child.name` which is more telling that the former.

megawac:
@rafde because it becomes tricky to interpret some of the path nuance, edge cases

```js
var x = {"a['x.a.\y']": 1};

x["a['x.a.\y']"]
```

jdalton:
I had Underscore's stance until user demand for it was overwhelming. Turns out edge cases are manageable and devs dig the shorter syntax. Baking it into core means devs get it in `_.pluck`, `_.filter`, `_.result`, `_.invoke`, `_.has`, and many other methods which would be cumbersome or not-possible otherwise.

rafde:
> var x = {"a['x.a.\y']": 1};

Fair statement. Although I have no idea why anyone would make a key name like that. But apparently it's possible so it's really a nice to have.

jdalton:
@rafde Issues can be avoided by checking if a path like `'x.a.\y'` exists as a single prop before crawling.

jashkenas:
I think it would be nice to continue to avoid string programming. That was actually one of the original motivations for releasing Underscore in the first place — kinda like Oliver Steele's Functional JavaScript (http://osteele.com/sources/javascript/functional/) ... but without all the string programming.

jdalton:
You could spin it that its less string programming and more paths (less magic than say 'x*x').
There just isn't a good alternative for a lack of this feature. With CS you could do
```js
[...]

Issue #1266, opened 2013-08-22
Add _.get(obj, "a.b.c.d") for easier duck typing


bradleyayers:
It would be handy to have a function like `_.result` that can handle multiple chained attribute lookups, and doesn't call the function at the end. This is useful for cases where I want branch based on if an API exists. e.g.

```
var foo = {bar: {baz: "qux"}};
_.get(foo, "bar.baz");
→ "qux"
```

A special case is when there's actually an attribute with a `.` in it. To handle this, the API could handle take an array:

```
var foo = {"bar.baz": "qux"};
_.get(foo, ["bar.baz"]);
→ "qux"
```

If you _did_ want to call the result if it's a function, you could just wrap it in `_.result`:

```
var foo = {"bar.baz": "qux"};
_.result(_.get(foo, ["bar.baz"]));
→ "qux"
```

PR #2486, opened 2016-03-30
Allow deep property matching via array syntax

captbaritone:
Inspired by `underscore-contrib`'s _.getPath` and lodash's deep property access, this adds a deep property access array shorthand to many aspects of Underscore, including:

* `_.property`
* `_.iteratee` and all methods that depend on it.

It also exposes a convenience function `_.get` for accessing deep properties with a default.

Possible Concerns
-----------------

While `_.property` does not get called in any place that would present a performance concern, `getLength` (which is created by `_.property`) gets used in many places which _may_ have performance ramifications.

I've placed the `isArray` call in the outer function of `_.property` in an attempt to reduce complexity inside the generated function. This required defining `isArray` earlier.

It's possible that this still isn't fast enough, and we should just define `getLength` manually.

Adds ~0.06kB (minified/gzipped)

Further work
------------

The following methods should probably also accept the array syntax:

* `_.result`
* `_.has`

The following companion methods should be created:

* `_.set`

Fixes: #2372, an elegant solution to an oft requested feature: (see: #2370, #2268, #1169, #1266)

### DIFF
--- a/test/functions.js
+++ b/test/functions.js
@@ -694,6 +694,9 @@
       assert.deepEqual(_.toArray(cb(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)), _.range(1, 11));
     });
 
+    var deepProperty = _.iteratee(['a', 'b']);
+    assert.equal(deepProperty({a: {b: 2}}), 2, 'treats an array as a deep property accessor');
+
     // Test custom iteratee
     var builtinIteratee = _.iteratee;
     _.iteratee = function(value) {
--- a/underscore.js
+++ b/underscore.js
@@ -93,7 +93,7 @@
     if (_.iteratee !== builtinIteratee) return _.iteratee(value, context);
     if (value == null) return _.identity;
     if (_.isFunction(value)) return optimizeCb(value, context, argCount);
-    if (_.isObject(value)) return _.matcher(value);
+    if (_.isObject(value) && !_.isArray(value)) return _.matcher(value);
     return _.property(value);
   };
 
@@ -138,9 +138,18 @@
     return result;
   };
 
-  var property = function(key) {
+  var isArray = nativeIsArray || function(obj) {
+    return toString.call(obj) === '[object Array]';
+  };
+
+  var property = function(path) {
+    if (!isArray(path)) path = [path];
     return function(obj) {
-      return obj == null ? void 0 : obj[key];
+      var index = 0, length = path.length;
+      while (obj != null && index < length) {
+        obj = obj[path[index++]];
+      }
+      return obj == null ? void 0 : obj;
     };
   };
 
@@ -1279,9 +1288,7 @@
 
   // Is a given value an array?
   // Delegates to ECMA5's native Array.isArray
-  _.isArray = nativeIsArray || function(obj) {
-    return toString.call(obj) === '[object Array]';
-  };
+  _.isArray = isArray;
 
   // Is a given variable an object?
   _.isObject = function(obj) {
@@ -1372,8 +1379,8 @@
 
   // Generates a function for a given object that returns a given property.
   _.propertyOf = function(obj) {
-    return obj == null ? function(){} : function(key) {
-      return obj[key];
+    return function(key) {
+      return property(key)(obj);
     };
   };
 
@@ -1444,6 +1451,13 @@
       value = fallback;
     }
     return _.isFunction(value) ? value.call(object) : value;
+  };
+
+  // Returns the, possibly deeply nested, value at `path` if defined.
+  // Otherwise, returns `fallback`.
+  _.get = function(object, path, fallback) {
+    var value = property(path)(object);
+    return value === void 0 ? fallback : value;
   };
 
   // Generate a unique integer id (unique within the entire client session).
--- a/test/utility.js
+++ b/test/utility.js
@@ -75,6 +75,12 @@
     assert.equal(_.property('name')(stooge), 'moe', 'should return the property with the given name');
     assert.equal(_.property('name')(null), void 0, 'should return undefined for null values');
     assert.equal(_.property('name')(void 0), void 0, 'should return undefined for undefined values');
+    assert.equal(_.property(null)('foo'), void 0, 'should return undefined for null object');
+
+    // Deep property access
+    assert.strictEqual(_.property('a')({a: 1}), 1, 'can get a direct property');
+    assert.strictEqual(_.property(['a', 'b'])({a: {b: 2}}), 2, 'can get a nested property');
+    assert.strictEqual(_.property(['a'])({a: false}), false, 'can fetch falsey values');
   });
 
   QUnit.test('propertyOf', function(assert) {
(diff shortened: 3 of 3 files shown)


underscore.js
```diff
@@ -1279,9 +1288,7 @@
 
   // Is a given value an array?
   // Delegates to ECMA5's native Array.isArray
-  _.isArray = nativeIsArray || function(obj) {
-    return toString.call(obj) === '[object Array]';
-  };
+  _.isArray = isArray;
```
akre54: Can we remove or at least change the eslint rule? It's totally valid to lazily define methods on the `_` object provided they're not getting used right away.

captbaritone: Unfortunately this time it _is_ getting called right away. The outer `_.property` function (which uses `isArray`) gets called when we create `getLength`.

akre54: :/

What if we re-arranged the sections so that the `_.property` parts comes after `_.is[...]`?

underscore.js
```diff
@@ -1446,6 +1453,13 @@
     return _.isFunction(value) ? value.call(object) : value;
   };
 
+  // Returns the, possibly deeply nested, value at `path` if defined.
+  // Otherwise, returns `fallback`.
+  _.get = function(object, path, fallback) {
+    var value = property(path)(object);
+    return value === void 0 ? fallback : value;
+  };
```
megawac: If we add `_.get` are we obliged to add `_.set`, `_.unset`?

Also you missed `_.result`

captbaritone: I mentioned `_.set`, `_.has` and `_.result` in the PR message as "Further Work". I wasn't sure if we wanted to tackle them all at once, or one at a time. (We could also pull `_.get` out of this PR if that would simplify things)

`_.unSet` was not on my radar.

To answer your question directly, I'm guessing we would be obliged, but I figured this PR would be a good forum to make that decision as a group.

akre54: I missed that in the commit message since I don't usually see the commit messages directly. We typically leave commit messages short (~1 sentence) and then write longer prose in the body of the Pull Request. Mind either copying that message to the pull request, or pointing out that there might be more text in the commit message in the future?

captbaritone: Hmm. Maybe I did something strange? I believe I posted (nearly) the same content in the commit and the pull request message. Here's what I see:

screen shot 2016-03-30 at 8 26 14 am

If that's not what you're seeing, let me know and I can try opening a fresh PR.

akre54: Oh no you're totally right. They just looked different to me. Carry on.

captbaritone: I'm going to delete these last few PR-message-related comments so that they don't distract from the topic at hand.

megawac: Okay, I'm aware you mentioned them, I didn't phrase my response well so apologies for that (was rushing to submit a paper). What I meant is what is the motivation and value of adding `_.get` (and committing to `_.set` and `_.unset`) rather than modifying `_.result` for paths.

captbaritone: I see. Yeah it make sense to start with `_.result` since it would be a superset of `_.get`'s functionality, and we can do it without committing to `_.set` and `_.unset` (We probably will need to update `_.has` as well).

Do you think we should aim for evaluating functions at every step a la https://github.com/lodash/lodash/commit/f65d36b3c1d87c95534e56fba98b576a452e8fb7 ?

captbaritone:
Based on @megawac's feedback, I'm reworking this PR to:

1. Not include `_.get`
2. Add deep matching to `_.has` and `_.result`

It may be difficult to do so without adding some weight, so I figure we should do them all at once so we know what we're getting ourselves into.

underscore.js
megawac: I'm not sure to be honest. I've never wanted that functionality personally, but the article that the commit you referenced was in response to was pretty convincing (ko.observables support)

megawac:
Sounds good, for now I prefer that direction

test/functions.js
```diff
@@ -694,6 +694,9 @@
       assert.deepEqual(_.toArray(cb(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)), _.range(1, 11));
     });
 
+    var deepProperty = _.iteratee(['a', 'b']);
+    assert.equal(deepProperty({a: {b: 2}}), 2, 'treats an array as a deep property accessor');
+
```
jdalton: Compat note:
Lodash treats an array as shorthand for `matchesProperty`.

The `matchesProperty` shorthand is `[propName,propValue]` and as fast path for
the common case of `{propName:propValue}`


Not a biggie but something to be aware of.

captbaritone: @jdalton Thanks for bringing this up. I'm a bit confused though. In lodash you use arrays as shorthand for a type of predicate which can already be expressed via an object (`_.matches`)?

I'm guessing from the phrase "fast path" and the perf graph you included that this somehow offers a performance gain? (Or maybe I'm misunderstanding?)

To me, deep property matching seems much more useful than a performance increase which users could presumably get in the places where performance was a concern by simply using `_.matchesProperty()` or their own optimized function directly.

Again, I could just be misinterpreting your comment, so feel free to set me straight. :smile:

underscore.js
captbaritone: @akre54 I'm not sure I understand. If we move `_.property` below where the current `_.is[...]` methods live, it' won't be instantiated when we go to create `getLength`.

test/functions.js
jdalton: > In lodash you use arrays as shorthand for a type of predicate which can already be expressed via an object (`_.matches`)?

Yep. The `_.matches` shorthand is a bit heavier as it supports multiple properties.
You can think of `_.matches` implemented by way of `_.matchesProperty`. The `_.matches` method handles a collection of source properties to compare.

> I'm guessing from the phrase "fast path" and the perf graph you included that this somehow offers a performance gain?

Yep. It's a common enough case that a fast path was created.

> To me, deep property matching seems much more useful than a performance increase which users could presumably get in the places where performance was a concern by simply using `_.matchesProperty()` or their own optimized function directly.

In lodash `matches` and `matchesProperty` already perform partial deep matches of source objects:
```js
_.filter(objects, { 'a': { 'b': { 'c': 1 }, 'd': 2 } });
```
In Lodash the common case for deep property paths is strings in with arrays as a fallback for edge cases which is where `_.property(array)` would be used.
```js
_.filter(objects, 'a.b.c');
```

captbaritone: Ah this all makes sense. Thanks for the clarification.

jridgewell: lodash's `matchesProperty` is tripping a few people up, but I can't seem to find the issues. And I'm biased, because I dislike the `a.b.c` deep property syntax. `['a', 'b', 'c']` FTW.

jdalton: I dig `a.b.c` but couldn't see supporting one without the other.
The array allows supporting symbols and other funky keys. I :heart: funk:

jdalton: Folks wanting cross-lib interop can just use the explicit forms like `_.property`.

captbaritone: Assuming Underscore is philosophically opposed to the `a.b.c` syntax, I think it makes most sense to treat arrays as deep property matching:

    // Underscore only (proposed)
    var adminPosts = _.filter(posts, ['author', 'isAdmin']);

    // Lodash only
    var adminPosts = _.filter(posts, 'author.isAdmin');
    var adminPosts = _.filter(posts, {author, {isAdmin: true}});

    // Compat (proposed)
    var adminPosts = _.filter(posts, _.property(['author', 'isAdmin']));

captbaritone:
Closing in favor of #2494

test/functions.js
jdalton: I've [created a poll](https://twitter.com/jdalton/status/718165294678552576) as a gut check. Feel free to RT.

jdalton: Results:

captbaritone: The Tweet responses add some interesting color as well. Thanks for running that @jdalton. That being said, I don't think this thread is actually about (re)considering the string syntax. It seems pretty clear to me that the string syntax is not something Underscore wants to support.

jdalton: Additional info at least lets them know they aren't inline with their users.

PR #2490, opened 2016-04-04
Prefer `assert.notOk()` to `assert.ok(!)`

captbaritone:
In #2389 we upgraded our browser tests to QUnit 1.18.0 which means we can now
use the `assert.notOk()` assertion. Using this gives a less confusing error
message when the assertion fails. Assuming my `shouldBeFalse()` function is
failing:

    assert.ok(!shouldBeFalse());
    => failed, expected argument to be truthy, was: false

    assert.notOk(shouldBeFalse());
    => failed, expected argument to be falsy, was: true


akre54:
Merge conflicts...

captbaritone:
@akre54 Updated.

PR #2491, opened 2016-04-04
Use descriptive assertions

captbaritone:
Using descriptive assertions can help give better failure messages.

These were found using @platinumazure's [`eslint-plugin-qunit`](https://github.com/platinumazure/eslint-plugin-qunit). Once it reaches version 1.0, we could consider including it as part of our linting.

Note: this will have merge conflicts with #2490. I'll rebase as needed.


akre54:
thanks

PR #2493, opened 2016-04-05
Point users to Gitter instead of IRC

captbaritone:


akre54:
We've [discussed this a bit before](https://github.com/jashkenas/backbone/pull/3735). I'm ok with merging this for practicality reasons, but let the record show in general I'm against the principle of moving from a free / open federated protocol to a more proprietary-ish platform.

Why not show both?

captbaritone:
@akre54 Updated.

captbaritone:
I've cherry-picked onto `gh-pages.

PR #2494, opened 2016-04-07
Allow deep property matching via array syntax

captbaritone:
Replaces #2486

Inspired by underscore-contrib`'s _.getPath` and lodash's deep property access,
this adds an Array deep property access shorthand to many aspects of
Underscore, including:

* `_.property`
* `_.iteratee` and all methods that depend on it.
* `_.has`
* `_.result` (including deep evaluation)

Possible Concerns
=================

`_.property`
------------

While `_.property` does not get called in any place that would present
a performance concern, `getLength` (which is created by `_.property`) gets used
in many places which _may_ have performance ramifications.

I've placed the `isArray` call in the outer function of `_.property` in an
attempt to reduce complexity inside the generated function. This required
defining `isArray` earlier.

It's possible that this still isn't fast enough, and we should just define
`getLength` manually.

`_.has`
-------

This adds some complexity to `_.has` which underpins many methods. It's
possible that we may want to use a simpler "flat" version of `_.has`
internally.

Further work
============

We may want to expose some additional helper methods:

* `_.get`
* `_.set`
* `_.unset`

Fixes: #2372, an elegant solution to an oft requested feature: (see: #2370,
 #2268, #1169, #1266)

### DIFF
--- a/test/utility.js
+++ b/test/utility.js
@@ -365,6 +365,20 @@
     assert.strictEqual(_.result({}, [], 'a'), 'a', 'returns the default when prop is empty');
     assert.strictEqual(_.result(obj, [], context), obj, 'uses the object as context when path is empty');
 
+    var nested = {
+      d: function() {
+        return {
+          e: function() {
+            return obj;
+          },
+          f: context
+        };
+      }
+    };
+    assert.strictEqual(_.result(nested, ['d', 'e']), obj, 'can unpack nested function calls');
+    assert.strictEqual(_.result(nested, ['d', 'f']).e(), obj, 'uses parent as context for nested function calls');
+    assert.strictEqual(_.result(nested, ['d', 'x'], context).e(), obj, 'uses last valid child as context for fallback');
+
     if (typeof Symbol !== 'undefined') {
       var x = Symbol('x');
       var symbolObject = {};
--- a/test/utility.js
+++ b/test/utility.js
@@ -336,6 +336,35 @@
     }), obj.a, 'called with context');
   });
 
+  QUnit.test('result can accept an array of properties for deep access', function(assert) {
+    var func = function() { return 'f'; };
+    var context = function() { return this; };
+
+    assert.strictEqual(_.result({a: 1}, 'a'), 1, 'can get a direct property');
+    assert.strictEqual(_.result({a: {b: 2}}, ['a', 'b']), 2, 'can get a nested property');
+    assert.strictEqual(_.result({a: 1}, 'b', 2), 2, 'uses the fallback value when property is missing');
+    assert.strictEqual(_.result({a: 1}, ['b', 'c'], 2), 2, 'uses the fallback value when any property is missing');
+    assert.strictEqual(_.result({a: void 0}, ['a'], 1), 1, 'uses the fallback when value is undefined');
+    assert.strictEqual(_.result({a: false}, ['a'], 'foo'), false, 'can fetch falsey values');
+
+    assert.strictEqual(_.result({a: func}, 'a'), 'f', 'can get a direct method');
+    assert.strictEqual(_.result({a: {b: func}}, ['a', 'b']), 'f', 'can get a nested method');
+    assert.strictEqual(_.result(), void 0, 'returns udefined if obj is not passed');
+    assert.strictEqual(_.result(void 1, 'a', 2), 2, 'returns default if obj is not passed');
+    assert.strictEqual(_.result(void 1, 'a', func), 'f', 'executes default if obj is not passed');
+    assert.strictEqual(_.result({}, void 0, 2), 2, 'returns default if prop is not passed');
+    assert.strictEqual(_.result({}, void 0, func), 'f', 'executes default if prop is not passed');
+
+    var childObj = {c: context};
+    var obj = {a: context, b: childObj};
+    assert.strictEqual(_.result(obj, 'a'), obj, 'uses the parent object as context');
+    assert.strictEqual(_.result(obj, 'e', context), obj, 'uses the object as context when executing the fallback');
+    assert.strictEqual(_.result(obj, ['a', 'x'], context), obj, 'uses the object as context when executing the fallback');
+    assert.strictEqual(_.result(obj, ['b', 'c']), childObj, 'uses the parent as context when accessing deep methods');
+
+    assert.strictEqual(_.result({}, [], 'a'), 'a', 'returns the default when prop is empty');
+  });
+
   QUnit.test('_.templateSettings.variable', function(assert) {
     var s = '<%=data.x%>';
     var data = {x: 'x'};
--- a/test/utility.js
+++ b/test/utility.js
@@ -363,6 +363,7 @@
     assert.strictEqual(_.result(obj, ['b', 'c']), childObj, 'uses the parent as context when accessing deep methods');
 
     assert.strictEqual(_.result({}, [], 'a'), 'a', 'returns the default when prop is empty');
+    assert.strictEqual(_.result(obj, [], context), obj, 'uses the object as context when path is empty');
 
     if (typeof Symbol !== 'undefined') {
       var x = Symbol('x');
--- a/test/utility.js
+++ b/test/utility.js
@@ -363,6 +363,18 @@
     assert.strictEqual(_.result(obj, ['b', 'c']), childObj, 'uses the parent as context when accessing deep methods');
 
     assert.strictEqual(_.result({}, [], 'a'), 'a', 'returns the default when prop is empty');
+
+    if (typeof Symbol !== 'undefined') {
+      var x = Symbol('x');
+      var symbolObject = {};
+      symbolObject[x] = 'foo';
+      assert.strictEqual(_.result(symbolObject, x), 'foo', 'can use symbols as keys');
+
+      var y = Symbol('y');
+      symbolObject[y] = {};
+      symbolObject[y][x] = 'bar';
+      assert.strictEqual(_.result(symbolObject, [y, x]), 'bar', 'can use symbols as keys for deep matching');
+    }
   });
 
   QUnit.test('_.templateSettings.variable', function(assert) {
--- a/underscore.js
+++ b/underscore.js
@@ -1354,7 +1354,7 @@
       }
       obj = obj[key];
     }
-    return true;
+    return !!length;
   };
 
   // Utility Functions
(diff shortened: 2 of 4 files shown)


underscore.js
```diff
@@ -1436,14 +1451,25 @@
...
+    for (var i = 0; i < length; i++) {
+      var prop = object == null ? void 0 : object[path[i]];
+      if (prop === void 0) {
+        prop = fallback;
+        i = length; // Ensure we don't continue iterating.
+      }
+      object = _.isFunction(prop) ? prop.call(object) : prop;
     }
-    return _.isFunction(value) ? value.call(object) : value;
+    return object;
```
captbaritone: Credit to @jdalton on for this one, as it's lifted almost directly from lodash. I spent a few days picking at it, and couldn't come up with a better implementation.

test/utility.js
```diff
@@ -360,6 +369,35 @@
...
+
+    var childObj = {c: context};
+    var obj = {a: context, b: childObj};
+    assert.strictEqual(_.result(obj, 'a'), obj, 'uses the parent object as context');
+    assert.strictEqual(_.result(obj, 'e', context), obj, 'uses the object as context when executing the fallback');
+    assert.strictEqual(_.result(obj, ['a', 'x'], context), obj, 'uses the object as context when executing the fallback');
+    assert.strictEqual(_.result(obj, ['b', 'c']), childObj, 'uses the parent as context when accessing deep methods');
+
+    assert.strictEqual(_.result({}, [], 'a'), 'a', 'returns the default when prop is empty');
+  });
```
jdalton: Might add tests for symbols.

jdalton:
Related to https://github.com/jashkenas/backbone/pull/3988.

captbaritone:
Thoughts @megawac ?

underscore.js
```diff
@@ -138,9 +138,18 @@
...
+  };
+
+  var property = function(path) {
+    if (!isArray(path)) path = [path];
+    var length = path.length;
+    return function(object) {
+      for (var i = 0; object != null && i < length; i++) {
+        object = object[path[i]];
+      }
+      return object == null ? void 0 : object;
```
megawac: is `null` not a valid result?

Consider

```js
_.property('x')({x: null})
```

underscore.js
```diff
@@ -1340,8 +1347,16 @@
 
   // Shortcut function for checking if an object has a given property directly
   // on itself (in other words, not on a prototype).
-  _.has = function(obj, key) {
-    return obj != null && hasOwnProperty.call(obj, key);
+  _.has = function(object, path) {
+    if (!_.isArray(path)) path = [path];
```
megawac: For this and property I wonder if it'd be better to fast track it (i.e. use the old implementation)

captbaritone: I had a similar concern. Especially for `_.has`. Do you have any ideas of ways we could validate that choice, or does it seem like an obviously worthwhile performance tweak? I must admit, the performance characteristics of JavaScript, and especially performance _tests_ are often anti-intuitive to me.

underscore.js
captbaritone: @megawac Good catch. Fixed with https://github.com/jashkenas/underscore/pull/2494/commits/587ce446105d2c1dda55821213803bde49953e8c

captbaritone:
This is ready for another review. I've fast-tracked the shallow case for each function, and cleaned up a few edge cases.

underscore.js
```diff
@@ -1370,12 +1381,27 @@
...
 
   // Generates a function for a given object that returns a given property.
   _.propertyOf = function(obj) {
-    return obj == null ? function(){} : function(key) {
-      return obj[key];
+    if (obj == null) {
+      return function(){};
+    }
+    return function(key) {
+      return _.property(key)(obj);
```
jridgewell: Man, this is expensive.

underscore.js
```diff
@@ -1370,12 +1381,27 @@
 
   _.noop = function(){};
 
-  _.property = property;
+  _.property = function(path) {
+    if (!_.isArray(path)) {
+      return shallowProperty(path);
+    }
+    var length = path.length;
```
jridgewell: Let's move this inside the function.

underscore.js
```diff
@@ -1438,14 +1464,26 @@
...
-    var value = object == null ? void 0 : object[prop];
-    if (value === void 0) {
-      value = fallback;
+  // Traverses the children of `obj` along `path`. If a child is a function, it
+  // is invoked with its parent as context. Returns the value of the final
+  // child, or `fallback` if any child is undefined.
+  _.result = function(obj, path, fallback) {
+    if (!_.isArray(path)) path = [path];
+    var length = path.length;
+    // If path is `[]`, step through the loop once so `fallback` gets used.
```
jridgewell: I'd rather just:

```js
return _.isFunction(fallback) ? fallback() : fallback;

File: underscore.js (hunk 4 above)
Comment (captbaritone) [inline underscore.js]: I've updated with a version that uses the old code for the `!_.isArray` case. Another option, would be to keep the old version as a private function and continue to use it internally. That will avoid the array check for all internal uses. This would be similar to the approach I took for `shallowProperty()/_.property`.

File: test/utility.js [hunk 8]
```diff
@@ -361,6 +375,47 @@
     }), obj.a, 'called with context');
   });
 
+  QUnit.test('result can accept an array of properties for deep access', function(assert) {
```
jridgewell: We need nested function call tests:

```js
  var nested = {
    a() {
      return {
        b() {
          return 1;
        }
      }
    },
  };
  assert.strictEqual(_.result(nested, ['a', 'b'], 1));

underscore.js
captbaritone: My thought was that eventually the deep case for both `_.property` and `propertyOf` could call out to `_.get`. Or we could give each their own implementation. Do you think the weight savings is worth the perf hit? I was feeling self-conscious about my weight ;)

underscore.js
captbaritone: Went with:

    return _.isFunction(fallback) ? fallback.call(obj) : fallback;

To preserve the context.

test/utility.js
captbaritone: Added.

underscore.js
jridgewell: We already have the implementations, we just need to separate them out from their partial applications (the function returned from `_.property` should be moved out).

captbaritone:
Feedback addressed.

jridgewell:
LGTM, thought we should definitely have someone else weigh in.

jashkenas:
I think this sort of thing makes sense — only if you do it all at once, and update the Underscore API to accept deep path traversals in any function where they'd be useful.

For example: `_.findWhere`, where you're looking for a match on a deep property within the objects.

I'm not seeing that sort of thing in the tests. Does this patch do that?

captbaritone:
@jashkenas We are only implementing array-style deep property access, and since an array cannot reasonably be used as an object key, this does not affect anything that uses a `_.matcher` object.

If we wanted to support deep property matching with `_.matcher`'s object syntax, we would need to allow for a string-based deep property access syntax (like `"a.b.c"` for example), which I believe, based upon previous discussion, has been ruled too magical.

By modifying `_.property` (and thus `_.iteratee`), `_.has` and `_.result` I believe we are adding the array-style deep property access to all places in Underscore that currently offer shallow property access via a passed string.

You bring up a good question regarding test. Should we add explicit tests to each function that we expect to support this syntax, or would that be redundant?

bjmiller:
> If we wanted to support deep property matching with `_.matcher`'s object syntax, we would need to allow for a string-based deep property access syntax (like `"a.b.c"` for example), which I believe, based upon previous discussion, has been ruled too magical.

Given how successful and rock-solid path strings have been in lodash, I wonder if an implementation here might be worth revisiting?  I only say this because it's now a battle-tested and widely appreciated concept.  When last this was discussed, a lot of people wanted it, but it seemed like a large and brittle proposition to write a mini-parser for path strings.  Things might look different now.

captbaritone:
@bjmiller Given the (possibly) controversial nature of path strings, and that the array syntax can be merged as a standalone enhancement, I'd rather keep this pull request discussion focused on just array syntax. Once/if we have the array syntax, we can readdress the string syntax question.

jashkenas:
> Given how successful and rock-solid path strings have been in lodash, I wonder if an implementation here might be worth revisiting? I only say this because it's now a battle-tested and widely appreciated concept. When last this was discussed, a lot of people wanted it, but it seemed like a large and brittle proposition to write a mini-parser for path strings. Things might look different now.

I agree. *If* we're going to add deep path access, we should add it with the most desirable API, and not a clunky one.

captbaritone:
Lodash implements both the string _and_ array syntax. This is because there are some keys (for example those containing a `.` character) which cannot be expressed unambiguously with the string syntax. The array syntax is also preferable for the case where a key is dynamic. I don't think it's reasonable to implement the string syntax _without_ the array syntax. As I see it, the string syntax is an additional layer to syntacic sugar on top of the array syntax. (See figure 1):

jaw breaker cross section

-- fig. 1

Does anyone (@jashkenash ?) believes both of the following?

1. The array syntax is only worth implementing if we are _also_ planning on implementing the string syntax.
2. It is nessesary to see a working implementation of the string syntax before we commit to supporting it.

If so, I'd be willing to start working on a string syntax implementation. However, I'm a firm believer that smaller pull requests are preferable where possible, since they are easier to review and allow for more focused discussion.

jashkenas:
> Does anyone (@jashkenash ?) believes both of the following?
> 
> The array syntax is only worth implementing if we are also planning on implementing the string syntax.
> It is nessesary to see a working implementation of the string syntax before we commit to supporting it.

I lukewarm believe #1, but I don't believe #2. If we want to go forward with this, this ticket is a fine place to start. LGTM.

jridgewell:
Lodash's string syntax is built using the array syntax's internals. This is the necessary first step.

captbaritone:
@jridgewell Shall we merge?

jridgewell:
Let's open a separate issue to discuss the string path.

captbaritone:
Will do. My next steps:

1. Update documentation to reflect array shorthand.
2. Pull request with proposed implementation of `_.get` (optionally `_.set` and `_.unset` at the same time).

captbaritone:
I've opened an issue to discuss the string path syntax: #2609

PR #2496, opened 2016-04-13
Optimization of _.isNaN

marijaselakovic:
You might consider optimizing _.isNaN function  by changing the order in which checks are executed.  I run all tests and observed that isNaN(obj) is false for most of cases. By putting isNaN(obj) to be checked first, every time when it is false, the time spent in _.isNumber(obj) is saved. 

By swapping checks, I got the following performance improvements on V8 engine:
Test: isNaN - 38%
Test: indexof with NaN - 46,37% 
Test: last indexof with NaN  - 23% 
Test: includes with NaN - 61,73% 

Even though this is a simple change, it seems to improve performance without affecting the readability of the code.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -1316,7 +1316,7 @@
 
   // Is the given value `NaN`?
   _.isNaN = function(obj) {
-    return _.isNumber(obj) && isNaN(obj);
+    return isNaN(obj) && _.isNumber(obj);
   };
 
   // Is a given value a boolean?


megawac:
Thanks

michaelficarra:
I would've liked to have seen the benchmark data before merging this.

marijaselakovic:
Do you have benchmark for assessing the performance of underscore? I would like to run it too.

akre54:
Yeah we typically use [jsperf](https://jsperf.com/) but it's down at the moment (see jsperf/jsperf.com#18). In the meantime we can bench using benchmark.js, or just a plain loop with `performance.now`.

marijaselakovic:
After modifying the code, I run all unit tests, such that each test is repeatedly executed several times (including time for warm up) in 5 fresh VM instances. Based on obtained measurements I returned tests for which the difference in execution time is statistically significant (the same methodology as Benchmark.js). I already reported the results.

megawac:
Yeah I benchmarked this as well a while ago but never submitted a patch so
was fine with merging

michaelficarra:
I was not questioning the *methodology*, I was questioning the *data*.

marijaselakovic:
I reported the data in the description and explained the methodology later to show how the data are obtained.

michaelficarra:
The data used as input to the function during the benchmark.

marijaselakovic:
I created benchmarks for 4 tests mentioned above and used the same input as in unit tests. You can download it from https://s3.eu-central-1.amazonaws.com/underscorebenchmark/benchmarks.zip. To run, just call 

> node run.js

from benchmarks directory (the output is the same as in jsperf, because I used Benchmark.js library)

CodeFalling:
However, when `obj = Symbol()`

```js
isNaN(obj)   ;; Uncaught TypeError: Cannot convert a Symbol value to a number
obj !== +obj ;; false 
```

PR #2499, opened 2016-04-14
optimization in flatten function

marijaselakovic:
Similarly as in [https://github.com/jashkenas/underscore/pull/2496](url), I found that  _flatten_ function can be optimized by swapping the order of checks in logical expression in line 500 (the second check is false most of the time). I provide the screenshot with the results for unit tests which performance got improved  (on V8 4.4 engine). Inputs are the same as in unit tests in the repository and performance is tested with Benchmark.js library.
selection_012

The code for tests can be download from: https://s3.eu-central-1.amazonaws.com/underscorebenchmark/benchmarks2.zip 

To run the tests, just call 

> node run.js

I am maybe wrong, but can isArrayLike(value) check be omitted here?

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -505,7 +505,7 @@
     var idx = output.length;
     for (var i = 0, length = getLength(input); i < length; i++) {
       var value = input[i];
-      if (isArrayLike(value) && (_.isArray(value) || _.isArguments(value))) {
+      if ((_.isArray(value) || _.isArguments(value)) && isArrayLike(value)) {
         // Flatten current level of array or arguments object.
         if (shallow) {
           var j = 0, len = value.length;


megawac:
Yes, `isArrayLike` can be removed here

jridgewell:
> Yes, isArrayLike can be removed here

Not quite. This opt will only benefit `isArray` since it's faster than `isArrayLike + isArray` on non-array objects. But this'll hurt the `isArguments` check, since that uses the slower `toString` comparison.

Instead, we should guard only the `isArguments` check:

```js
if (_.isArray(value) || (isArrayLike(value) && _.isArguments(value)) {
  //...
}

jgonggrijp:
`isArrayLike` *could* be removed, as it is only there to improve performance. The current check is already optimal, however:

&nbsp; | arrays | arguments | other
---|---|---|---
current | `isArrayLike` + `isArray` | `isArrayLike` + `isArray` + `isArguments` | `isArrayLike`
marijaselakovic | `isArray` + `isArrayLike` | `isArray` + `isArguments` + `isArrayLike` | `isArray` + `isArguments` + `isArrayLike`
megawac | `isArray` | `isArray` + `isArguments` | `isArray` + `isArguments`
jridgewell | `isArray` | `isArray` + `isArrayLike` + `isArguments | `isArray` + `isArrayLike`

The `toString`-based checks `isArray` and `isArguments` are more expensive than the heuristic `isArrayLike`. See the comments in the resolved part of the second review in #2840. Avoiding `isArrayLike` for arrays is only a small optimization while having `isArray` for other values is a big pessimization.

PR #2500, opened 2016-04-14
[add] a simple wrapper function for _.identity(), with slightly different semantic meaning.

rexfordkelly:
I found this helpful when learning to use underscore and while helping others learn to use underscore.js


akre54:
Thanks but no thanks.

Issue #2456, opened 2016-03-01
_.eq


jdalton:
Related to #2453. An `_.eq` method to enable the [SameValueZero](http://ecma-international.org/ecma-262/6.0/#sec-samevaluezero) comparison for composition.

mikefaraponov:
Hello, @jdalton! Can you give an use case example for composition of  this functions?

jdalton:
Sure thing. See [lodash/blob/npm/eq.js](https://github.com/lodash/lodash/blob/4.6.1-npm/eq.js).

jdalton:
Use case is to extract the `SameValueZero` comparison which would be what methods like `_.indexOf`, `_.includes`, `_.difference`, `_.intersection`, `_.lastIndexOf`, `_.uniq`, & `_.union`. It would be used much like `_.isEqual` without the deep equality comparisons.

mikefaraponov:
@jdalton, may I follow ECMA to implement SameValueZero functionality and why you doesn't like private method eq?

jdalton:
> may I follow ECMA to implement SameValueZero functionality

You can though it's just the long way around how I've implemented it (which is `===` + `NaN`).

> why you doesn't like private method 'eq'

Private helpers and their names didn't factor into it.

jdalton:
For reference [here are the spec steps](http://ecma-international.org/ecma-262/6.0/#sec-strict-equality-comparison) for `===`.

PR #2507, opened 2016-04-20
Implemented _.isEq() method.

aebrow4:
Added a SameValueZero check per [#2456 ](https://github.com/jashkenas/underscore/issues/2456)
Used the same implementation as @jdalton in [lodash](https://github.com/lodash/lodash/blob/4.6.1-npm/eq.js)

> _.isEq(a, b) performs a SameValueZero comparison on values a and b.
> http://ecma-international.org/ecma-262/6.0/#sec-samevaluezero

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -1264,6 +1264,13 @@
     return eq(a, b);
   };
 
+  // Performs a SameValueZero comparison (http://ecma-international.org/ecma-262/6.0/#sec-samevaluezero)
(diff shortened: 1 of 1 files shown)


megawac:
Why is an implementation of `SameValueZero` favourable over an implementation of `SameValue`?

jdalton:
`SameValueZero` is what `Map` and `Set` use so following that allows those to be used internally.

PR #2509, opened 2016-04-23
default data support for _.template()

eschmid72:
I've added default data support for the template() method. It makes it possible to

* add default data
* adding helper functions

Example:

```javascript
var tpl = '<%= greet(name) %>';

var hello = function(name){
  return 'Hello '+name+'!';
};

var hey = function(name){
  return 'Hey '+name+'!';
};

var defaultData = {
  greet: hello,
  name: 'Jack'
};

var settings = _.templateSettings;
var compiled = _.template(tpl,settings,settings,defaultData);
console.log(compiled());
// => "Hello Jack!"
console.log(compiled({name:'Bill'}));
// => "Hello Bill!"
console.log(compiled({greet:hey}));
// => "Hey Jack!"
```

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -1488,7 +1488,7 @@
   // Underscore templating handles arbitrary delimiters, preserves whitespace,
   // and correctly escapes quotes within interpolated code.
   // NB: `oldSettings` only exists for backwards compatibility.
-  _.template = function(text, settings, oldSettings) {
+  _.template = function(text, settings, oldSettings, defaultData) {
     if (!settings && oldSettings) settings = oldSettings;
     settings = _.defaults({}, settings, _.templateSettings);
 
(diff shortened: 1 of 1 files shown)


captbaritone:
I think Underscore expects you to solve this problem at the data layer using something like http://backbonejs.org/#Model-defaults would something like that work for you?

eschmid72:
Well, I must confess, I've never worked with backbonejs...

I've started to work on a "code factory", which should be able to render code templates / snippets from template files.
So I've been looking for a rendering engine, written in node.js, which can render the files and fill with data and I decided to use underscorejs.
I also need to be able to pass default values, helper functions and sub renderers to the code renderer, so I thought this little enhancement could make it easier.

bjmiller:
It looks like you're in the market for a more full-featured, dedicated templating library.  Maybe check out dust, or marko?

PR #2513, opened 2016-04-24
check length in _.first and _.last, fix #2495

### LINKED ISSUE #2495: _.first and _.last should check length
```js
Array.prototype[0] = 'boo';
_.first([]) // 'boo'
```
or
```js
var arr = [];
arr[-1] = 'boo'
_.last(arr); // 'boo'
```

CodeFalling:
Fix #2495

Now return `undefined` when use `_.first` or `_.last` on empty array.

 ```js
Array.prototype[0] = 'boo';
_.first([]) // undefined

var arr = [];
arr[-1] = 'boo'
_.last(arr); // undefined
```

### DIFF
--- a/test/arrays.js
+++ b/test/arrays.js
@@ -15,6 +15,10 @@
     result = _.map([[1, 2, 3], [1, 2, 3]], _.first);
     assert.deepEqual(result, [1, 1], 'works well with _.map');
     assert.equal(_.first(null), void 0, 'returns undefined when called on null');
+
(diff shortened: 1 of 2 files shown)


jridgewell:
Thanks!

PsiXdev:
I would like to understand the motivation for this change. Why in this case `first` should return `undefined` instead of `boo`? Moreover, this does not work completely for sparse arrays:
image
image

PR #2514, opened 2016-04-25
Make functions in docs filterable

captbaritone:
With a utility library like Underscore, it is very common to open the docs with the aim of locating the documentation for a specific function.

Currently I use my browser's search functionality, but since many functions reference other functions, I frequently have to cycle through multiple matches before I arrive at the actual function's description.

This patch aims to improve this common use case by offering a "filter" input above the list of functions. As a user types, the list of functions (and section titles) is reduced, leaving only the matching functions. If the user presses enter, they are jumped to the documentation for the first matching
function.

This functionality (and implementation) is greatly inspired by the implementation that is included in the [Ramda docs](http://ramdajs.com/0.21.0/docs/).

You can read their source code [here](https://github.com/ramda/ramda.github.io/blob/master/main.js)

To aid in review, you can find try out the functionality here: <https://rawgit.com/captbaritone/underscore/search/index.html>

Possible Improvements
---------------

1. The look of the search input could be improved.
2. The "autofocus" feature is not currently working on Firefox.


Florian-R:
The input seems a bit out of place (at least on FF 45, on Chromium it's OK). Could be cool to add it a width: 80% or something like this. 

Except this :+1: very nice work.

captbaritone:
@Florian-R Good idea. Added.

jridgewell:
👍. Someone else want to weigh in?

captbaritone:
I'll merge this later today, unless somebody objects.

captbaritone:
There are some merge conflicts between this PR and the `gh-pages`. I'll resolve those tomorrow or Saturday and push.

captbaritone:
I've cherry-picked this over to `gh-pages`. The only conflict was that `master` includes docs for `isSymbol` which is not yet in the current release.

jridgewell:
This might be the best thing to ever happen to underscore.

jashkenas:
This is super nice! @captbaritone — Would you be interested in adding the same functionality over on the Backbone docs?

captbaritone:
@jashkenas Glad you like it! @msrose and I will take a look at porting it to Backbone.

msrose:
FYI: https://github.com/jashkenas/backbone/pull/4075

PR #2518, opened 2016-04-27
Updated the _.debounce function to accept a function for the `wait` a…

paulocheque:
…ttribute.

Since I need a dynamic optimisation using a debounced function, I need to add some logic to determine the wait time to execute the function. For that, I let the _.debounce function to accept functions.

Example:

```
var dynamicWait = function() {
    if(a) { return 10; }
    else { return 1000; }
}
_.debounce(someFunc, dynamicWait);
```

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -861,13 +861,14 @@
     };
 
     var debounced = restArgs(function(args) {
+      var waitTime = _.isFunction(wait) ? wait() : wait;
       if (timeout) clearTimeout(timeout);
       if (immediate) {
         var callNow = !timeout;
-        timeout = setTimeout(later, wait);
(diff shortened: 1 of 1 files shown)


jgonggrijp:
This change looks relatively harmless, but I still see three reasons not to merge it:

- This is obscure functionality that almost nobody needs. Evidence: this PR was opened four years ago and since then nobody has asked for its continuation or even added a thumbs-up.
- I found a mention in the above linked PR (https://github.com/pluralsight/guides-cms/pull/47) that the project that does need this already has its own `debounce` implementation to cater to this need.
- No tests are provided for the new functionality.

Closing for now, but could in theory be reopened.

PR #2520, opened 2016-04-28
_.sortBy, min, max errors on symbols

### LINKED ISSUE #2501: _.sortBy errors on symbols.
``` js
_.sortBy([1, Symbol()]);
// errors
```

This affects `_.max` and `_.min` too.

xcodebuild:
fix #2501, now

``` js
_.sortBy([4, Symbol()], 3, 2);       // [4, Symbol(), 2, 3], just as sort a array contain object
_.max([Symbol(), 1, 4, 6]);          // 1
_.min([Symbol(), 1, 4, 6]);           // 6
```


jashkenas:
Out of honest curiosity — what exactly are Symbols in JavaScript good for? 

Do they do anything that can't be better accomplished with strings?

Is their *only* value that you can get names that are guaranteed not to conflict? When is that useful in real world code?

jgonggrijp:
Closing for the same reason as #2501.

@jashkenas `Symbol` probably has some kind of purpose, but in any case, it is not sortable with operator `<`.

PR #2521, opened 2016-04-29
Use the irregular id when jumping to `functions`

captbaritone:
Because the "Functions" section heading already uses `#functions` the
individual method uses an irregular id. I missed this in #2514.


captbaritone:
I'll cherry-pick to `gh-pages`.

captbaritone:
Cherry-pick complete.

PR #2523, opened 2016-05-02
Add support for maps in _.pairs method ...Closes #2517

### LINKED ISSUE #2517: _.pairs and maps
I was working with maps and reached for a `_.pairs` method to get its key-value pairs and thought it would be handy to support.

``` js
var m = new Map([['a', 1], ['b', 2]]);
_.pairs(m);
// => [['a', 1], ['b', 2]]
```

penDerGraft:
_add support for maps within _.pairs method by returning array of map entries._

I decided to go with a different approach. Returning an array of `obj.entries()` seemed a little bit cleaner than iterating over each entry, and seemed to follow the pattern of type-check -> return in the `_.keys` method. Happy to use a loop though if it is deemed the better option.

### DIFF
--- a/test/objects.js
+++ b/test/objects.js
@@ -82,7 +82,10 @@
     assert.deepEqual(_.pairs({one: 1, two: 2}), [['one', 1], ['two', 2]], 'can convert an object into pairs');
     assert.deepEqual(_.pairs({one: 1, two: 2, length: 3}), [['one', 1], ['two', 2], ['length', 3]], '... even when one of them is "length"');
     if (typeof Map === 'function') {
-      assert.deepEqual(_.pairs(new Map([['one', 1], ['two', 2]])), [['one', 1], ['two', 2]], '... or when object is a Map');
+      var mapObj = new Map();
+      mapObj.set('one', 1);
+      mapObj.set('two', 2);
+      assert.deepEqual(_.pairs(mapObj), [['one', 1], ['two', 2]], '... or when object is a Map');
     }
   });
 
--- a/test/objects.js
+++ b/test/objects.js
@@ -81,6 +81,9 @@
   QUnit.test('pairs', function(assert) {
     assert.deepEqual(_.pairs({one: 1, two: 2}), [['one', 1], ['two', 2]], 'can convert an object into pairs');
     assert.deepEqual(_.pairs({one: 1, two: 2, length: 3}), [['one', 1], ['two', 2], ['length', 3]], '... even when one of them is "length"');
+    if (typeof Map === 'function') {
+      assert.deepEqual(_.pairs(new Map([['one', 1], ['two', 2]])), [['one', 1], ['two', 2]], '... or when object is a Map');
+    }
   });
 
   QUnit.test('invert', function(assert) {
--- a/underscore.js
+++ b/underscore.js
@@ -435,6 +435,15 @@
     if (_.has(result, key)) result[key]++; else result[key] = 1;
   });
 
+  var mapToArray = function(map) {
+    var index = -1;
+    var result = Array(map.size);
+    map.forEach(function(value, key) {
+      result[++index] = [key, value];
+    });
+    return result;
+  };
+
   var reStrSymbol = /[^\ud800-\udfff]|[\ud800-\udbff][\udc00-\udfff]|[\ud800-\udfff]/g;
   // Safely create a real, live array from anything iterable.
   _.toArray = function(obj) {
@@ -1009,21 +1018,12 @@
 
   // Convert an object into a list of `[key, value]` pairs.
   _.pairs = function(obj) {
-    var pairs, i, length;
-    if (_.isMap(obj)) {
-      var entries = obj.entries();
-      length = obj.size;
-      pairs = Array(length);
-      for (i = 0; i < length; i++) {
-        pairs[i] = entries.next().value;
-      }
-    } else {
-      var keys = _.keys(obj);
-      length = keys.length;
-      pairs = Array(length);
-      for (i = 0; i < length; i++) {
-        pairs[i] = [keys[i], obj[keys[i]]];
-      }
+    if (_.isMap(obj)) return mapToArray(obj);
+    var keys = _.keys(obj);
+    var length = keys.length;
+    var pairs = Array(length);
+    for (var i = 0; i < length; i++) {
+      pairs[i] = [keys[i], obj[keys[i]]];
     }
     return pairs;
   };
--- a/underscore.js
+++ b/underscore.js
@@ -1009,12 +1009,21 @@
 
   // Convert an object into a list of `[key, value]` pairs.
   _.pairs = function(obj) {
-    if (_.isMap(obj)) return Array.from(obj.entries());
-    var keys = _.keys(obj);
-    var length = keys.length;
-    var pairs = Array(length);
-    for (var i = 0; i < length; i++) {
-      pairs[i] = [keys[i], obj[keys[i]]];
+    var pairs, i, length;
+    if (_.isMap(obj)) {
+      var entries = obj.entries();
+      length = obj.size;
+      pairs = Array(length);
+      for (i = 0; i < length; i++) {
+        pairs[i] = entries.next().value;
+      }
+    } else {
+      var keys = _.keys(obj);
+      length = keys.length;
+      pairs = Array(length);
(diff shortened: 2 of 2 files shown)


penDerGraft:
Looking into failed tests...

akre54:
Should this support Sets too?

penDerGraft:
What would be the desired behavior for `_.pairs` of Sets? For Maps/Objects it is key, value, but Sets don't have quite the same structure. What should be returned?

akre54:
I'm not entirely sure. It could be the same entries call, with the output
just passing that along (2 arrays of arrays, if I'm not mistaken)

On Mon, May 2, 2016, 11:44 Brent Pendergraft <notifications@github.com>
wrote:

> What would be the desired behavior for _.pairs of Sets? For Maps/Objects
> it is key, value, but Sets don't have quite the same structure. What should
> be returned?
>
> —
> You are receiving this because you commented.
> <https://github.com/jashkenas/underscore/pull/2523#issuecomment-216272557>
>
-- 

Adam K (mobile)

penDerGraft:
Using the logic in the last commit, if a check for `_.isSet` is added as well, it will return an array of arrays which contain a two of each entry in the Set.

```javascript
var s = new Set([1,2,3]);
_.pairs(s);
// => [[1,1], [2,2], [3,3]]
```

@akre54 If that's what you're envisioning, I can add the check.

akre54:
Yeah I think it should be treated like an object with `{[key]: key}` (a la [keymirror](https://www.npmjs.com/package/keymirror))

penDerGraft:
Ok, as part of this PR or a new one? Seems like it could be a separate issue.

underscore.js
```diff
@@ -1009,11 +1009,21 @@
...
   // Convert an object into a list of `[key, value]` pairs.
   _.pairs = function(obj) {
-    var keys = _.keys(obj);
-    var length = keys.length;
-    var pairs = Array(length);
-    for (var i = 0; i < length; i++) {
-      pairs[i] = [keys[i], obj[keys[i]]];
+    var pairs, i, length;
+    if (_.isMap(obj)) {
+      var entries = obj.entries();
```
akre54: Why'd you remove the `Array.from` here? Seems like if you had `Map` you'd also have `Array.from`.

akre54: I'm really not a fan of this much branching logic for Map (which I think you probably wouldn't even use Underscore with in the first place).

akre54:
Separate pull is fine, interested to hear the other maintainers thoughts on it too.

underscore.js
penDerGraft: Yeah, I agree that `Array.from` is a cleaner solution. Unfortunately it isn't supported in node 0.12.*, hence the failed tests.  
```javascript
> typeof Map
'function'
> Array.from
undefined
```
Not sure on the question of whether or not to even use underscore with Maps. I saw [#2517](https://github.com/jashkenas/underscore/issues/2517#issuecomment-216385065) and decided to submit a PR.

akre54: What's the first version it was supported in? We could also drop support
for node 0.12.x since it's pretty ancient by this point

On Mon, May 2, 2016, 18:44 Brent Pendergraft <notifications@github.com>
wrote:

> In underscore.js
> <https://github.com/jashkenas/underscore/pull/2523#discussion_r61817716>:
>
> > @@ -1009,11 +1009,21 @@
> >
> >    // Convert an object into a list of `[key, value]` pairs.
> >    _.pairs = function(obj) {
> > -    var keys = _.keys(obj);
> > -    var length = keys.length;
> > -    var pairs = Array(length);
> > -    for (var i = 0; i < length; i++) {
> > -      pairs[i] = [keys[i], obj[keys[i]]];
> > +    var pairs, i, length;
> > +    if (_.isMap(obj)) {
> > +      var entries = obj.entries();
>
> Yeah, I agree that Array.from is a cleaner solution. Unfortunately it
> isn't supported in node 0.12.*, hence the failed tests.
>
> > typeof Map'function'> Array.fromundefined
>
> Not sure on the question of whether or not to even use underscore with
> Maps. I saw #2517
> <https://github.com/jashkenas/underscore/issues/2517#issuecomment-216385065>
> and decided to submit a PR.
>
> —
> <https://github.com/jashkenas/underscore/pull/2523/files/742033f3de979e46a99d3e9553088e5897d49952#r61817716>
>
-- 

Adam K (mobile)

penDerGraft: Looks like 4.0.0, which I think was the next official node.js release. Everything between the two was `io.js`.

akre54: Ah right. Ok we may have to hold off a bit more.

Let's be honest though. How useful is this change actually? Are there that
many folks who wouldn't just use Array.from(map.entries())?

On Mon, May 2, 2016, 19:01 Brent Pendergraft <notifications@github.com>
wrote:

> In underscore.js
> <https://github.com/jashkenas/underscore/pull/2523#discussion_r61819425>:
>
> > @@ -1009,11 +1009,21 @@
> >
> >    // Convert an object into a list of `[key, value]` pairs.
> >    _.pairs = function(obj) {
> > -    var keys = _.keys(obj);
> > -    var length = keys.length;
> > -    var pairs = Array(length);
> > -    for (var i = 0; i < length; i++) {
> > -      pairs[i] = [keys[i], obj[keys[i]]];
> > +    var pairs, i, length;
> > +    if (_.isMap(obj)) {
> > +      var entries = obj.entries();
>
> Looks like 4.0.0, which I think was the next official node.js release.
> Everything between the two was io.js.
>
> —
> <https://github.com/jashkenas/underscore/pull/2523/files/742033f3de979e46a99d3e9553088e5897d49952#r61819425>
>
-- 

Adam K (mobile)

penDerGraft: I think that's fair. I wondered that as I was coding it actually. Although, I can see value in not having to think about the details, i.e. whether it's a map or an object (which are similar enough data structures that Mozilla gives examples on when to use each). Still, I'm fine if you want to close it. Might want to have @jdalton weigh in though given that he submitted the issue in the first place.

jdalton: You can use `Map#forEach` for all environments that support `Map`. You can then recycle the same code for `_.toPairs` to enable `_.toArray` to convert maps to entries too. I've found lib methods like this useful to support various Node versions without worrying about shims or what narrow support paths are available.

penDerGraft: Thanks for the tip. I had thought about that, but wasn't sure if it would cause confusion since `Array#forEach` is handled by `_.each`. So, `_.toArray` would handle Maps and `_.pairs` would use `_.toArray`?

jdalton: See [npm/toArray](https://github.com/lodash/lodash/blob/4.11.2-npm/toArray.js#L42-L56) for an example of using a helper function like [mapToArray](https://github.com/lodash/lodash/blob/4.11.2-npm/_mapToArray.js).

penDerGraft: Sorry, just want to make sure I'm on the same page. So are we looking for a separate helper function here, or just the same `_.pairs` function being used in other places?

jdalton: A helper function to implement both. I'm showing that the code cost is spread around because it can be used for more than instance.

penDerGraft: Ah, got it. Thank you.

test/objects.js
```diff
@@ -82,7 +82,10 @@
     assert.deepEqual(_.pairs({one: 1, two: 2}), [['one', 1], ['two', 2]], 'can convert an object into pairs');
     assert.deepEqual(_.pairs({one: 1, two: 2, length: 3}), [['one', 1], ['two', 2], ['length', 3]], '... even when one of them is "length"');
     if (typeof Map === 'function') {
-      assert.deepEqual(_.pairs(new Map([['one', 1], ['two', 2]])), [['one', 1], ['two', 2]], '... or when object is a Map');
```
penDerGraft: I read in a previous PR that IE11 does not support initializing Maps with arrays in the constructor, hence the refactor here.

underscore.js
```diff
@@ -435,6 +435,15 @@
     if (_.has(result, key)) result[key]++; else result[key] = 1;
   });
```
penDerGraft: For this PR I just made the change in `_.pairs`, but the helper function allows quick checks in other underscore functions as well, as @jdalton mentioned (i.e. `_.toArray`).

jgonggrijp:
Closing for the same reason as #2808. See also #2147.

PR #2525, opened 2016-05-02
combine _.every and _.some

ahonn:
combine some duplicate code in _.every and _.some


Florian-R:
See #1637, #2383 and some others. This has been proposed now and then, but was dismissed for perf implications.

PR #2527, opened 2016-05-04
fix #2497, make _.memoize work with symbol

### LINKED ISSUE #2497: _.memoize hasher result cannot be a symbol.
Because the `_.memoize` hasher is resolved with `'' +` it will throw an error if the hasher result is a Symbol. A similar issue exists in `createEscaper`.

xcodebuild:


CodeFalling:
Faild because a nodejs 0.12 bug, fixed in https://github.com/jashkenas/underscore/pull/2528

test/functions.js
```diff
@@ -173,6 +173,13 @@
     assert.notStrictEqual(myObj, void 0, 'object is created if second argument used as key');
     assert.strictEqual(myObj, myObjAlias, 'object is cached if second argument used as key');
     assert.strictEqual(myObj.value, 'a', 'object is not modified if second argument used as key');
+
+    // test symbol
+    if (typeof Symbol !== 'undefined') {
+      var itSelf = _.memoize(function(self){return self;});
+      var sym = Symbol();
+      assert.strictEqual(sym, itSelf(sym), 'symbol can be memoized, too');
+    }
```
jgonggrijp, line 182: This is not enough. Firstly, it invokes the function only once, so the lookup behavior is never triggered. Secondly, it doesn't check that distinct symbols can actually be distinguished from each other.

underscore.js
```diff
@@ -782,7 +782,7 @@
   _.memoize = function(func, hasher) {
     var memoize = function(key) {
       var cache = memoize.cache;
-      var address = '' + (hasher ? hasher.apply(this, arguments) : key);
+      var address = _(hasher ? hasher.apply(this, arguments) : key).toString();
```
jgonggrijp, line 785: This isn't going to solve the problem:

```js
Symbol() === Symbol()
// false
String(Symbol())
// 'Symbol()'
String(Symbol()) === String(Symbol())
// true
```

Issue #2455, opened 2016-03-01
isSafeInteger


jdalton:
Is this something like [isSafeInteger](http://www.ecma-international.org/ecma-262/6.0/#sec-number.issafeinteger) something Underscore should tackle too?

I went down the rabbit hole a bit with the number methods and ended up with `isSafeInteger`, `isInteger`, `isLength`, `toInteger`, `toSafeInteger`, `toLength`, & `toNumber`.

devanp92:
@jdalton were you proposing that isInteger, isLength, toInteger, toSafeInteger, toLength, & toNumber were also to be supported?

jdalton:
I'm ok with it.

jgonggrijp:
Personally, I think this is the kind of thing that would be more suited for an extension library like Underscore-Contrib.

PR #2529, opened 2016-05-05
Add support for isSafeInteger

devanp92:
Include [isSafeInteger](http://www.ecma-international.org/ecma-262/6.0/#sec-number.issafeinteger) as referenced in [issue 2455](https://github.com/jashkenas/underscore/issues/2455) by @jdalton.

The conditions are:
1. If Type(number) is not Number, return false.
2. If number is NaN, +∞, or −∞, return false.
3. Let integer be ToInteger(number).
4. If integer is not equal to number, return false.
5. Otherwise, return true.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -1344,6 +1344,10 @@
     return obj != null && hasOwnProperty.call(obj, key);
   };
 
+  _.isSafeInteger = function(obj) {
+    return _.isNumber(obj) && _.isFinite(obj) && toString.call(parseInt(obj, 10)) === '[object Number]' && Math.floor(obj) === obj;
+  };
+
   // Utility Functions
   // -----------------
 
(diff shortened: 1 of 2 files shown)


underscore.js
```diff
@@ -1344,6 +1344,10 @@
     return obj != null && hasOwnProperty.call(obj, key);
   };
 
+  _.isSafeInteger = function(obj) {
+    return _.isNumber(obj) && _.isFinite(obj) && toString.call(parseInt(obj, 10)) === '[object Number]' && Math.floor(obj) === obj;
```
jgonggrijp, line 1348: Is the third condition really necessary here? It seems kind of redundant with the first. Also, maybe the first is a bit too lenient. The standard seems to suggest that object-wrapped numbers are not allowed, so it should probably be `typeof obj === 'number'`.

jgonggrijp:
Postponing decision on this. As I wrote in #2455, I think this rather belongs in an extension library.

Issue #2286, opened 2015-08-25
Underscore could use a _.propertyResult


machineghost:
The new `_.property` is really neat, and I love using it, but there's one problem: it only returns static values. 

For instance, let's say I want to find out if any object in an array contains has an `isNew` property; I can do the following:

    var objects = [{isNew: true}, {isNew: false}];
    _(objects).any(_('isNew').property()); // === true

That's a lot better than before `property`, when I had to do:

    _(objects).any(function(x) { return x.isNew; });

But what if instead of random objects I have a bunch of `Backbone.Model` objects.  Those don't have an `isNew` property, they have an `isNew` *method*.  So:

    _(objects).any(_('isNew').property());

won't work; instead I'm back to the old style:

    _(objects).any(function(x) { return _(x).result('isNew'); });

It would be really great if there was a method that was just like `_.property`, except that if it would return a function, it returns the result of executing that function instead (like `_.result`).  In other words, if I could do ...

    _(objects).any(_('isNew').propertyResult());

Since it's a fairly common practice to have `is*` methods on objects, and also common practice to want to use Underscore methods like `any` or `filter` with those methods, I really think this would be a useful addition to the library.

jridgewell:
Lodash [added](https://github.com/lodash/lodash/pull/1091) `_.method` and `_.methodOf` for this.

carpben:
I wonder if this feature is still relvant, and is worthy of development in Javascript's current eco-system. 

Since this suggestion was originally published, arrow function became a popular and wide spread feature in the Javascript ecosystem. 

At the time @machin had to write: 
```js
_(objects).any(function(x) { return x.isNew; });
```
But today we can simply write: 
```js
_(objects).any( (x)=>x.isNew );
```

@machin wanted to be able to write:
```js
_(objects).any(_('isNew').propertyResult());
```
But today we can write: 
```js
_(objects).any( (x)=>x.isNew() );
```
Personally I find the arrow function version much more readable.

jgonggrijp:
@carpben I think you are right that with modern syntax, `_.propertyResult` wouldn't be a game changer. At least not as an iteratee; there may be other convincing use cases that we haven't considered. But as far as I know, nobody has reported one yet.

[...]

PR #2531, opened 2016-05-05
[#2286]: Underscore could use a _.propertyResult

monkpow:
I chose the name _.method per the above suggestion, but am happy to adjust if this does not match the project guidelines.


underscore.js
```diff
@@ -1370,6 +1370,24 @@
 
   _.property = property;
 
+  // Generates a function for a given object that returns the passed function's return value.
+  // Accepts a string or function literal for value.
+  // If value is a string, function assumes (and verifies) this string is a method name on the
+  // referenced object.
+  _.method = function(value) {
+    var isFunction = this.isFunction,
+        rest = this.rest,
```
megawac: use `restArgs` instead

underscore.js
```diff
@@ -1370,6 +1370,24 @@
 
   _.property = property;
 
+  // Generates a function for a given object that returns the passed function's return value.
+  // Accepts a string or function literal for value.
+  // If value is a string, function assumes (and verifies) this string is a method name on the
+  // referenced object.
+  _.method = function(value) {
+    var isFunction = this.isFunction,
```
megawac: don't use `this` to reference underscore functions. Use `_`

underscore.js
```diff
@@ -1370,6 +1370,24 @@
...
+        rest = this.rest,
+        func;
+
+    return function(obj) {
+      if (obj == null) { return; }
```
megawac: is this not valid when a function is provided?

underscore.js
```diff
@@ -1370,6 +1370,24 @@
...
+      func = isFunction(value) ? value : obj[value];
```
megawac: could check if `value` is a function outside of this function

monkpow:
@megawac Thanks for the quality feedback.  Updated diff addresses:
- use restArgs
- don't use `this` keyword, use direct reference
- Clarified comment and token in response to Question: "is this not valid when a function is provided?".  Added narrative above in this thread.
- check if value is a function in the outer function.

underscore.js
monkpow: This method responds to both a string literal (a method name) or a function literal.   This check is to sort out if the value passed is a function literal, so that later we know what to call apply on.

underscore.js
```diff
@@ -1370,6 +1370,24 @@
...
+  // Accepts a method name or function literal for value.
+  // If value is a string, function assumes this string is a method name on the
+  // referenced object.
+  _.method = function(value) {
+    var isFunctionLiteral = _.isFunction(value),
+        func;
+
+    return restArgs(function(obj, args) {
+      if (obj == null) { return; }
+      func = isFunctionLiteral ? value : obj[value];
```
megawac: define `func` here rather than outer scope

underscore.js
```diff
@@ -1370,6 +1370,24 @@
...
+  // Generates a function for a given object that returns the passed function's return value.
+  // Accepts a method name or function literal for value.
+  // If value is a string, function assumes this string is a method name on the
+  // referenced object.
+  _.method = function(value) {
+    var isFunctionLiteral = _.isFunction(value),
+        func;
+
+    return restArgs(function(obj, args) {
+      if (obj == null) { return; }
```
megawac: any rational why should func not be called if it is a function literal and obj is null?

jdalton:
fyi [lodash/method](https://github.com/lodash/lodash/blob/4.11.2-npm/method.js).

underscore.js
monkpow: My current rational is that this invocation pattern doesn't make much sense to me.  e.g.,

```
_().method(function() { /* function body *. })
_.method(null, function () { /* function body */ })
```

since both would be more easily written as a plain function literal, and this invocation pattern reads to me as 'call the provided method on undefined'.

That said, if allowing an undefined object in this position is more in keeping with underscore's style, I'm happy to update so we can call _.method on undefined objects.

underscore.js
monkpow: Thanks, updated diff with this change.

PR #2532, opened 2016-05-05
Overload "values" method so that pick arguments can be passed in to l…

### LINKED ISSUE #2445: "pickValues" Overload Syntax for Values
Currently Underscore has a `pick` method which returns a specific set of keys/values (in object form), and it has a `values` method which returns all values.  However, there is no overlapping function.  In other words, there's no way to select a specific set of values from an object without combining the two functions:

```
_.chain({foo: 1, bar: 2, baz: 3}).pick('foo', 'baz').values().value(); // returns [1, 3]
```

Wouldn't it be great if you could do all of that with a single Underscore function?  And better yet, what if you could do it without even adding any new functions to the library?

In other words, wouldn't it be great if you could do:

```
_.values({foo: 1, bar: 2, baz:3}, 'foo', 'bar'); // returns [1, 3]
```

Of course a new `pickValues` method would work also, but this overload of `values` seems very natural; even with it `values` still returns the values of an object, it just returns a more limited sub-set.  And because `values` has only ever taken one argument, it would be perfectly safe to add support for another argument, to enable this new functionality.

Just a suggestion.

monkpow:
…imit the response.  Closes #2445.

### DIFF
--- a/test/objects.js
+++ b/test/objects.js
@@ -76,6 +76,11 @@
   QUnit.test('values', function(assert) {
     assert.deepEqual(_.values({one: 1, two: 2}), [1, 2], 'can extract the values from an object');
     assert.deepEqual(_.values({one: 1, two: 2, length: 3}), [1, 2, 3], '... even when one of them is "length"');
+    assert.deepEqual(_.values({foo: 1, bar: 2, baz: 3}, 'foo', 'baz'), [1, 3], 'limits results when passed a list of "pick" keys');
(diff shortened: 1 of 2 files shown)


jgonggrijp:
- I believe that a correct implementation would be consistent with `_.pick` and ignore nonexistant keys instead of defaulting to `void 0`.
- I believe that functions that are just trivial compositions of two other Underscore functions (in this case `_.pick` and the "bare" `_.values`) shouldn't be added to Underscore. Let's keep Underscore a lean, highly orthogonal library where the amount of functionality offered is a high power of the number of predefined functions.

Sorry for the long silence and the disappointing end, @monkpow.

Issue #2453, opened 2016-02-29
Change to SameValueZero for comparisons


jdalton:
I thought I brought this up before but I can't find the issue. Currently for comparisons in methods like `indexOf`, `lastIndexOf`, `includes`, `uniq`, `union`, `difference`, `intersection` a strict equality, `===` is used. If this was tweaked a bit to [SameValueZero](http://ecma-international.org/ecma-262/6.0/#sec-samevaluezero), `===` + `NaN` matches,  then it would allow for optimizations with `Set` and `Map` as well as cover the `NaN` match case.

Thoughts?

jdalton:
It would mean:
```js
_.indexOf(['a', NaN, 'b'], NaN);
// => 1 instead of -1
```

PR #2537, opened 2016-05-13
Optimize "uniq" to be O(n) instead of O(n^2)

Tzook:
The uniq function is a heavy function that could cause issues on big arrays.
This commit utilizes the Map object if supported (95% browser support according to caniuse), to make the operation much faster.
All unit tests pass.
Running a simple example in Chrome on an array with 10,000 unique elements:
  * Before - takes about 8 seconds
  * After - takes about 0.03 seconds
Thanks.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -535,6 +535,7 @@
   // been sorted, you have the option of using a faster algorithm.
   // Aliased as `unique`.
   _.uniq = _.unique = function(array, isSorted, iteratee, context) {
+    var hasMapSupported = typeof Map !== 'undefined';
     if (!_.isBoolean(isSorted)) {
       context = iteratee;
       iteratee = isSorted;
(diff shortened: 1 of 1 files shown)


jdalton:
Keep in mind this toggles the matching behavior from [strict equality](http://www.ecma-international.org/ecma-262/6.0/#sec-strict-equality-comparison) to [same value zero](http://www.ecma-international.org/ecma-262/6.0/#sec-samevaluezero) depending on environment support.

Tzook:
Good point @jdalton.
The only differences according to [this table](https://developer.mozilla.org/en/docs/Web/JavaScript/Equality_comparisons_and_sameness) are that -0 is not equal to +0, and that NaN equals to NaN.
In my opinion it makes even more sense for the function to use same-value-zero comparison. Having an array with multiple NaNs should result in only one NaN after using unique.

jdalton:
Yep.

forty:
Hello everyone, 
is this PR ever going to be merged, or does it need additional work?

Thanks

jgonggrijp:
- It would have to consistently use SameValueZero, even in environments that don't have `Map`.
- There would have to be a new major release of Underscore in which we actually decided to use SameValueZero instead of strict equality *everywhere*. See #2453. It is not yet decided whether we will ever do that.
- It shouldn't do `if`/`else` all over the place. Move the `supportsMap` out of the function and switch over its value only once instead of repeatedly in every loop iteration.

PR #2539, opened 2016-05-25
Add linspace() to array functions.

codypiersall:
This PR adds `linspace()` to the array functions.  It also adds tests and documentation for the function.

`linspace()` is a useful function for generating _n_ evenly-spaced numbers between two values.  It's pretty closely related to `range()`, but with range you're usually more interested in doing stuff with integers.  linspace() is especially useful when you want to generate values for plotting; if I want to plot cos(x) from 0 to 2pi, I can just do 

```
x = _.linspace(0, 2*Math.PI);
y = x.map(Math.cos);
// However you want to plot it; e.g.
Plotly.plot('myDiv', [{x: x, y: y}]);
```

`linspace()` gives a nice way of approaching this use case; and if it turns out you need more or fewer points, you just have to tweak the third parameter of linspace.

### Prior Art

[NumPy](http://docs.scipy.org/doc/numpy-1.10.0/reference/generated/numpy.linspace.html) and [MATLAB](http://www.mathworks.com/help/matlab/ref/linspace.html) both have this function.  Note that they differ on the default value for number of points; I opted to go with 50, like NumPy, but it was an arbitrary decision that I made because I like Python more than MATLAB.

P.S. I added documentation because I read CONTRIBUTING.md after I already added it, but I will rebase this PR to not include it if that works better for you.  I based the docs I added off of the _.range() function.

### DIFF
--- a/index.html
+++ b/index.html
@@ -257,6 +257,7 @@
         <li data-name="findIndex">- <a href="#findIndex">findIndex</a></li>
         <li data-name="findLastIndex">- <a href="#findLastIndex">findLastIndex</a></li>
         <li data-name="range">- <a href="#range">range</a></li>
+        <li data-name="linspace">- <a href="#linspace">linspace</a></li>
       </ul>
     </div>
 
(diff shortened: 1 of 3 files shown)


captbaritone:
Hey Cody!

Thanks for your effort. Looks like this would be quite useful, as you said, for plotting and other types of data work. That being said, I don't believe this has a broad enough use-case for inclusion in Underscore proper. Having it as a pull request will likely help anyone else who is looking to implement it as a mixin. If there is more interest in the future we can revisit, but for now I'm going to close it.

Thanks for your effort, I'm sorry we aren't able to merge it at this time.

jridgewell:
This could just be written with `_.range`, right?

```js
_.range(0, Math.PI, ((Math.PI - 0) / 50));
```

codypiersall:
@captbaritone Thanks for the feedback!  No sweat on not being able to merge it.  Thanks for the feedback!

@jridgewell This gives you a different way of thinking about getting a sequence of numbers than `_.range`.  You could definitely implement linspace using range, but you would normally use `linspace` when you know the number of points you want, and `range` when you know the spacing.  Of course, if you know either the spacing or the number of points you can calculate the other, but it just gives different ways of approaching the problem.  Also, `linspace` includes the endpoints, where range doesn't; however, if you use non-integer spacing for `range`, sometimes you get one more number than you might expect due to floating point weirdness. That never happens with linspace.  

An Example of an extra point using range: `_.range(0.5, 0.55, 0.01)` yields `[0.5, 0.51, 0.52, 0.53, 0.54, 0.55]`.  This isn't a problem with the implementation of range, of course, just a problem with all computers everywhere.

PR #2540, opened 2016-05-26
added deep merge functionality

sanket-redkar:
I have given deep merger functionality,
_.deepMerge({a: {b: 1}, c: [1, 2]} {a: {c: 2}, c: [3, 1]})
=> {a: {b: 1 c: 2}, c: [3, 1]}

### DIFF
--- a/test/collections.js
+++ b/test/collections.js
@@ -171,6 +171,10 @@
     assert.strictEqual(_.collect, _.map, 'is an alias for map');
   });
 
+  QUnit.test('deep Merge', function(assert){
+    assert.deepEqual(_.deepMerge({a: {b: 1, c: 3}}, {a: {b: 2, d: 3}}).a, {b: 2, c: 3, d: 3}, 'deep Merge nexted hash.' );
+  });
+
   QUnit.test('reduce', function(assert) {
     var sum = _.reduce([1, 2, 3], function(memo, num){ return memo + num; }, 0);
     assert.equal(sum, 6, 'can sum up an array');
--- a/test/collections.js
+++ b/test/collections.js
@@ -171,10 +171,6 @@
     assert.strictEqual(_.collect, _.map, 'is an alias for map');
   });
 
-  QUnit.test('deep Merge', function(assert){
-    assert.deepEqual(_.deepMerge({a: {b: 1, c: 3}}, {a: {b: 2, d: 3}}).a, {b: 2, c: 3, d: 3}, 'deep Merge nexted hash.' );
-  });
-
   QUnit.test('reduce', function(assert) {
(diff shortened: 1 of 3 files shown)


syagev:
This is very useful. Although I would add it as a flag (or option hash) under the existing `extend` method. Can any maintainer comment about when this is expected to be merged? Thanks

sanket-redkar:
Ok I am working on it.

sanket-redkar:
I have made the changes like this 
_.extend({a: {b: 1}, c: [1, 2]} {a: {c: 2}, c: [3, 1]}, 'deepMarge' )
=> {a: {b: 1 c: 2}, c: [3, 1]} 
_.extend({a: {b: 1}, c: [1, 2]} {a: {c: 2}, c: [3, 1]}, 'deepMarge' )
=> {a: {c: 2}, c: [3, 1]}

syagev:
Cool!

I would follow the notion of adding options as a hashset rather than a string. Take a look at [throttle](http://underscorejs.org/#throttle) for example. So the method signature would be:

    _.extend(destination, *sources, [options])

where `options` can be `{ deep: true/false }` with the default behaviour being `{deep: false}` so that the method is backward compatible. 

Please note that I am **not** a moderator on this repo and do not have push permissions. I am just a fellow user like you making some friendly suggestions so treat it as you see fit ;)

sanket-redkar:
yes I am also think in that way that last one is options, but how would I achieve this with
`_.extend(destination, *sources, [options])`
because when we gave *source then last option  will also consider in source.

syagev:
You're right... I missed that... So probably better to create a new method like you suggested in the beginning: `_.extendDeep(destination, *sources)`.

I guess that's why they have `extendOwn`...

sanket-redkar:
@syagev  My first commit only merge two hash, If we have to go with multiple hash then same issue will come..
Shall we do in bellow structure that 
`_.extend(destination, *sources, {options: {deep: true} })`

and always consider last hash as option when it contain options key and as deep inner key

syagev:
What I meant is that it's probably better to go with your original suggestion of a separate method for this functionality (e.g `extendDeep`). That way it can have an identical signature to `extend`. I think the existence of an `extendOwn` method is good justification of this approach.

jridgewell:
Thanks for the PR work, but this idea's been rejected [many times before](https://github.com/jashkenas/underscore/pull/1585#issuecomment-40612083).

jdalton:
@sanket-redkar @syagev  See [lodash/merge](https://github.com/lodash/lodash/blob/4.13.1-npm/merge.js) or [lodash.merge](https://www.npmjs.com/package/lodash.merge).

sanket-redkar:
Thanks @jdalton its useful.....

PR #2545, opened 2016-06-15
Update collections.js

VimalChaudhary:


megawac:
No thanks, same reason as the other.

PR #2546, opened 2016-06-15
Update collections.js

VimalChaudhary:


megawac:
Hi, thanks for the submission but this is a test file. I don't think its worthwhile to include documentation of the underscore methods here (the docs are present in underscore.js and index.html)

captbaritone:
@VimalChaudhary If you'd like to contribute documentation to our tests, the most useful way is to improve the test descriptions (in this case `'every'`) or improve the assertion message in some of our assertions.

Let me know if you need to me clarify what I'm suggesting.

PR #2553, opened 2016-06-25
matcher move to propertyOf in index.html

ikasumiwt:
I propose that functions are listed as the same order.

I move matcher under propertyOf in sidebar,Because container is “property/propertyOf/matcher” order.


jashkenas:
This @coveralls thing is pretty obnoxious, isn't it?

michaelficarra:
Yep.

jdalton:
You can always keep coverage and disable the in PR notifications.

PR #2555, opened 2016-06-25
fix (eslint) warning "index" used outside of binding context

tnga:


underscore.js
```diff
@@ -109,9 +109,10 @@
   var restArgs = function(func, startIndex) {
     startIndex = startIndex == null ? func.length - 1 : +startIndex;
     return function() {
-      var length = Math.max(arguments.length - startIndex, 0);
-      var rest = Array(length);
-      for (var index = 0; index < length; index++) {
+      var length = Math.max(arguments.length - startIndex, 0),
+          rest = Array(length),
+          index = 0;
+      for (index = 0; index < length; index++) {
```
captbaritone: I believe you should be able to remove this initialization:

tnga:
yeah! I did it first time, but was thinking that it had no "style". but now, considering you suggest it, maybe i can reconsider it !

underscore.js
tnga: yeah! I did it first time, but was thinking that it had no "style". but now, considering you suggest it, maybe i can reconsider it !

PR #2556, opened 2016-06-25
improve `isElement` for precision of DOM element's type

tnga:
A precision can be now done with a given element type's name (egg: *HTMLLink, HTMLDiv, SVGEllipse, ...*).

*examples* 
```js
_.isElement(jQuery('body')[0]);
//=> true
_.isElement(document.createElement('a'), 'html-anchor');
//=> true
// given type precision can also be in camelCase following DOM naming's spec.
_.isElement(document.body, 'HTMLScript');
//=> false
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -1658,11 +1658,17 @@ _.isEmpty({});
       <p id="isElement">
         <b class="header">isElement</b><code>_.isElement(object)</code>
         <br />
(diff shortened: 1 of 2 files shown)


akre54:
This is too complicated for Underscore's purposes. Would make a great standalone library though.

PR #2573, opened 2016-07-30
[Documentation] Swap findWhere and where in documentation

### LINKED ISSUE #1811: Documentation: confusing find/filter/where/findWhere link ordering/naming
Hi,

I would like to propose reordering of these functions in the [documentation](http://underscorejs.org/#find). The thing is, one often needs to go back to the docs page to figure out which one of these 4 you need since their function names and their ordering in the documentation are a bit inconsistent. Right now they are in the following order:

 * find (works with context, finds the first occurrence)
 * filter (works with context, finds all that pass the truth test)
 * where (works with properties, finds all that pass the match criteria)
 * findWhere (works with properties, finds first that passes the match criteria)

And the logic is a bit off. I think it would be a little less confusing if the order was either:

 * find (context, first)
 * filter (context, array)
 * findWhere (properties, first)
 * where (properties, array)

Or:

 * find (context, first)
 * findWhere (properties, first)
 * filter (context, array)
 * where (properties, array)

Also if you take a look at the names of these functions (even with the current aliases), you can notice that there is no intuitive logic which would let you know right away, which one of these is the one you need at a certain moment. If you add the strange ordering from the documentation to that, you get to the point of going to the docs page every time to check. Perhaps if `where` was called `filterWhere` by default and aliased with `where` (or some other naming improvement along that lines) it would feel a little more consistent? Or if at least ordering on the docs page could be changed? :)

Anyway, just a simple suggestion on improving the ease of use.
Thanks for this amazing library, it saves a loooot of time. Awesome piece work.

Cheers

jdelbello:
I re-ordered findWhere and where in the documentation to make it less confusing

Fixes #1811

### DIFF
--- a/index.html
+++ b/index.html
@@ -582,18 +582,6 @@ var evens = _.filter([1, 2, 3, 4, 5, 6], function(num){ return num % 2 == 0; });
 =&gt; [2, 4, 6]
 </pre>
 
-      <p id="where">
-        <b class="header">where</b><code>_.where(list, properties)</code>
-        <br />
-        Looks through each value in the <b>list</b>, returning an array of all
-        the values that contain all of the key-value pairs listed in <b>properties</b>.
-      </p>
-      <pre>
-_.where(listOfPlays, {author: "Shakespeare", year: 1611});
-=&gt; [{title: "Cymbeline", author: "Shakespeare", year: 1611},
-    {title: "The Tempest", author: "Shakespeare", year: 1611}]
-</pre>
-
       <p id="findWhere">
         <b class="header">findWhere</b><code>_.findWhere(list, properties)</code>
         <br />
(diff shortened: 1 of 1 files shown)


megawac:
Can you also move `find` to be beside `findWhere`?

jdelbello:
@pootzko from #1811 suggested that the order should be as I presented it in this pull request, and I agree. The reasoning is that we want to keep `find` and `filter` together because both search for matches based on context. We also want to keep `where` and `findWhere` together because both search for matches based on properties. I moved `findWhere` before `where` because I wanted to preserve the pattern in the first pair of "find first occurrence" and "find all occurrences".

1. find (context, first)
2. filter (context, array)
3. findWhere (properties, first)
4. where (properties, array)

Does that make sense?

megawac:
wasn't aware of the previous discussion on this. I'm fine with this rationale

PR #2576, opened 2016-08-03
improved intersection performance

landvibe:
Intersection can be improved by sorted array.
The time complexity
- previous code: `m*n^2`
- improved code: `m*n*logn`

below is my test result on my machine
- two arrays with 100,000 items
- previous code: 16830 ms
- improved code: 93 ms

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -570,35 +570,30 @@
 
   // Produce an array that contains every item shared between all the
   // passed-in arrays.
-  _.intersection = function(array) {
-    if (array == null) return [];
-    var result = [];
-    var sortedList = _.sortBy(array);
+  _.intersection = function() {
     var argsLength = arguments.length;
     var i, length;
-    for (i = 1; i < argsLength; i++) {
+    for (i = 0; i < argsLength; i++) {
       if (arguments[i] == null) return [];
+    }
+    var intersectionList = _.sortBy(arguments[argsLength - 1]);
+    for (i = argsLength - 2; i >= 0; i--) {
       var inputList = arguments[i];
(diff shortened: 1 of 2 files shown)


megawac:
A solution more likely to be accepted would be to implement a side case using [ES2015 Sets](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Set)

```js
let set = new Set(array);

// ...

set = new Set(_.filter(inputList, x => set.has(x)));
```

landvibe:
@megawac 
MDN says "Set objects are collections of values, you can iterate its elements in insertion order"
I think it is not sorted data structure.

megawac:
Yes, `intersection` is supposed to preserve original order. The PR as is is a breaking change.

```js
_.intersection([2, 3,1], [1,2])

```

underscore.js
```diff
@@ -571,16 +571,32 @@
...
+        if (_.indexOf(sortedList, inputList[j], true) >= 0) {
+          tempList.push(inputList[j]);
+        }
+      }
+      sortedList = tempList;
+    }
+    // preserves the order of the first array
+    // and get a duplicate-free array
+    var occurMap = [];
+    for (i = 0, length = array.length; i < length; i++) {
```
landvibe: @megawac 
The PR preserves original order

landvibe:
I fixed a bug
bug: wrong output when there are more than 2 arrays
and I added new test case

jgonggrijp:
It's a nice idea but it only works for sortable values (i.e., strings and numbers). It has to work with unsortable values, too.

Issue #2280, opened 2015-08-20
isNil


jdalton:
I recently added an `_.isNil` method to check for nullish, `== null`, values after repeated dev requests and figured I'd bing it up here.

It's a pretty common check which makes it a good candidate for inclusion in core. Looking at Underscore's code base there are 27 `== null` checks compared to 1 `=== null` checks <br>and 6 `=== void 0` checks.

akre54:
:-1: What's the use case? Just for predicates? Otherwise it's almost universally better to use a language feature than a library function.

jdalton:
> :-1: What's the use case? Just for predicates? Otherwise it's almost universally better to use a language feature than a library function.

Well the same could be said about `_.isNull` or `_.isUndefined`. I don't see you advocating for their removal.

jdalton:
Since `_.isNull` and `_.isUndefined` are already in core and nullish checks appear to be more broadly appealing than them I think `_.isNil` is a good addition.

akre54:
Sure. Let's get rid of em.

jdalton:
3,000+ and growing.

akre54:
Huh?

jdalton:
I don't see removal of `_.isNull` or `_.isUndefined` happening without a major bump and since they're already in core and nullish checks appear to be more broadly appealing than them I think `_.isNil` is a good addition.

bjmiller:
I think that a combined function is more useful than the individual `_.isNull` and `_.isUndefined` functions.  I'm not crazy about the name, though.  "Nil" isn't a language concept that exists in JS, and anyone familiar with Ruby might be tempted to think the some of the behavior of `nil` might somehow be present.

The inverse of this function is already in underscore-contrib, by the name `_.isExisty`, FWIW.  Maybe, `_.inNonExisty`?

lucasfanning:
@bjmiller :+1:

jdalton:
I chose nil because it's similar to the Existential Operator, Goovy's safe-dereference operator (?.), Ruby's nil? function, pythons None and others. See https://en.wikipedia.org/wiki/Null_Object_pattern and search for nil.

JosephClay:
I find that, when using `underscore` or `lodash` as a type checking solution, it's very jarring to encounter `== null`. There's a context switch that has to happen from an is + Something pattern to double-equals-no!-oh-wait-it's-against-null-we're-okay pattern.

Also very handy for use in predicates. I usually mixin `_.exists`.

[...]

Issue #2144, opened 2015-04-08
Equivalent to "val != null"


sateffen:
Hey,

I've run into a "problem", that I've got across multiple projects. In JSHint we prevent the !=, so we need to do !== (script equal).

Because of this, I'm used to creating a mixin, which just is like:

    _.mixin({
        isDefined: function (aValue) {
            return aValue !== null && aValue !== undefined;
        }
    });

I use this mixin in every project with underscore again, so I think it is worth to tell you about this. Maybe some other people know my situation, and it's a nice idea for underscore by default, and I didn't see another request for this.

Additionally, this is a better way then relaying on the 'val != null' behavior.

Any thoughts?

jridgewell:
We've gotten one request so far: #2051.

However, there's nothing wrong with `== null`. There's even a [flag](http://jshint.com/docs/options/#eqnull) for allowing just it in jshint.

PR #2581, opened 2016-08-30
Add  object utility function

jonva:
## Overview

Very(too) often in Javascript, one comes across a situation where a conditional block of code has to be executed, based on whether the value of a variable is not `null`, `undefined` or `NaN` i.e. check for the variable's valid existence. Here, I have used the `_.isNull`, `_.isUndefined` and `_.isNaN` functions to create an abstraction called  `_.is` which checks for this popular composite condition.

### DIFF
--- a/test/objects.js
+++ b/test/objects.js
@@ -864,6 +864,22 @@
     assert.ok(_.isUndefined(void 0), 'undefined is undefined');
   });
 
+  QUnit.test('is', function(assert) {
+    assert.ok(_.is(1), 'positive numbers exist');
+    assert.ok(_.is(-1), 'negative numbers exist');
+    assert.ok(_.is(0), 'number 0 exists');
+    assert.ok(_.is('0'), 'string 0 exists');
+    assert.ok(_.is('foo'), 'strings exist');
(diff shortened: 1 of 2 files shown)


Florian-R:
@akre54 @megawac Could you cut or at least reduce the Coveralls notification? It's getting very noisy for little if no benefits. Thanks.

megawac:
Is there a way to only get notifications on change?

jdalton:
FWIW I just disable PR notifications for coverage and just check it from time to time (in travis runs or on whichever coverage site it is)

akre54:
Coveralls aside, I think we've brought this up before. aside from the extra `NaN` checking, what does this give you over a `obj == null` check? This seems like it'd be better off as a mixin.

jonva:
@akre54 I did see a similar mixin on underscore-contrib, but the thought process behind raising a PR here is that arbitrary as it might be, this is such a common use case on most javascript projects that I think as a utility library, underscore is in fact where this belongs.

Florian-R:
FTR #2144 #2280 and probably a few others. Better suited as a mixin in my opinion.

PR #2582, opened 2016-08-30
Remove no-extra-parens ESLint config rule

nik-john:
## Overview

The `no-extra-parens` ESLint rule was causing the test suite to throw warnings. But the location of the warning did not look like it had any problematic parens. This is likely a bug with ESLint. Till it is resolved, I think it makes sense to **remove this flag** to clean up the test suite logs. This will also not have any effect on the code anyway, as extra parentheses are ignored by most JS Engines

## Screenshots

screen shot 2016-08-30 at 1 07 02 pm

NOTE: The code in question here is [this line](https://github.com/jashkenas/underscore/blob/master/underscore.js#L313)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -40,7 +40,6 @@
     "no-ex-assign": 2,
     "no-extend-native": 2,
     "no-extra-boolean-cast": 2,
-    "no-extra-parens": 1,
     "no-extra-semi": 2,
     "no-fallthrough": 2,
     "no-floating-decimal": 2,


captbaritone:
ESLint is correct in its assertion. The parentheses on these lines are technically redundant as far as the parser is concerned. As an exercise, you could try removing them and running the tests. You'll find they all still pass. However, for human readers, I think the parens are quite useful. 

An ideal solution would be to upgrade ESLint to >= v2.8 and enable the ``nestedBinaryExpressions`](http://eslint.org/docs/rules/no-extra-parens#nestedbinaryexpressions) option, which I believe would permit this type of clarifying parentheses.

PR #2583, opened 2016-08-31
Update _.object and _.pairs docs to show relation

sdwebster:
Long time user, first time contributor. Reading through this well-documented source code, I noticed the link between _.pairs and _.object for the first time. As presented, these 2 related functions are far away from each other (with good reason), so I suggest this highlighting the connection in the annotations. Please let me know if there's a better contribution workflow, if I need to PR first to Underscore-Contrib, etc.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1015,7 +1015,8 @@ _.unzip([["moe", 30, true], ["larry", 40, false], ["curly", 50, false]]);
         <b class="header">object</b><code>_.object(list, [values])</code>
         <br />
         Converts arrays into objects. Pass either a single list of
-        <tt>[key, value]</tt> pairs, or a list of keys, and a list of values.
+        <tt>[key, value]</tt> pairs, or a list of keys, and a list of values. The former
+        usage is invertible through <code>_.pairs</code>.
         If duplicate keys exist, the last value wins.
       </p>
       <pre>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -1015,7 +1015,8 @@ <h2 id="arrays">Array Functions</h2>
         <b class="header">object</b><code>_.object(list, [values])</code>
         <br />
         Converts arrays into objects. Pass either a single list of
-        <tt>[key, value]</tt> pairs, or a list of keys, and a list of values.
+        <tt>[key, value]</tt> pairs, or a list of keys, and a list of values. The former
+        usage is invertible through <code>_.pairs</code>.
```
captbaritone: I'm on-board with this change, but I'm not sure to what "The former usage" refers. Could you reword?

index.html
```diff
@@ -1424,7 +1425,8 @@ <h2 id="objects">Object Functions</h2>
       <p id="pairs">
         <b class="header">pairs</b><code>_.pairs(object)</code>
         <br />
-        Convert an object into a list of <tt>[key, value]</tt> pairs.
+        Convert an object into a list of <tt>[key, value]</tt> pairs. Inverse of first
+        usage of <code>_.object</code>.
```
captbaritone: > Inverse of first usage of <code>_.object</code>.

In order to better fit with the rest of our documentation, could you reword this as a complete sentence?

sdwebster: Hi @captbaritone , thanks for the feedback. Is this clearer? I tried to match the wording of other "opposites" such as in _.unescape and _.unzip.

 Also -- I edited the corresponding comments in underscore.js so that the annotated source will stay consistent w/ the main page. Take a look!

captbaritone:
Thanks!

PR #2587, opened 2016-09-11
Tried to add the ability to generate arithmetic ranges using _.range (i.e. factorials)

threequal:
Hey Jeremy, tried to add the ability to generate non-linear ranges (using functions such as factorials) using ` _.range `.  I am sure there are better ways to get it working, but I think it does the job! 

The sort of thing I was hoping to get working... 

```javascript
_.range( 1, 5, 1, Math.pow, Math, [2] );
// => [ 1, 4, 9, 16 ]

_.range( 4, 10, 2, Math.pow, Math, [4] );
// => [ 256, 1296, 4096 ]

var factorial = function(n) {
  if ( n === 0 || n === 1 ) { return 1; }
  return n * factorial(n - 1);
};
_.range( 1, 6, 1, factorial, null, [2] );
// => [ 1, 2, 6, 24, 120 ]
```

Sorry, but I changed a few things in the test suites! Only minimal changes, but I tried to get rid of the throws syntax error message, and I also had to up the max-params eslint rule to allow for 6 parameters (probably not a great idea - but that allowed for the new params I tried to let _.range allow).

The parameters that I have added...

```javascript
_.range( start, stop, step, iteratee, toBind, providedArgs );
```

- ` start `, ` stop ` and ` step ` are all the same
- ` iteratee ` is a callback for each iteration of the loop
- ` toBind ` is what the user wants ` this ` to be, mainly set up to allow for things like Math.pow
  - I probably could have used ` bind ` but I am using the toBind variable for the ` apply ` method
- ` providedArgs ` expects an array, just so we can pass in static values to the ` apply ` method later on (if that is necessary)

Hopefully it is okay! Let me know if I can do anything (or any advice, suggested edits etc.)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -19,7 +19,7 @@
     "key-spacing": 1,
     "linebreak-style": 2,
     "max-depth": [1, 4],
-    "max-params": [1, 5],
+    "max-params": [1, 6],
     "new-cap": 2,
     "no-alert": 2,
     "no-caller": 2,
--- a/docs/underscore.html
+++ b/docs/underscore.html
@@ -29,13 +29,13 @@
               </div>
               <pre><code>Underscore.js <span class="hljs-number">1.8</span><span class="hljs-number">.3</span>
 http:<span class="hljs-comment">//underscorejs.org</span>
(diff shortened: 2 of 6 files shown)


jashkenas:
It's a neat feature, but I think a little too complex and nice for core Underscore. Try for _.contrib, or as a plugin. Thanks!

PR #2598, opened 2016-10-02
Always use strictEqual in tests

captbaritone:
By switching all assertions to use `strictEqual` rather than `equal` I was able to discover three worthwhile improvements to our tests:

* https://github.com/jashkenas/underscore/pull/2597
* https://github.com/jashkenas/underscore/pull/2596
* https://github.com/jashkenas/underscore/pull/2595

I think it's a good idea to have our tests generally prefer the more explicit: `strictEqual`. Not only could it help avoid some subtle bug going unnoticed, but it also gives the reader of the test suite a more concrete description of the expected behavior.

At some point, after we've upgraded to a newer version of ESLint, we might even consider enableing the rule [`qunit/https://github.com/platinumazure/eslint-plugin-qunit/blob/master/docs/rules/no-assert-equal.md`](https://github.com/platinumazure/eslint-plugin-qunit/blob/master/docs/rules/no-assert-equal.md) (cc @platinumazure)

### DIFF
--- a/test/arrays.js
+++ b/test/arrays.js
@@ -4,20 +4,20 @@
   QUnit.module('Arrays');
 
   QUnit.test('first', function(assert) {
-    assert.equal(_.first([1, 2, 3]), 1, 'can pull out the first element of an array');
-    assert.equal(_([1, 2, 3]).first(), 1, 'can perform OO-style "first()"');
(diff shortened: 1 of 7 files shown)


megawac:
👍 good stuff!

PR #2599, opened 2016-10-02
Clarify order of operations in cross-document test

captbaritone:
The previous code (introduced in #1424) used the `iDoc` variable name for two purposes on the
same line. Not only was that very confusing, it meant `iDoc` was being
used before it was actually assigned.


captbaritone:
Review @braddunbar ?

jashkenas:
Looks good!

PR #2603, opened 2016-10-08
Move `_.property` and `_.propertyOf` tests to objects

captbaritone:
These functions are listed in the docs under "Objects". To keep things
consistent, I've moved their tests to the `tests/objects.js` file.


jridgewell:
LGTM.

PR #2613, opened 2016-10-17
[optimizeCb] Combine null and 3 as multi-case for argCount switch statement

kingdido999:
Instead of checking null value of argCount within switch expression, I combined null and 3 as a multi-case statement, which makes optimizeCb function more readable.


akre54:
Yeah I like this a lot better. Thanks.

PR #2615, opened 2016-10-20
Spelling

jsoref:
I haven't run any tests.

I know that some downstreams have expectations of spellings in the underscore test output. I believe it's reasonable for them to update their spelling expectations when they update this library.

Some words in underscore are intentionally misspelled (`gettin`, `ridanculous`) and I've left them alone.

Before this pr, both `falsy` and `falsey` were used (roughly evenly). I'm trying to push underscore to a single spelling. I have no particular preference as to which spelling it uses, but it's really helpful if it standardizes on a single spelling.

My spell checker objects to `Javascript`, and as it's brand (`JavaScript`), I happen to agree with it. There shouldn't be much harm in making this change. But you can certainly drop this change if you object strongly.

### DIFF
--- a/docs/underscore.html
+++ b/docs/underscore.html
@@ -3517,7 +3517,7 @@ NB: <code>oldSettings</code> only exists for backwards compatibility.</p>
               <div class="pilwrap ">
                 <a class="pilcrow" href="#section-171">&#182;</a>
               </div>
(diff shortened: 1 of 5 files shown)


jashkenas:
This is wonderful! Thanks.

PR #2619, opened 2016-10-28
Improve debounce performance

jridgewell:
Re: https://github.com/jashkenas/underscore/pull/2340#issuecomment-217624935

This is as simple as I can make it while being much faster for multiple rapid calls.

/cc @dcorb


jashkenas:
So this is more or less reverting the merge...

How does this compare with what we started out with? Faster? Edge cases? Clearer?

jgonggrijp:
Superseded by #2892.

PR #2620, opened 2016-10-29
Make _.invoke support deep method paths

### LINKED ISSUE #2612: _.invoke should support deep access array syntax
I'm working on documentation for the changes introduced in #2494. As part of that work, I've gone through each method and identified exactly how it's affected by the addition of deep property access. In doing so, I realized that I missed one method: `_.invoke`.

I'm working on an implementation, but I figured that in the meantime I'd document the shortcoming here.

The implementation I'm working on will have the following edge-case behavior:
- Methods are executed in the context of their immediate parent.
- If the method, or any parent object, doesn't exist, the result will be `undefined`.

captbaritone:
Fixes #2612

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -292,7 +292,7 @@
 
   // Invoke a method (with arguments) on every item in a collection.
   _.invoke = restArgs(function(obj, path, args) {
-    var contextPath = [], func;
+    var contextPath, func;
     if (_.isFunction(path)) {
       func = path;
     } else if (_.isArray(path)) {
@@ -302,7 +302,9 @@
     return _.map(obj, function(context) {
       var method = func;
       if (!method) {
-        if (contextPath.length) context = deepGet(context, contextPath);
+        if (contextPath && contextPath.length) {
+          context = deepGet(context, contextPath);
+        }
         if (context == null) return void 0;
         method = context[path];
       }
--- a/underscore.js
+++ b/underscore.js
@@ -300,12 +300,13 @@
       path = path[path.length - 1];
     }
     return _.map(obj, function(context) {
-      if (!func) {
+      var method = func;
+      if (!method) {
         if (contextPath.length) context = deepGet(context, contextPath);
         if (context == null) return void 0;
-        func = context[path];
+        method = context[path];
       }
-      return func == null ? func : func.apply(context, args);
+      return method == null ? method : method.apply(context, args);
     });
   });
 
--- a/underscore.js
+++ b/underscore.js
@@ -146,6 +146,15 @@
     };
   };
 
+  var deepGet = function(obj, path) {
+    var length = path.length;
+    for (var i = 0; i < length; i++) {
+      if (obj == null) return void 0;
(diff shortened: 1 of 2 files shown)


captbaritone:
Review @jridgewell ?

underscore.js
```diff
@@ -282,11 +291,21 @@
...
+      func = path;
+    } else if (_.isArray(path)) {
+      contextPath = path.slice(0, -1);
+      path = path[path.length - 1];
+    }
+    return _.map(obj, function(context) {
+      if (!func) {
+        if (contextPath.length) context = deepGet(context, contextPath);
+        if (context == null) return void 0;
+        func = context[path];
```
jridgewell: This will cause an issue with the second iteration, since it won't get the function from the next `obj` again. Please add a test.

underscore.js
```diff
@@ -282,11 +291,21 @@
...
+    if (_.isFunction(path)) {
+      func = path;
+    } else if (_.isArray(path)) {
+      contextPath = path.slice(0, -1);
+      path = path[path.length - 1];
+    }
+    return _.map(obj, function(context) {
+      if (!func) {
+        if (contextPath.length) context = deepGet(context, contextPath);
+        if (context == null) return void 0;
```
jridgewell: return `context`?

underscore.js
```diff
@@ -282,11 +291,21 @@
   };
 
   // Invoke a method (with arguments) on every item in a collection.
-  _.invoke = restArgs(function(obj, method, args) {
-    var isFunc = _.isFunction(method);
-    return _.map(obj, function(value) {
-      var func = isFunc ? method : value[method];
-      return func == null ? func : func.apply(value, args);
+  _.invoke = restArgs(function(obj, path, args) {
+    var contextPath = [], func;
```
jridgewell: Let's not initialize it.

underscore.js
captbaritone: Thanks for catching this! Fixed https://github.com/jashkenas/underscore/pull/2620/commits/d3a1020d9d61b99d5dbe182ab240010dd6370375

underscore.js
captbaritone: Fixed: https://github.com/jashkenas/underscore/pull/2620/commits/97cfcbcbbcedf544a13127dcca3e0ddad94ff830

underscore.js
captbaritone: It seems more appropriate to me to always return `undefined` if the path specified cannot be resolved.

If we return `context` here, we would get the following behavior, which I think is incorrect/confusing.

```
var userNames = _.invoke([{user: null}], ['user', 'getName']);
// => [null]
```

Did you have a specific example in mind where returning `undefined` would be unintuitive/wrong?

captbaritone:
```
lodash.invokeMap([{user: null}], ['user', 'getName']);
// => [undefined]
```

`_.invoke([{user: null}], 'user')` will still return `[null]`.

Issue #2629, opened 2016-11-21
_.eq redundant comparison when both null


ahonn:
in `_.eq` function:
```
 eq = function(a, b, aStack, bStack) {
    // Identical objects are equal. `0 === -0`, but they aren't identical.
    // See the [Harmony `egal` proposal](http://wiki.ecmascript.org/doku.php?id=harmony:egal).
    if (a === b) return a !== 0 || 1 / a === 1 / b;
    // A strict comparison is necessary because `null == undefined`.
    if (a == null || b == null) return a === b;
   ......
}
```

a === b is true when a = null and b = null, so the function will return true.
and then, if `(a == null || b == null)` is true, return a === b. 
It would never be true, why not directly return false:
```
if (a == null || b == null) return false;
```

I checked the #2180 and #2179 , but I still don't understand why not directly return false?

jridgewell:
Yah, looks like we could. Want to open a PR?

ahonn:
Nice, I have opened a PR #2630

demoneaux:
@ahonn since we are using `a == null || b == null`, `a` and `b` could be `null` or `undefined` due to `==` not being a strict equal comparison. Hence `a === b` is needed to ensure that `a` and `b` are not the same, if one is `null` and the other is `undefined`.

jridgewell:
Yes, but we've already done a `a === b` check above it. So, if either `a` or `b` is `null` or `undefined`, we know that both aren't.

demoneaux:
> Yes, but we've already done a a === b check above it

Ah, I completely missed that :| Thanks @jridgewell!

ahonn:
When running `if (a == null || b == null) ...`, `a === b` would never be true.
Because if `a === b` is true, will return `a !== 0 || 1 / a === 1 / b`. So, That must be false.

PR #2630, opened 2016-11-21
optimized _.eq function, remove redundant comparison when both null

ahonn:
PR for #2629


underscore.js
```diff
@@ -1185,7 +1185,7 @@
     // See the [Harmony `egal` proposal](http://wiki.ecmascript.org/doku.php?id=harmony:egal).
     if (a === b) return a !== 0 || 1 / a === 1 / b;
     // A strict comparison is necessary because `null == undefined`.
```
jridgewell: Let's update the comment as well.

PR #2634, opened 2016-12-08
Dogfood _.times in _.unzip

captbaritone:
I've included @earforth's commit from #2632 and cleaned up the formatting.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -621,8 +621,8 @@
   // each array's elements on shared indices.
   _.unzip = function(array) {
     var length = array && _.max(array, getLength).length || 0;
-    return _.times(length, function(index) { 
-        return _.pluck(array, index);
+    return _.times(length, function(index) {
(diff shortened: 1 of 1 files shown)


jashkenas:
Performance impact?

captbaritone:
Not sure of actual timing, but this will add an additional function call (using `.call()`) for each item in the collection. On the plus side it will also pre-allocate the array.

jashkenas:
Might be worth checking — I'm pretty sure that's the reason why it's using a `for` loop in the first place.

captbaritone:
I'm not personally very invested in this change. I'll close this issue. If someone else (@earforth @craigmichaelmartin, @jridgewell) want's to do the legwork of doing a perf evaluation, we can reopen.

earforth:
If you need performance, please replace the pluck into codes , it's a  function call, too.

PR #2637, opened 2016-12-17
perf(isObject): Don't cache typeof & check value first

bigopon:
Hi, from this [benchmark](https://gist.run/?id=680c1d0d904fad05549ab93fa033897d), avoid caching can help improve performance around 25 - 30%

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -1310,8 +1310,7 @@
 
   // Is a given variable an object?
   _.isObject = function(obj) {
-    var type = typeof obj;
-    return type === 'function' || type === 'object' && !!obj;
+    return obj !== null && (typeof obj === 'function' || typeof obj === 'object');
   };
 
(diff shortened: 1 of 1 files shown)


jgonggrijp:
I tried pasting the original implementation into your benchmark, as it wasn't included. I got underwhelming differences as a result:

```
strict x 1,214,052 ops/sec ±0.95% (60 runs sampled)
original x 1,148,546 ops/sec ±2.52% (59 runs sampled)
loose x 1,023,326 ops/sec ±1.45% (59 runs sampled)
cache x 1,186,416 ops/sec ±2.51% (58 runs sampled)
no-cache x 1,034,401 ops/sec ±1.68% (59 runs sampled)
```

Adding to that that nobody has actually complained about `isObject` being unreasonably slow, I'm just closing this.

PR #2647, opened 2017-01-22
Fix parameter name inconsistencies in index.html

wtokuno:
Almost all Collection functions have the parameter named `list` but one of the collection function `partition` takes the parameter named `array`.
Similary the `sortedIndex` is an Array function, but it takes the parameter named `list`.


akre54:
Thanks

Issue #1189, opened 2013-07-06
Remove `/` from `entityMap.escape`.


jdalton:
Similar to #615 `_.escape` doesn't need to worry about '/'.

mitar:
What? `/` should be escaped? I opened #2650 to fix that, but now it has been closed with a link to this issue. What reason is that?

> Characters like > and / don't require escaping in HTML and have no special meaning unless they're part of a tag or an unquoted attribute value (there are no tags as they escaped).

`>` and `/` have a meaning. The are special characters and could be lead to tricky situations. We should not try to be too smart here, it does not really work well for security.

For example, do you know that the following is a valid HTML 4 snippet?

```
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN"
  "http://www.w3.org/TR/html4/strict.dtd">
<html>
<head>
<title/hello/
<body>
<p/hello/
```

See how `/` have special meaning?

mitar:
Also, in meantime, it seems that underscore is escaping `>` again. Could we start escaping `\` as well again?

mitar:
I think that conversation is invalid. Nobody here made any serious computer security argument. Just saying that "because I do not know how this could be exploited" is not an argument why this should not be done.

mitar:
> Why all those characters, instead of just the ones the HTML spec defined after careful research of existing behavior? It seems no one knows, and there are no examples either.

Because there are multiple HTML specs you have to care about. Not just the latest one. And multiple browser implementations. Not just Chrome, Firefox, and IE. Libraries as general as this one can be used in so many contexts. Do you really know of all parser engines somebody implemented? Various browser implementations for accessibility, and how they convert bad HTML into something they operate on?

jdalton:
@mitar The escape/unescape methods aren't intended to be a full-on robust solution. It hits the high notes. You might check out [he (HTML entities)](https://www.npmjs.com/package/he) instead of spamming a 3 yrs old thread.

mitar:
> The escape/unescape methods aren't intended to be a full-on robust solution.

But they could be by adding one more character. You are like almost there.

I am spamming this thread because it was mention as an argument why #2650 is closed. And by reading through comments here I was not convinced by them.

jdalton:
> But they could be by adding one more character. You are like almost there.

Naw, it's way more complicated than that.

> And by reading through comments here I was not convinced by them.

You don't have to be.

mitar:
[...]

PR #2650, opened 2017-02-10
Escape / as well

mitar:
It is on [OWASP](https://www.owasp.org/index.php/XSS_%28Cross_Site_Scripting%29_Prevention_Cheat_Sheet) recommended list.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -1462,7 +1462,8 @@
     '>': '&gt;',
     '"': '&quot;',
     "'": '&#x27;',
-    '`': '&#x60;'
+    '`': '&#x60;',
+    '/': '&#x2F;'
   };
   var unescapeMap = _.invert(escapeMap);
 


captbaritone:
Can you point to a specific example of where this would be a problem?

mitar:
https://security.stackexchange.com/questions/49852/owasp-escaping-recommendation-why

mitar:
This is a valid HTML 4:

```
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN"
  "http://www.w3.org/TR/html4/strict.dtd">
<html>
<head>
<title/hello/
<body>
<p/hello/
```

See how `/` has a special meaning?

akre54:
#1189

mitar:
I think that conversation is invalid. Please reopen this pull request. Nobody there made any serious computer security argument. Just saying that "because I do not know how this could be exploited" is not an argument why this should not be done.

akre54:
Can you give a good example of when it could be?

As @jdalton pointed out in the linked issue, if you need strict guarantees about correctness you should be using [he](https://github.com/mathiasbynens/he).

Issue #2627, opened 2016-11-14
_.isEmpty, _.size, etc. fail for ES2015 Set objects


jonathaneunice:
Underscore returns incorrect results for `Set` objects. E.g. believes they are empty when they are not. I assume other Underscore functions for iteration, inclusion testing, and so on will similarly fail.

```
~$ node
> u = require('underscore')
...
> s = new Set()
Set {}
> u.isEmpty(s)
true
> s.add(44)
Set { 44 }
> u.isEmpty(s)
true
> u.size(s)
0
> u.VERSION
'1.8.3'
```

Several open issues mention ES2015 `Set` and `Map` (e.g. [this one](https://github.com/jashkenas/underscore/issues/2451)); none seem the same as this bug. Reading source for v1.9.0, it seems to rely on `defaultKeys` (alias for `Object.keys`). `Object.keys` does not produce keys for `Set` objects, so this is likely to fail in the same ways it does now. Similarly `_.size` seems to rely on the the `.length` property, which `Set` does not have; it has an idiosyncratic `.size` property instead, which the source does not seem to query.

`Set` objects are new, and seem very immature. Still, when JS fails to do the right thing, I look to Underscore to make things right, or at least a little more logical. Hoping that a near release can help wrangle the new structures.

akre54:
I agree. In the short term, we should have a logic branch for `_.isMap` and `_.isSet` and check for `obj.count`. Want to open a pull?

jonathaneunice:
Considering taking a crack at it. Slightly daunted by the need to redefine core functions (e.g. `getLength`, `nativeKeys`) to incorporate `Set` and `Map`, or to add alternative logic paths ("are we dealing with these newfangled types? Okay, well then...") to a wide variety of Underscore functions. It's a considerable surgery for a first-time contributor. Any guidance?

captbaritone:
@jonathaneunice Two tips I would give:

1. Trust the tests. We have very good test coverage. As long as all existing tests continue to pass, you can assume you haven't broken anything too meaningful.

2. Feel free to open incomplete pull requests for [early feedback](https://42floors.com/blog/startups/thirty-percent-feedback). If you are unsure if you are on the right track, we can point you in the right direction.

jgonggrijp:
Duplicate of #2147.

PR #2651, opened 2017-02-10
add support for ES6 Set and Map objects

eduardbcom:
Have added conditions to fix https://github.com/jashkenas/underscore/issues/2627 issues.

I've added additional functions to check if we work with Set or Map.
Let me know if it's okay or not.
I decided not to use _.isSet(obj) cuz, well, that's a problem to use underscore functions within underscore code. Sometimes I tried to call the '_.function_name_here' but it was not defined yet.

That's more like 30% percent work, so I need to have some feedback from you guys @captbaritone @akre54 

Thanks!

> let _ = require('./underscore');
undefined
> let s = new Set([1, 2, 1])
undefined
> let m = new Map([['one', 1], ['two', 2], ['one', 1.0]])
undefined
> _.each(s, function() { console.log(arguments); }) 
{ '0': 1, '1': 1, '2': Set { 1, 2 } }
{ '0': 2, '1': 2, '2': Set { 1, 2 } }
Set { 1, 2 }
> _.each(s, (key, value, s) => console.log(key, value, s))
1 1 Set { 1, 2 }
2 2 Set { 1, 2 }
Set { 1, 2 }
> _.each(m, (key, value, m) => console.log(key, value, m))
1 'one' Map { 'one' => 1, 'two' => 2 }
2 'two' Map { 'one' => 1, 'two' => 2 }
Map { 'one' => 1, 'two' => 2 }
> _.each(m, (value, key, m) => console.log(value, key, m))
1 'one' Map { 'one' => 1, 'two' => 2 }
2 'two' Map { 'one' => 1, 'two' => 2 }
Map { 'one' => 1, 'two' => 2 }
> _.isEmpty(s)
false
> _.isEmpty(m)
false
> _.size(s)
2
> _.size(m)
2
>

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -14,6 +14,7 @@ module.exports = function(config) {
 
     // list of files / patterns to load in the browser
     files: [
+      'node_modules/babel-polyfill/dist/polyfill.js',
       'test/vendor/qunit-extras.js',
       'test/qunit-setup.js',
       'underscore.js',
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
   "main": "underscore.js",
   "version": "1.8.3",
   "devDependencies": {
+    "babel-polyfill": "^6.22.0",
     "coveralls": "^2.11.2",
     "docco": "*",
     "eslint": "1.10.x",
--- a/test/collections.js
+++ b/test/collections.js
@@ -23,16 +23,6 @@
     assert.deepEqual(answers, ['one', 'two', 'three'], 'iterating over objects works, and ignores the object prototype.');
     delete obj.constructor.prototype.four;
 
-    answers = [];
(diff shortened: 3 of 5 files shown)


akre54:
> I decided not to use .isSet(obj) cuz, well, that's a problem to use underscore functions within underscore code

Why do you say that? I'd rather we dogfood a function than creating a new one.

eduardbcom:
I've removed logic for the _.each method, let's focus on _.isEmpty and _.size for now.
You can see that I have added babel polyfill in order to work with Set and Map within karmajs.

Is it expected code for isEmpty and size ? @akre54 @captbaritone 

I am wondering why I still receive build failed, my local version (npm run test-browser and test-node) works fine.

Thx.

PR #2652, opened 2017-02-10
Allow users to prefix searches with `_.`

captbaritone:


docs/main.js
```diff
@@ -3,6 +3,10 @@
   var sections = document.querySelectorAll('.searchable_section');
   var searchInput = document.getElementById('function_filter');
 
+  function searchValue() {
+    return searchInput.value.trim().replace(/^_\./, '');
```
akre54: How about `/^_\.?/` so you don't need to type the dot? Otherwise lgtm

captbaritone: Updated.

captbaritone:
I'll cherry-pick this over.

captbaritone:
Done.

PR #2657, opened 2017-02-22
_.segmentize function to split numeric arrays into arrays of conseque…

ephraimberkovitch:
_.segmentize function to split numeric arrays into arrays of consequent values, i.e.:
[1,2,3,4,5,10,11,12,20,30,31,40,41,42] => [ [1,2,3,4,5], [10,11,12], [20], [30,31], [40,41,42] ]
Useful in mathematical calculations, plotting graphs, etc.

_.isNumericArray function to verify numeric arrays

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules
 *.swp
 nyc_output
 coverage
+scratchpad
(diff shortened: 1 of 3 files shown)


jashkenas:
Thanks, but this is too special-case to be included in Underscore.

PR #2660, opened 2017-03-06
Don't cache Date.now

andrewminer:
Certain test frameworks override Date.now to match their stubbed
versions of `setTimeout`, `setInterval`, etc.  While Underscore works
well with replaced versions of those function (because it gets the
current copy each time they are used), this isn't true of the `Date.
now` function, which is cached when Underscore is first initialized.

This change ensures that Underscore uses the current version of the
`Date.now` function each time it is called for, and therefore
improves compatibility with such test libraries.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -1451,9 +1451,15 @@
   };
 
   // A (possibly faster) way to get the current timestamp as an integer.
-  _.now = Date.now || function() {
-    return new Date().getTime();
-  };
+  if (Date.now) {
+    _.now = function() {
+      return Date.now();
+    };
+  } else {
+    _.now = function() {
+      return new Date().getTime();
+    };
+  }
 
   // List of HTML entities for escaping.
   var escapeMap = {


andrewminer:
To add a little detail about this PR...  I'm using [sinon.js](http://sinonjs.org/) with its [fake timers](http://sinonjs.org/releases/v1.17.7/fake-timers/) functionality in a project, and I ran into a problem where a `debounce`'d function was never called, even though I advanced the fake clock well beyond the time when the debounce was expired.  After much digging around, I figured out that this was because the `_.now` function was initialized to equal `Date.now` and was never updated thereafter to use the version provided by the timer stub.

While there are a many ways to address the issue, I figured this would be the simplest all around, and it matches the general policy in the rest of Underscore's code of not caching such functions internally.

jashkenas:
I don't know about this — the current Underscore semantics seem more desirable than looking up `Date.now` every time.

Wouldn't a cleaner fix be to ensure that Sinon's Date stubbing (and any other global object manipulation) is the first thing that runs in your tests? 

It seems to me like if that global manipulation *doesn't* run first, all bets are off in general.

andrewminer:
> Wouldn't a cleaner fix be to ensure that Sinon's Date stubbing (and any other global object manipulation) is the first thing that runs in your tests?

I did try that at first, but while I was trying to get that straightened out, I noticed that Underscore's use of `setTimeout` and similar functions does *not* get cached, and therefore was working properly.  I figured it made sense that Underscore should be consistent in either caching or not caching functions used from the global namespace.  Since the caching was causing me problems, I opted to move things toward "not caching".

It turns out Lodash had a similar problem (https://github.com/lodash/lodash/issues/304), and they solved it by offering a work around using their `runInContext` function.  Since Underscore doesn't have such a mechanism, and was really only one small change away from supporting Sinon, I thought it would be nice to save others the confusion.  If it was something they got enough complaints about to address, I figured you would like to have a fix as well.

Of course, I'll trust your judgement on what works best for Underscore.

andrewminer:
Just wanted to check in on this PR.  I'm happy to close it if you don't think it would make a good addition to Underscore, but it would good to resolve it one way or the other.  😉

jashkenas:
Just thinking it through from first principles, Underscore is *not* the only library or piece of third-party code that might be using `setTimeout` and `Date.now`. The correct way to do this is to load Sinon's global object modifications as the first thing on the page. Otherwise Sinon forces every other JavaScript library to patch themselves to accommodate.

PR #2666, opened 2017-04-01
very small optimizations (0.1-0.5% faster on different browsers)

NightmareZ:
You must repeat the tests 10 times per one test session or more to see the difference.
IE and Edge shows difference more then Firefox and especially Chrome.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -111,10 +111,7 @@
         case 1: return func.call(this, arguments[0], rest);
         case 2: return func.call(this, arguments[0], arguments[1], rest);
       }
-      var args = Array(startIndex + 1);
-      for (index = 0; index < startIndex; index++) {
-        args[index] = arguments[index];
-      }
+      var args = ArrayProto.slice.call(arguments, 0, startIndex);
       args[startIndex] = rest;
       return func.apply(this, args);
     };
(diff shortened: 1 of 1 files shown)


captbaritone:
What was your methodology for measuring the performance change? Can you share some reproduceable benchmarks?

NightmareZ:
Just add on first string to each test file this code:
`for (var kkk = 0; kkk < 10; ++kkk)`

captbaritone:
@NightmareZ I'm not sure I follow. Could you include a full benchmark that others can run? Perhaps on https://jsperf.com/ ?

Performance is very nuanced and I doubt we will merge a change that has such a minor improvement, since it's hard to know how it might behave (negatively) in other environments/interpreters. If you are passionate about getting this change merged, you will need to make it easy for others to see what the affect is across different browsers.

NightmareZ:
default version tests (x20):
https://cdn.russiancoders.ru/underscore-test/default/test/index.html

optimized version tests (x20):
https://cdn.russiancoders.ru/underscore-test/optimized/test/index.html

NightmareZ:
Edge (std):  348290 ms
Edge (opt):  347261 ms

captbaritone:
Our test suite is not intended to be used as a performance test. Performance benchmarks must be carefully written to ensure that they actually measure the thing that you think they are measuring. JavaScript engines (like Chakra or V8) do all kinds of advanced optimizations to your code depending upon how it is being run. This means that unless you write your test very carefully, you might see a huge speedup simply because (for example) your test was written in such a way that the JavaScript engine was able to figure out that your code didn't have any side affects, and thus never ran the code at all!

JavaScript performance benchmarking is a very interesting area. If you are interested in learning more, I think this talk is a good place to start:

https://www.youtube.com/watch?v=65-RbBwZQdU

captbaritone:
I'm going to close this pull request for now. If you end up writing some robust benchmark tests feel free to comment here and we can look at them more deeply. Thanks for taking the time to work on Underscore!

jdalton:
(former Chakra perf PM here, currently on the Edge team). FWIW I actually get the `opt` version taking 2ms longer. I know from my Chakra days that it more sensitive than other engines with respects to `arguments` objects and it doesn't like passing `arguments` to functions. In many cases that will cause it to creating the full `arguments` object (stack and variable binding behavior in all). This PR seemed sketchy from the jump.

PR #2674, opened 2017-05-26
Add missing function alias to documentation

matildasmeds:
[`contains` has two aliases `include` and `includes`](https://github.com/jashkenas/underscore/blob/master/underscore.js#L288)
This patch adds `include` to documentation


captbaritone:
Thanks for this! Could you also update the `data-aliases` attribute in the table of contents so that search works properly for this alias?

matildasmeds:
The search actually works with the current patch too, because `data-aliases` attribute has `includes` as a value. So if you search with `include`, `contains` is returned. Would you still prefer me to update `data-aliases`?

captbaritone:
Thanks! Might as well be explicit.

captbaritone:
I'll cherry pick this onto fh-pages later today.

PR #2678, opened 2017-06-09
Update debounce result

mqyqingfeng:
In debounce function, when immediate is false, since use the
setTimeout (_.delay), when debounced function return result, the result value is always undefined.


akre54:
This is working as intended.

Issue #2676, opened 2017-06-06
New function findKeys proposition


FabriceGaudin:
I'd like to use a function but it seems it's missing. I didn't find any issue for this, and I didn't find it in https://github.com/documentcloud/underscore-contrib (but I'm not sure how to search efficiently here)

function description:
### findKeys
`_.findKeys(object, predicate, [context])`
Returns an array of keys where the predicate truth test passes (or empty array). 

`_.findKeys({ a: 1, b: 2, c: 3, d: 2}, function (val, key) {
    return val === 2;
})`
` => ["b", "d"]`

`_.findKeys({ a: 1, b: 2, c: 3, d: 2}, function (val, key) {
    return val === 4;
})`
` => []`

Am I the only one who would need this ? If not, i can do a PR

wesvetter:
I can't speak for the maintainers but it seems like a pretty specific use-case. And the fact that Lodash doesn't implement anything similar makes it even more suspect. I can see wanting a _list_ of objects that match the predicate, but I've never needed to find keys matching a predicate on a _single_ object.

Can you give a real-world example of where you need to find keys matching a predicate on an object?

All that said, it's pretty easy to implement yourself using Underscore methods:
```javascript
_.findKeys = function(object, predicate, context) {
  var keyOrNil = function(value, key) {
    if (predicate(value, key)) return key;
  };
  return _.compact(_.mapObject(object, keyOrNil, context));
};
```

jashkenas:
Hey @FabriceGaudin. Can you talk a little bit about the specific use cases where this function would be useful?

sebastien-mignot:
@wesvetter Lodash implements it : https://lodash.com/docs/4.17.4#findKey

sebastien-mignot:
@jashkenas By @wesvetter's logic, that should be enough to implement it in underscore, no ? ^^

jdalton:
@sebastien-mignot You're mixing up [findKey](http://underscorejs.org/#findKey) (singular) with `findKeys` (plural)

sebastien-mignot:
@jdalton 
Ok, my bad. And if every one agrees that sometime you need to find one such key, isn't it reasonable to expect that sometime one will need to find all such keys ?

wesvetter:
> By @wesvetter's logic, that should be enough to implement it in underscore, no ?

Nope, there are many functions that Lodash implements that Underscore does not.

[...]

PR #2681, opened 2017-06-19
Add new function findKeys

FabriceGaudin:
From issue https://github.com/jashkenas/underscore/issues/2676

I adapt this function from existing one `findKey`, but tell me if you think about other needed test/documentation.


anvyne:
`_.findKeys` is a nice method! For what ever it is worth, you could also save some bytes on `_.filter` using this method:

    _.filter = function(obj, predicate, context) {
      return _.map(_.findKeys(obj, predicate, context), function(key) {
        return obj[key];
      });
    };

That may or may not be desirable because it would probably entail a little bit less efficiency.

N.B. I have not tested this implementation of `_.filter`, but it is accurate to the best of my knowledge.

FabriceGaudin:
@anvyne I think this implementation is great, but after testing it, `_.each` doesn't make findKeys pass [this test](https://github.com/jashkenas/underscore/pull/2681/commits/9d66959ac87beca8f2da520dc91e5ac3aed6a308#diff-56642018eb131d449a77225aa1a59123R1138) (which comes directly from [this one](https://github.com/jashkenas/underscore/blob/master/test/objects.js#L1095-L1098)).
The array `[ 1, 2, 3, 4, match: 55 ]` passes `isArrayLike`, so it doesn't use `_.keys` method and length is still 4. I don't really understand when we have to deal with this kind of array, do you know what should be the correct behavior ?

If we manage to modify findKeys as a collection method, I totally agree to use it in `_.filter`

anvyne:
There is a sort of precedence, actually, if we look at `_.find`, which is analogous to `_.filter`.  `_.find` returns the first value that passes a truth test, while `_.filter` returns all values that pass a truth test. 

`_.find` decides whether to use `_.findIndex` or `_.findKey` depending on whether the passed in object is array-like or not. It would be symmetrical to implement `_.filter` in an analogous way. However, there is no method `_.findIndexes`.

Arguably, if there exists the object method `_.findKeys`, there should also exist the array method `_.findIndexes`, and `_.findKeys` should only behave like an object method (that is, in a way that searches all keys regardless of whether the passed object is array-like or not). It would be appropriate to treat array-likes differently than objects only if `_.findKeys` were a collection method. 

However, adding another method like `_.findIndexes` would need careful consideration because it is ideal to keep Underscore short. See #2060. That's not to say `_.findIndexes` definitely should not also be implemented.

If `_.findIndexes` were to never be implemented, then I would prefer a `_.findKeys` collection method to a `_.findKeys` object method because it would probably be more useful even though it would be a slight misnomer and a break of precedence. The ideal situation, however, is to have both `_.findIndexes` and `_.findKeys`.

PR #2687, opened 2017-07-03
Add function _.mergeBy

ershing:
manytime we want to combine two object array through some properties  they own , we just need to put two object arrays into the function and point out the property . we can get the result.


akre54:
No thanks. Feel free to host this as a plugin yourself.

PR #2693, opened 2017-08-17
Support for typed array, DataView and ArrayBuffer in isEquals

### LINKED ISSUE #2692: isEqual incorrectly compares DataView objects
isEqual returns true for the following comparison, when it should return false:

```javascript
_.isEqual(new DataView((new Uint8Array([1,2])).buffer), new DataView((new Uint8Array([5,6,10])).buffer))
```

This is in underscore 1.8.3.

maartenbreddels:
Types much match for typed arrays, and for all of them the byte content must be equal. Fixes #2692

### DIFF
--- a/test/objects.js
+++ b/test/objects.js
@@ -585,7 +585,6 @@
     var i8 = new Int8Array([1,2]);
     var u16 = new Uint16Array([1,2]);
     var u16one = new Uint16Array([3]);
-    b = new Uint8Array([5,6,10]);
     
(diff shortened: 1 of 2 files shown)


jashkenas:
Happy to fix this, but we're not going to throw a naked try with an unrestricted catch right in the middle there.

maartenbreddels:
Ok, a more constructive reply would be appreciated. You're welcome.

PR #2694, opened 2017-08-18
Support for typed array, DataView and ArrayBuffer in isEquals: take 2

### LINKED ISSUE #2692: isEqual incorrectly compares DataView objects

maartenbreddels:
supersedes #2693 fixes #2692 , feedback welcome,

### DIFF
--- a/test/objects.js
+++ b/test/objects.js
@@ -585,7 +585,6 @@
     var i8 = new Int8Array([1,2]);
     var u16 = new Uint16Array([1,2]);
     var u16one = new Uint16Array([3]);
-    b = new Uint8Array([5,6,10]);
     
     assert.ok(_.isEqual(u8, u8b), 'Identical typed array data are equal');
     assert.ok(_.isEqual(u8.buffer, u8b.buffer), 'Identical ArrayBuffers are equal');
--- a/test/objects.js
+++ b/test/objects.js
@@ -588,13 +588,18 @@
     b = new Uint8Array([5,6,10]);
     
     assert.ok(_.isEqual(u8, u8b), 'Identical typed array data are equal');
+    assert.ok(_.isEqual(u8.buffer, u8b.buffer), 'Identical ArrayBuffers are equal');
     assert.ok(_.isEqual(new DataView(u8.buffer), new DataView(u8b.buffer)), 'Identical DataViews are equal');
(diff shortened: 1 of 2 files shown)


jashkenas:
Looking good! Let's add one more check to make sure ArrayBuffer is an object that exists in the runtime before starting, and I think we're there.

maartenbreddels:
Yes I see the error. What's the best way to check. `window.ArrayBuffer` or `root.ArrayBuffer`? And I also have to skip the tests, there I don't have a reference to root, can I use window?

jashkenas:
Why don't you test and cache a `typeof ArrayBuffer` check up at the top of Underscore, and then just check the variable within the deepEquals. You can't use `window`.

jashkenas:
Great. Do you mind cleaning up the formatting and making it nice -- like the rest of the surrounding code? Spaces near parens, semicolons, and so on...

maartenbreddels:
Yeah, installing eslint as we speak, the only issue that I don't know to tackle yet is node 0.8 which doesn't seem to support isView.

jashkenas:
> the only issue that I don't know to tackle yet is node 0.8 which doesn't seem to support isView.

It doesn't? That seems like a bit of a dealbreaker.

akre54:
Do we care to support node 0.8? [Last release](https://nodejs.org/en/download/releases/) was 2014.

jashkenas:
> Do we care to support node 0.8? Last release was 2014.

Fair enough, then.

jasongrout:
>> Do we care to support node 0.8? Last release was 2014.

> Fair enough, then.

Moot point since d73f20c solves that case too, right?

vidartf:
It seems like everyone was OK with this PR. Any chance of getting it merged?

jgonggrijp:
@jashkenas Could you have another look and confirm that you still approve of this? If you do, I can take care of the merge conflicts.

I can also add `isTypedArray` to the exports and polish its implementation a little bit, if you like. Personally, I'd make that a condition for merging.

jashkenas:
Sounds like a fine plan to me. It would also be great if we could drop that big list of `toString` fallback name checks, if they’re not necessary for modern JS environments anymore.

PR #2696, opened 2017-08-27
Ensure all _.random values are integers as said in documentation

### LINKED ISSUE #2695: _.random can return double sometimes
```javascript
_.random = function(min, max) {
	if (max == null) {
		max = min;
		min = 0;
	}
        // if min is not integer, we can not get integer as return
	return min + Math.floor(Math.random() * (max - min + 1));
}
```

_.random(5.1, 6) then we will get double not integer return. I think it is a bug because the doc told me I can get a random integer as return

VzqzAc:
Fixes #2695 

In the current approach, if `min` value is, for example `1.3`, `_.random` can produce values starting at `1.3`, it's an edge case, but with smaller ranges it's more probable.

I used `Math.round` to change the `min` value into an integer, but if this is not the best behavior I'm happy to change it :).

Current implementation:
```
  _.random = function(min, max) {
    if (max == null) {
      max = min;
      min = 0;
    }
    return min + Math.floor(Math.random() * (max - min + 1));
  };
```

New implementation:
```
  _.random = function(min, max) {
    if (max == null) {
      max = min;
      min = 0;
    }
    return Math.round(min) + Math.floor(Math.random() * (max - min + 1));
  };
```

Another approach I thought of was:
`return Math.round(min + Math.floor(Math.random() * (max - min + 1)));`
but I would like to hear your thoughts.

Here are some jsperf snippets: https://jsperf.com/underscore-random2

### DIFF
--- a/test/utility.js
+++ b/test/utility.js
@@ -74,6 +74,7 @@
     var array = _.range(1000);
     var min = Math.pow(2, 31);
     var max = Math.pow(2, 62);
+    var nonIntegerMin = min + 0.1;
 
     assert.ok(_.every(array, function() {
       return _.random(min, max) >= min;
@@ -82,6 +83,11 @@
     assert.ok(_.some(array, function() {
       return _.random(Number.MAX_VALUE) > 0;
     }), 'should produce a random number when passed `Number.MAX_VALUE`');
+
+    assert.ok(_.every(array, function() {
+      var randomValue = _.random(nonIntegerMin, max);
+      return randomValue === parseInt(randomValue, 10);
+    }), 'should produce only Integer values');
   });
 
   QUnit.test('now', function(assert) {
--- a/underscore.js
+++ b/underscore.js
@@ -1447,7 +1447,7 @@
       max = min;
       min = 0;
     }
-    return min + Math.floor(Math.random() * (max - min + 1));
+    return Math.round(min) + Math.floor(Math.random() * (max - min + 1));
   };
 
   // A (possibly faster) way to get the current timestamp as an integer.


craigmichaelmartin:
Thanks for taking the initiative! I think you account for a piece of the problem, but there is still another aspect.

Indeed, the `min` acts in part as the "base" that the random number will be added to, and so having a non-integer "base" number will offset any random number with the same decimals. Your proposed implementation correctly handles this.

However the `min` value is also used in conjunction with the `max` to distribute the random number across the desired range. Having a non-integer `min`/`max` affects this distribution, resulting in a returned "random" value is not, in fact, correctly proportioned (uniformly distributed) from the range. Your proposed implementation does not account for this.

With the proposed implementation:
```diff
random = function(min, max) {
  if (max == null) {
    max = min;
    min = 0;   
  }
- return min + Math.floor(Math.random() * (max - min + 1));
+ return Math.round(min) + Math.floor(Math.random() * (max - min + 1));
};
```

```javascript
const arr = [0, 0, 0, 0];
for (var i = 0; i< 10000; i++) {
  arr[random(1.49, 3)]++
};
arr // [0, 4042, 4016, 1942]
// random numbers are indeed integers, but not uniformly distributed
```

If we round both numbers at the outset, both issues are resolved:
```diff
random = function(min, max) {
  if (max == null) {
    max = min;
    min = 0;
  }
+ max = Math.round(max);
+ min = Math.round(min);
  return min + Math.floor(Math.random() * (max - min + 1));
};
```

```javascript
const arr = [0, 0, 0, 0];
for (var i = 0; i< 10000; i++) {
  arr[random(1.49, 3)]++
};
arr // [0, 3348, 3344, 3308]
// random numbers are indeed integers, and are uniformly distributed
```

VzqzAc:
Right! I totally missed `max` and your explanation makes a lot of sense, though, as you can see in my first comment
> Another approach I thought of was:
> `return Math.round(min + Math.floor(Math.random() * (max - min + 1)));`

Do you think that works? Now after your comment I think it'd be better to go by your proposed solution so the random number is taken from both `max` and `min` properly sanitized, instead of deforming the resulted random.

Making the proposed changes asap, and thanks for pointing it out!

craigmichaelmartin:
Happy to help! Unfortunately, your alternative approach still succumbs to the range distribution problem, since you're rounding after the range is specified (`max - min + 1`). Thanks for tackling this issue! It remains to be seen whether the maintainers want to support this case.

jashkenas:
I don't think we do want to "support" this case, because the function is supposed to work with integers. If we were returning a random float between `min` and `max`, this would be different.

Instead, you can round before you call `random`

PR #2703, opened 2017-10-17
fixed max array index

### LINKED ISSUE #2699: MAX_ARRAY_INDEX  =  Math.pow(2, 53) - 1;   is  a worry set  in line 163
from this property name we may consider a array max lenth is Math.pow(2, 53) - 1, but I have do a simple test ,like this:   var arr = [];  arr.length = Math.pow(2, 53) -1 ;   then running in lastest chrome browser, the console is  print a error: Uncaught RangeError: Invalid array length,  after searching related infomation and test,  I found  Math.pow(2, 32) - 1  is the critical value, this should be corrected, If I understand it wrong also please point out.

iamnapo:
Fixes #2699

As per [8th Edition - ECMAScript 2017](http://www.ecma-international.org/ecma-262/8.0/index.html#sec-array-len).  :octocat:

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -160,7 +160,7 @@
   // should be iterated as an array or as an object.
   // Related: http://people.mozilla.org/~jorendorff/es6-draft.html#sec-tolength
   // Avoids a very nasty iOS 8 JIT bug on ARM-64. #2094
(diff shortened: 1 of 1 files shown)


craigmichaelmartin:
Looks good to me!

iamnapo:
Should this still be open?

PR #2707, opened 2017-10-27
make _.isUndefined work logically

jeffborders:
`isUndefined` currently checks that a variable is declared but not defined. it should check that a variable's type is `undefined`. 

for example:
```js
var un;
_.isUndefined(un); // true
_.isUndefined(notDeclared) // ReferenceError notDeclared is not defined
```


craigmichaelmartin:
That would be nice! The problem, though, is that the ReferenceError is not thrown in underscore's `isUndefined` function, it is thrown in your application code. Therefore, this isn't possible for underscore to guard against this, since the problem occurs in the application code. You can see this if you examine the ReferenceError in your example, or if you try adding a test to underscore for this change.

jeffborders:
@craigmichaelmartin you're absolutely right. i just realized that myself.

PR #2713, opened 2017-11-21
Add _.isNumeric

GaryPEGEOT:
Added `isNumeric` function, to verify whether a variable is a number or a numeric string:
```js
_.isNumeric('12'); // true
_.isNumeric(12); // true
_.isNumeric('foo'); // false
```


jashkenas:
It's a fine function, but not, I think, something that we want for core Underscore...

Issue #2714, opened 2017-11-23
_.uniq sorted array with iteratee which is a nonlinear function,produce unexpected result 


zenghj:
_.uniq([-2,-1,0,1,2], true, function(x) {return x * x})

the result is  [-2, -1, 0, 1, 2]

the expected result should be [-2, -1, 0], am I right?

jashkenas:
Yep, that's a bug!

PR #2715, opened 2017-11-27
[BUGFIX] (Bug #2714) For _.uniq which acted incorrectly on sorted lists using a non injective iteratee

jojois74:
When using an iteratee that is not a one-to-one function, and isSorted=true, then _.uniq would attempt to use the fast algorithm which fails for non one-to-one (injective) functions, because they map to resultant values which are not sorted.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -558,6 +558,8 @@
 
   // Produce a duplicate-free version of the array. If the array has already
   // been sorted, you have the option of using a faster algorithm.
+  // The faster algorithm will not work with an iteratee if the iteratee is not a one-to-one function, so providing an iteratee will disable the faster algorithm
+  // Perhaps a warning should be thrown if an iteratee is provided while isSorted is true, to warn that the faster algorithm will not be used
   // Aliased as `unique`.
   _.uniq = _.unique = function(array, isSorted, iteratee, context) {
     if (!_.isBoolean(isSorted)) {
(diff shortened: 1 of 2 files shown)


jojois74:
This is my first pull request ever, so please, kindly point out if I am doing something wrong, or if I have made any mistakes. Thank you so much.

underscore.js
```diff
@@ -558,6 +558,8 @@
 
   // Produce a duplicate-free version of the array. If the array has already
   // been sorted, you have the option of using a faster algorithm.
+  // The faster algorithm will not work with an iteratee if the iteratee is not a one-to-one function, so providing an iteratee will disable the faster algorithm
```
jashkenas: Comments should be complete sentences, with punctuation, wrapped to an ~80 line width.

underscore.js
```diff
@@ -558,6 +558,8 @@
...
+  // Perhaps a warning should be thrown if an iteratee is provided while isSorted is true, to warn that the faster algorithm will not be used
```
jashkenas: Let's get rid of this sentence — we don't need to publish it along with the library.

jojois74:
Thanks! I updated the lines that the bot flagged, and now I will update the code with the comments that you added. In a minute I will make another commit.

jojois74:
I fixed the comments in the code as per your request. Let me know if you would like me to change anything else.

jashkenas:
Nice work!

jojois74:
Thank you, I really appreciate it!

PR #2720, opened 2018-01-03
Update year of licensure to 2018

sdbedi:
Year range in LICENSE and underscore.js have been updated to 2018. 

TODO:
Consider 2009-Present format to avoid future commits.


jashkenas:
We do this when we cut a new release, but thanks.

PR #2723, opened 2018-01-29
add _.cloneDeep

wy1009:
add `_.cloneDeep` function

### DIFF
--- a/test/objects.js
+++ b/test/objects.js
@@ -286,6 +286,41 @@
     assert.strictEqual(_.clone(null), null, 'non objects should not be changed by clone');
   });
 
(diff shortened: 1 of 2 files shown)


jgonggrijp:
@wy1009 Thanks for taking the effort to submit a pull request and sorry for the long silence!

I just wanted to remark that you did a really good job. Your tests focus on the important things and the code is short and conventional.

I think this should probably be submitted to Underscore-Contrib, but unfortunately that project is not open to pull requests anymore. I'm looking into a way to blow new life into it, however. @jashkenas let's keep this open in the meanwhile.

PR #2726, opened 2018-02-10
uniqBy function added

gurhankokcu:
uniqBy function added for getting duplicate-free version of the array of objects


jashkenas:
Nice function, but too specific for core Underscore — feel free to mix it in to your copy...

Issue #1514, opened 2014-03-09
_.intersection should ignore non-array values.


jdalton:
I noticed that the arrays method `_.intersection` allows object values while others, like `_.union`, ignore them.

``` js
_.intersection([1,2], {'0': 1 });
// => [1];

_.union([1,2], {'0': 3 });
// => [1, 2]
```

I know there was [recent work](https://github.com/jashkenas/underscore/pull/1428) to ignore non-array/arguments for `_.union` and `_.difference`, and others. I think this falls into that bucket.

BSteephenson:
What behavior do you propose instead? Should it just pass over the object?
```
_.intersection([1,2], {'0': 1 });
// => []
or 
// => [1,2]

olleboll:
Is this still Open as in you want this implemented? I've made a solution that simply ignores arguments that aren't arrays and would love to contribute, even though it's very basic :)

jgonggrijp:
This would be a breaking change, which would mean it could not happen until 2.0. In 2.0, however, the set operations should operate on actual `Set`s, which makes the point moot. Closing for this reason.

PR #2730, opened 2018-03-08
Resolves issue #1514 - patches _.intersection to ignore non array values

miguelsolano:
TLDR:

- PR - Summary: Patches _.intersection to ignore non Array values
    - Adds 3 QUnit test testing against string, int, and object
    - Uses Array.isArray() to determine whether something is an array


miguelsolano:
Will there be any movement on this? Or will it most likely not be merged?

jashkenas:
I don't think we ought to just be silently ignoring non-array values. If we do go out of our way to sniff them, it would be to throw an error, not to omit.

But in general, we don't try to overly type-check every argument to every function. That isn't Underscore-style JS...

Issue #2731, opened 2018-03-11
undefined as an argument of argCount in optimizeCb (version 1.8.3)


xlshen:
**undefined as an argument of argCount**

In version 1.8.2: `switch(argCount == null ? 3 : argCount){...} `
If the argCount is undefined, such as` _.each()`, will work fine because of `undefined == null`.
But if merged in version 1.8.3, the _.each() will go to the default condition:
`return function() {
            return func.apply(context, arguments);
        };`
Because the `switch()` use the `===` to test and in the inner of Underscore, the case of 3 will never invoked.

[[optimizeCb] Combine null and 3 as multi-case for argCount switch statement #2613](https://github.com/jashkenas/underscore/pull/2613)

captbaritone:
I don't see where it's possible for `optimizeCb` to get passed `null`. Can you give an example of using the external API that will cause `optmizeCb` to get passed `null`?

PR #2732, opened 2018-03-14
undefined

xlshen:
undefined, en ....
[undefined as an argument of argCount in optimizeCb (version 1.8.3)](https://github.com/jashkenas/underscore/issues/2731)

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -73,7 +73,7 @@
       };
       // The 2-parameter case has been omitted only because no current consumers
       // made use of it.
-      case null:
+      case undefined:
       case 3: return function(value, index, collection) {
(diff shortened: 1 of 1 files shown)


underscore.js
```diff
@@ -73,6 +73,7 @@
       };
       // The 2-parameter case has been omitted only because no current consumers
       // made use of it.
+      case void 0:
```
jgonggrijp, line 76: This is an unnecessary change because there is already a null check a few lines above this:

https://github.com/jashkenas/underscore/blob/5d8ab5e37c9724f6f1181c5f95d0020815e4cb77/modules/index.js#L60

Issue #2745, opened 2018-04-23
_.range throws error on Infinite values (1.9.0 regression)


lubelski:
`1.8.3`

```js
_.range(Infinity, -Infinity); // =>  []
```

`1.9.0`

```js
_.range(Infinity, -Infinity); // =>  throws error on array construction 
```

the code that caught us with this change was something like...

```js
const items = [];
const render_range = _.range(_.min(items), _.max(items)); 
```

where previously this would return an empty array.   now it throws.  we put a guard in for `items` being empty but we use range all over.  

It looks like the old assumption that step was positive would implicitly prevent the computed array length from being nonsensical.  

`1.8.3`

https://github.com/jashkenas/underscore/blob/e4743ab712b8ab42ad4ccb48b155034d02394e4d/underscore.js#L683-L685

`1.9.0`

https://github.com/jashkenas/underscore/blob/master/underscore.js#L726-L730

should it just guard for finite values? any non-finite value passed in for stop or start seems wrong, and I'd want it to return `[]` in that case

jashkenas:
Hmm, I'm not sure. Since `_.range` constructs eager arrays of numbers, and not lazy ones — it seems like infinite ranges ought to throw an error instead of returning an empty range.

What do other folks think?

captbaritone:
I would agree that this should throw.

jashkenas:
Then let's leave this alone for now. Happy to add more test cases to assert that it's behaving properly, if anyone has any.

PR #2747, opened 2018-04-24
infinite range fix

lubelski:
Issue: https://github.com/jashkenas/underscore/issues/2745

checks for certain array length values and returns an empty array rather than trying to construct an infinite array


jgonggrijp:
Closing this because #2745 has a "wontfix" label.

PR #2754, opened 2018-05-05
Mimicked effort from Backbone PR 4193

unquietwiki:
I see there are a few other PRs in; and figured the work from [Backbone PR 4193](https://github.com/jashkenas/backbone/pull/4193) would be useful here. There's actually room for some further refinement: more variables can be localized; and StandardJS+ESLint testing had flagged mixed AND/OR tests, that I tried to fix (the automation testing will have to judge that).


unquietwiki:
[testing.txt](https://github.com/jashkenas/underscore/files/1977115/testing.txt) Here are test results; they happen to be encoded with color codes from the utility.

PR #2757, opened 2018-05-10
Add `has` internal function

### LINKED ISSUE #2756: 1.9 regression failing in groupBy
Hi folks, I couldn't find an issue containing the `groupBy` function. So I'm creating one. Please redirect if I missed anything. Thanks for all the hard work. Appreciate it!

When updating from version 1.8.3 to 1.9.0 the behaviour of `groupBy` changed in a way that grouping by arrays was possible in 1.8.3, but it's not possible in 1.9.0. At least not as I'm expecting it to work 🤔 

The below test would succeed in 1.8.3 where it throws an exception in 1.9.0. Am I doing something wrong?

```javascript
var anArray = [1,2];
var list = [{
	arrayProp: anArray,
	otherProp: 1
}, {
	arrayProp: anArray,
	otherProp: 2
}];

var res = _.groupBy(list, 'arrayProp');
// var res = _.groupBy(list, function(element) {
//  	// stringified version of key is equal and gets grouped
//  	return JSON.stringify(element.arrayProp);
// });
var groupByKey = Object.keys(res)[0];
console.log(Object.keys(res).length) // 1
console.log(groupByKey); // 1,2

var expected = 2
var actual = res[groupByKey].length;
if (actual !== expected) {
	throw new Error("TEST FAILURE. EXPECTED " + 2 + ". GOT " + actual)
}
```

`-> Uncaught Error: TEST FAILURE. EXPECTED 2. GOT 1`

captbaritone:
In some cases, `_.countby()` and `_.groupBy()`, we want to support
arrays as valid keys, since they can be coerced to strings. In the other
cases, the keys are known to always be strings anyway, so we can save
having to do the `_.isArray` check.

Fixes #2756.

### DIFF
--- a/test/collections.js
+++ b/test/collections.js
@@ -758,6 +758,9 @@
       6: [jordan]
     };
     assert.deepEqual(_.groupBy(collection, ['stats', 'power']), expected, 'can group by deep properties');
+
+    var foos = [{foo: [1, 2]}, {foo: [1, 2]}]
+    assert.deepEqual(_.groupBy(foos, 'foo'), {"1,2": foos}, 'will coerce arrays to string keys');
   });
 
   QUnit.test('indexBy', function(assert) {
@@ -805,6 +808,9 @@
     grouped = _.countBy(array);
     assert.strictEqual(grouped['1'], 2);
     assert.strictEqual(grouped['3'], 1);
+
+    var foos = [{foo: [1, 2]}, {foo: [1, 2]}]
+    assert.deepEqual(_.countBy(foos, 'foo'), {'1,2': 2}, 'will coerce arrays to string keys');
   });
 
   QUnit.test('shuffle', function(assert) {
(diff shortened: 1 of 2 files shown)


jashkenas:
Let's call this internal function `hasOwnProperty` instead of `shallowHas`. After that, sounds good to me!

captbaritone:
We already have a [`hasOwnProperty`](https://github.com/jashkenas/underscore/blob/4bd6f69b33179517d4ff9f6020637d6f336c5f99/underscore.js#L30), which is the unbound version of `Object.prototype.hasOwnProperty`. 

Maybe we could rename that one to `nativeHasOwnProperty`?

jashkenas:
Oh, whoops. How about just `has()` ?

captbaritone:
@jashkenas How do you want to run releases going forward? This seems like it could be part of a point release since it's technically a regression.

jashkenas:
Sure, I'd absolutely love to get all 1.9 regressions patched up in a quick 1.9.1...

PR #2758, opened 2018-05-17
Speed up _.clone() on arrays

CoderMing:
This is my test：
image
arr.concat() has better performance than arr.slice().

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -1162,7 +1162,7 @@
   // Create a (shallow-cloned) duplicate of an object.
   _.clone = function(obj) {
     if (!_.isObject(obj)) return obj;
-    return _.isArray(obj) ? obj.slice() : _.extend({}, obj);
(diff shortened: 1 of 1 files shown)


jgonggrijp:
@CoderMing Thank you for taking the effort to submit a pull request.

Would you mind extending the `clone` test in `test/objects.js` to add a check that `_.clone` on an array does indeed return a shallow copy and not the array itself? Once you do this, I'll take care of the merge conflicts for you.

jgonggrijp:
@CoderMing While you're at it, please apply this optimization also to `_.toArray` and please double-check whether there are any other places where it might be applicable. Thanks in advance.

jgonggrijp:
A quick test of my own didn't confirm the performance advantage. I just noticed that @CoderMing was doing `.slice(0)` instead of just `.slice()` in their microbenchmark, which might have disabled an optimization in the JS engine.

Closing this until somebody offers stronger evidence that there is an actual performance benefit.

Issue #2741, opened 2018-04-19
1.9.0 regression. Calling first or last on an empty array returns undefined.


ksullivan:
With 1.9.0 `_.first([], 10)` returns `undefined` instead of `[]`.

I think this is a big problem :stuck_out_tongue:

ksullivan:
Wow... It almost looks like this was an intended change... https://github.com/jashkenas/underscore/pull/2513

Personally, I think you would expect `_.first([], 2)` to be `[]` following from `_.first([1,2,3], 2)` being `[1,2]`

jashkenas:
That does look like a nasty little regression to me — when called with the number of items specified explicitly. Any other opinions? @jridgewell?

That said, it does raise the question of what `_.first([], 1)` should be.

jridgewell:
Whoops, poor review on my part. Passing in a `n` argument should always return an array.

jashkenas:
No worries, we can do a 1.9.1 with any other cleanups that arise — plus maybe settling some of the other open PRs...

lubelski:
gracias.  this one bit us too today.

andreialecu:
This one leads to very obscure bugs in edge cases in lots of various places in a bigger app in my case.

Should it not be a priority?

tgolen:
This one also bit us. Luckily it was caught by some unit tests before we rolled it out live. I'd be grateful for a quick fix.

jashkenas:
Alright! Underscore 1.9.1 is now published, with this bugfix in it.

captbaritone:
Thanks @jashkenas

PR #2760, opened 2018-05-26
Fix bug #2741 : modify _.first() and _.last()

benjaminviaud:
Before '_.first([])' return 'undefined'
Now '_.first([])' return []
Same thing for _.last()
And add tests


Zainking:
Your change is not rationale.

jgonggrijp:
@benjaminviaud I'm sorry to say it but @Zainking is right. If the first element of an empty array would be that same array, it would be an array with length 1, so it wouldn't be empty.

jgonggrijp:
Also, #2741 was already fixed.

Issue #2740, opened 2018-04-18
v1.9.0 has failing tests


rodoabad:
Looks like a couple of them are failing after releasing `1.9.0`

https://travis-ci.org/jashkenas/underscore/jobs/368290774

Running against Node 4.

<img width="830" alt="screen shot 2018-04-18 at 3 08 08 pm" src="https://user-images.githubusercontent.com/1240667/38955598-c191af60-431a-11e8-93ec-d06b58c2636d.png">

Will you guys add testing against the newer versions of Node?

jashkenas:
Happy to take a PR that fixes the test suite / adds new versions of Node.

jashkenas:
I've fixed it for now by removing the attempt at Karma/browser testing, instead using the latest supported versions of Node + the latest released Node version:

https://travis-ci.org/jashkenas/underscore

Happy to take a patch that brings back a working Karma config.

jgonggrijp:
@jashkenas Travis doesn't seem to work at all today. Did you disable it altogether or did it break somehow? How about the saucelabs tests?

jashkenas:
> Travis doesn't seem to work at all today. Did you disable it altogether or did it break somehow? How about the saucelabs tests?

I’m afraid I don’t remember. The Travis and Saucelabs tests were external contributions.

If we’re going to bring back a simple CI, both getting off of QUnit, and using GitHub Actions, seem like the right way to go.

jgonggrijp:
By "getting off of QUnit", do you mean replacing QUnit by another test framework such as Jasmine? I like Jasmine, but on the other hand, I don't really think there is anything wrong with QUnit except for it being perhaps a bit less declarative.

Do you agree that we can treat (1) re-enabling CI and (2) moving to a different test framework as separate and independent issues? I think that (1) is sort of a bug while (2) is a lot of work.

jashkenas:
> Do you agree that we can treat (1) re-enabling CI and (2) moving to a different test framework as separate and independent issues? I think that (1) is sort of a bug while (2) is a lot of work.

Definitely. 

My reflexive and uninformed impression was that QUnit was an ancient beast, used in the old days to test jQuery, but not widely used anymore. If it’s still solid, and it isn’t hard to get it working with CI, then I have absolutely no problem with keeping it.

jgonggrijp:
QUnit appears to be actively used and actively maintained.

On an unrelated note, I'm closing this one as it is a bit outdated.

PR #2763, opened 2018-06-05
updating addEventListener to support older browsers

gonzalez:
added saucelab updated browser versions.
updated addEventListener in qunit.js to support older browsers
update isSet and isMap


gonzalez:
Issue https://github.com/jashkenas/underscore/issues/2740

jashkenas:
Thanks! This is important work.

PR #2765, opened 2018-06-11
Utilities - _.uniqueId: Replace 3 digit random integer with timestamp

n1kkou:
Timestamp is always unique, no test needed to check if the value was already used.


jashkenas:
We're not going to just change out this function from under peoples' feet at this point.

n1kkou:
Just as an addition: it will be very helpful in many cases when user has to generate lots of dynamic DOM elements(e.g more than 999).

PR #2766, opened 2018-06-12
Add license annotation

evaliyev:
When bundling a project with Webpack, all comments are removed by default except license annotated ones.
This is required in some companies for legal reasons.
It would be nice to have it in this project too.


evaliyev:
@jashkenas PTAL

jashkenas:
I think that's alright — this project doesn't have anything to do with Webpack ... over any other minifier or bundler.

PR #2784, opened 2018-10-16
 new method eachElementCount 

ar000n:
gives the occurrence count of each element in a array
eachElementCount takes an array and returns a object
[5,4,6,4] => {5: 1, 4: 2, 6:1}


jashkenas:
Thanks, but this is a bit too specific for core Underscore. Feel free to `_.mixin` it in!

PR #2787, opened 2018-11-02
Added lodash compat

def14nt:
Added lodash chain compatability, because if you look at the source:

```js
    function lodash(value) {
      if (isObjectLike(value) && !isArray(value) && !(value instanceof LazyWrapper)) {
        if (value instanceof LodashWrapper) {
          return value;
        }
        if (hasOwnProperty.call(value, '__wrapped__')) {
          return wrapperClone(value);
        }
      }
      return new LodashWrapper(value);
    }
```

It checks for the `__wrapped__` property, so this would add Lodash compatability.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -40,9 +40,9 @@
 
   // Create a safe reference to the Underscore object for use below.
   var _ = function(obj) {
-    if (obj instanceof _) return obj;
+    if ('__wrapped__' in obj && '_wrapped' in obj) return obj;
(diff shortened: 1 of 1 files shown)


jgonggrijp:
Closing this because it seems motivated by incorrect usage of Underscore. In case the owner of the deleted account that originally submitted this PR is reading this: please feel free to comment. We may be able to help you create a correct setup that includes both Underscore and Lodash.

PR #2789, opened 2018-11-07
Symbol support for omit, allKeys

harshitharao:
Fix for [#2457](https://github.com/jashkenas/underscore/issues/2457)

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -1140,7 +1140,10 @@
       iteratee = _.negate(iteratee);
       if (keys.length > 1) context = keys[1];
     } else {
-      keys = flatten(keys, false, false);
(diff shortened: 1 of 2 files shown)


underscore.js
```diff
@@ -1009,6 +1009,7 @@
     if (!_.isObject(obj)) return [];
     var keys = [];
     for (var key in obj) keys.push(key);
+    keys = keys.concat(Object.getOwnPropertySymbols(obj));
```
megawac: I dont know if we can do this due to <ES5 support

jgonggrijp: This wouldn't necessarily be a problem. We could do feature detection à la `nativeKeys` and transparently ignore symbol keys in environments where they are not supported.

PR #2791, opened 2018-12-18
Add _.single

### LINKED ISSUE #2775: Adding _.single
Would anyone be interested in a function that behaves much in the same way as _.find, but only returns the first item to pass the predicate truth test IF only a single item passes it, otherwise undefined.

I'm not sure whether this is useful enough, although there have been a couple of times I've needed something like this to avoid having to validate whether we only have a single item that passes the predicate.

If so i'm happy to do the PR.

haykam821:
Closes #2775.

This function acts similarly to `_.find`, but returns undefined if more than one item passes the predicate test (or there are no matching items).

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -245,6 +245,14 @@
     var key = keyFinder(obj, predicate, context);
     if (key !== void 0 && key !== -1) return obj[key];
   };
+  
+  // Similar to `find`, but returns the first item to pass the predicate
+  // test only if a single item passes; otherwise returns undefined.
+  _.single = function(obj, predicate, context) {
+    var filter = _.filter(obj, predicate, context);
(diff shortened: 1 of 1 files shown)


megawac:
What is the justification for such a method?

haykam821:
@megawac See [this comment](https://github.com/jashkenas/underscore/issues/2775#issue-345784673):

> I'm not sure whether this is useful enough, although there have been a couple of times I've needed something like this to avoid having to validate whether we only have a single item that passes the predicate.

megawac:
I think this is better left up to the user as Im not sure its a common enough case

```
const filteredItems = _.filter(items, predicate);
if (filteredItems.length === 1) {
  const result = _.first(filteredItems);
  // ...
}
```

PR #2796, opened 2019-01-23
Update the copyright date in Underscore.js file

Phe0nix:
A very small changes. Just change the date of copyright in Underscore.js file


jashkenas:
We generally update the date only alongside actual changes made. That way it's accurate.

PR #2801, opened 2019-03-25
Fix deprecation warning about `GLOBAL`

BridgeAR:
This came up on CITGM (https://github.com/nodejs/citgm) and it's
easy to fix.


megawac:
@jashkenas tag dis

PR #2806, opened 2019-04-26
edited the _.each function

alDuncanson:
I changed the variable 'length' in the _.each function to something more obscure and less likely to be used as a key in an object in order to prevent errors while iterating over object keys in the meteor framework.


megawac:
Not an issue as this is a local scope variable reference

PR #2808, opened 2019-05-28
Set and Map support for _.isEmpty

ushakov-igor:
Added support for Set and Map for _.isEmpty

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -1297,11 +1297,12 @@
     return eq(a, b);
   };
 
-  // Is a given array, string, or object empty?
+  // Is a given array, string, set, map or object empty?
   // An "empty" object has no enumerable own-properties.
   _.isEmpty = function(obj) {
     if (obj == null) return true;
     if (isArrayLike(obj) && (_.isArray(obj) || _.isString(obj) || _.isArguments(obj))) return obj.length === 0;
+    if (toString.call(obj) == '[object Map]' || toString.call(obj) == '[object Set]') return !obj.size
     return _.keys(obj).length === 0;
   };
 


patceev:
That's a nice idea. Used to write my own wrapper for this functionality

jgonggrijp:
@ushakov-igor Thank you for taking the effort to submit a pull request.

I think this would be nice to have, but it is a breaking change, so we have to be careful. Your PR may remain open for a while because of this.

For consistency, it would be best if all collection functions would treat `Map` and `Set` in the same way. This is not necessarily a bad idea, but it does make this a rather *big* breaking change.

I also recently learned from #2840 that the `toString`-based type checks like `isMap` are relatively slow. So to avoid a big slowdown in all functions that support `Map` and `Set`, there would have to be an `isAssociative` fuzzy checker analogous to the internal `isArrayLike`. I imagine the implementation might go like this:

```js
var getSize = shallowProperty('size');
var getForEach = shallowProperty('forEach');

function isAssociative(collection) {
    var size = getSize(collection);
    var forEach = getForEach(collection);
    return typeof size == 'number' && typeof forEach == 'function' && size >= 0 && size <= MAX_ARRAY_INDEX;
}
```

jashkenas:
> For consistency, it would be best if all collection functions would treat `Map` and `Set` in the same way. This is not necessarily a bad idea, but it does make this a rather _big_ breaking change.

Agreed. If we’re going to support Maps and Sets (and WeakMaps and WeakSets?), we should do it across the board, in a single release, and not piecemeal function-by-function.

jgonggrijp:
WeakMaps and WeakSets cannot be iterated, so I don't think there's much to support about them in collection functions.

PR #2820, opened 2020-01-07
Fix documentation entry for "sortedIndex" function

tvibliani:
Documentation for the `sortedIndex()` was at least ambiguous at this point, or even wrong, consider for example `let m = [ 1, 2, 4, 4, 4, 5, 6, 7 ];` array, if we are searching for `4`, like so: `_.sortedIndex(m, 4);`, it actually returns `2` which is a **smallest** index of 4 between all it's occurences, not simply the index found by binary search, at this point true binary search could return `3` instead of `2`. So, it is important to state explicitly in the documentation that the index returned is indeed **the smallest index**, as I corrected here. Please apply this update.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1091,7 +1091,7 @@ _.lastIndexOf([1, 2, 3, 1, 2, 3], 2);
       <p id="sortedIndex">
         <b class="header">sortedIndex</b><code>_.sortedIndex(array, value, [iteratee], [context])</code>
         <br />
-        Uses a binary search to determine the index at which the <b>value</b>
+        Uses a binary search to determine the smallest index at which the <b>value</b>
         <i>should</i> be inserted into the <b>array</b> in order to maintain the <b>array</b>'s
         sorted order. If an <a href="#iteratee"><b>iteratee</b></a> function is provided,
         it will be used to compute the sort ranking of each value, including the <b>value</b> you pass.
--- a/test/arrays.js
+++ b/test/arrays.js
@@ -130,9 +130,9 @@
   });
 
   QUnit.test('sortedIndex', function(assert) {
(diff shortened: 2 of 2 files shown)


index.html
```diff
@@ -1091,7 +1091,7 @@ <h2 id="arrays">Array Functions</h2>
       <p id="sortedIndex">
         <b class="header">sortedIndex</b><code>_.sortedIndex(array, value, [iteratee], [context])</code>
         <br />
-        Uses a binary search to determine the index at which the <b>value</b>
+        Uses a binary search to determine the smallest index at which the <b>value</b>
```
jashkenas: How about "lowest"?

tvibliani: Please see my second commit, I also improved the test case for this scenario which was already present in your tests & I noticed that you already use word "smallest" in your code, for better consistency across the project let's leave "smallest" as is, although "lowest" wold do as well, of course.

tvibliani:
The test which is supposed to check if sortedIndex() finds the smallest index is already there, see [line 137](https://github.com/jashkenas/underscore/blob/5a55dd13edd148718a0a0177d711b71b5887d3da/test/arrays.js#L137)  which is quoted here:
```
assert.strictEqual(indexFor30, 2, 'finds the smallest index at which a value could be inserted to retain order');
```
-however the `var number` array used for the test case does not contain duplicates, which is fixed in [this commit](https://github.com/jashkenas/underscore/pull/2820/commits/934052aa9471e537ec9ae175a9e6868a5d4be28b).
This is to ensure consistence of behavior of the sortedIndex() function.

All in all this change makes the quoted test from the line 137 more meaningful by including duplicates, given that a pure binary search algorithm could have a different result in the same scenario.

PR #2821, opened 2020-01-12
Add additional info for isDate documentation

khasburrahman:
This additional info might be helpful for some people that doesn't aware that initializing Date with invalid date string will still returns a Date object. So that checking invalid Date with `isDate` will returns true. 

`_.isDate()` function only checks if a variable type is `Date` not check if it the value is valid.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1826,7 +1826,7 @@ _.isBoolean(null);
       <p id="isDate">
         <b class="header">isDate</b><code>_.isDate(object)</code>
         <br />
-        Returns <i>true</i> if <b>object</b> is a Date.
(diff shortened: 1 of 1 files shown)


jgonggrijp:
Thank you for taking the effort to submit a pull request, @khasburrahman.

I understand that `_.isDate(new Date('invalid')) === true` is confusing. But the confusing part is that `new Date('invalid')` returns an actual `Date`. I think an explanation for that belongs on a web page about `Date`, such as [this one](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date), instead of on a web page about Underscore. Would you agree with that?

jgonggrijp:
@khasburrahman if you have not replied yet by June 30, I will close this PR for the time being. The PR can still be reopened after that if you wish. Thank you for understanding.

PR #2825, opened 2020-02-27
Fix order of assertion parameters

ppenicka:
In the unit test for _.max() several assertions had the 'actual' and 'expected' parameters swapped. Fixed it yields the same result, but looks more consistent and potentially less confusing (a colleague actually got confused by this and we spent some time figuring out if there's a specific reason for them being swapped, concluding there isn't - hence this PR).

### DIFF
--- a/test/collections.js
+++ b/test/collections.js
@@ -578,17 +578,17 @@
   });
 
   QUnit.test('max', function(assert) {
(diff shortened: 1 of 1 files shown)


jashkenas:
Thanks for the cleanup!

PR #2828, opened 2020-02-28
Use npx for local browser testing

jgonggrijp:
I changed the `test-browser` script to use `npx`. This prevents `npm` from messing up the `package.json` and `package-lock.json` when you `npm run test-browser`.

To be on the safe side, I copied the old version of the command to `test-browser-ci` for the Travis build.

@jashkenas FYI. I'm going to submit a bunch of small PRs in parallel to #2826. Please don't worry about the order, I'll fix any merge conflicts in #2826 if the need arises.


jgonggrijp:
When testing on a different machine, I found out that Karma cannot reliably find PhantomJS when following this approach. It happened to work on the machine where I created this PR because I already had a global installation of PhantomJS. I'm withdrawing the PR for this reason.

PR #2831, opened 2020-03-16
Take this from an isolated script context

### LINKED ISSUE #2827: Version 1.9.2 no longer works with ExtendScript .jsx
```
// #include "underscore-1.8.3.js"
#include "underscore-1.9.2.js"

var artists = ['Pharrel Williams', 'Led Zeppelin', 'Rolling Stones'];

_.each(artists, function(artist, index, artists) {
  alert('artist: ' + artist);
});
```

**Error** : _ is undefined

Now try version **1.8.3** it works like a charm

jgonggrijp:
Fixes #2827, in a way that will continue to work after the rollup conversion from #2826. As a bonus, it will also stop rollup from complaining about `this` being undefined at module scope.

@lohriialo this is version D, but with a cleaner commit history.

@jashkenas it might be possible to skip the `self` and `global` checks and just always rely on `Function('return this')()`, but I'll leave that as an optional future step.

### DIFF
--- a/underscore.js
+++ b/underscore.js
@@ -13,7 +13,7 @@
   // instead of `window` for `WebWorker` support.
   var root = typeof self == 'object' && self.self === self && self ||
             typeof global == 'object' && global.global === global && global ||
-            this ||
+            Function('return this')() ||
             {};
 
   // Save the previous value of the `_` variable.


jgonggrijp:
@lohriialo and other ExtendScript users: until this is merged and released, you can just use the version D that I provided in #2827. It is equivalent to Underscore 1.9.2, but with the current fix as well as the fix from #2829.

jashkenas:
I have a feeling like some overzealous linters may complain about the "insecure" use of `eval` after this change — but hey, let’s give it a try.

jgonggrijp:
Thanks @jashkenas. At least the project's own linter doesn't complain.

PR #2837, opened 2020-03-30
Hotfix for #2836

### LINKED ISSUE #2836: Altered export format/API in semver minor breaks community packages
I encountered some very weird bugs today in a third-party package (`natural`) with errors like `TypeError: Cannot read property 'without' of undefined`. This turns out to be because it imports `underscore` like this:

```
var Tokenizer = require('./tokenizer'),
    util = require("util"),
    _ = require('underscore')._;
```

It appears that the latest version exports the whole shebang as a default export, not as a named value on the exports. That’s probably fine…except for any packages that do the above.

Perhaps `_` should still be exposed as an alias for the default export? Changing something so major as the top-level API seems like a major breaking change (not a _minor_ semver change!), and after all `underscore` is used as a dependency by a *lot* of packages out there.

jgonggrijp:
Hopefully fixes #2836.


jashkenas:
I did the same thing in https://github.com/jashkenas/underscore/commit/2a932470303157072015d5ab7f26d40b9deb4634 ... which also includes a test to make sure we don't lose this API in the future.

Issue #2804, opened 2019-04-09
Improve _.isEmpty documentation relating primitive types


mcurrao:
I know this has already been reported way too many times. And I know it because I've faced the issue myself and found the breadcrumbs, which have lead me here. 

`_.isEmpty` behaviour by design only checks enumerable objects. Any primitive value is therefore, considered empty. 
While changing this behaviour has already been discussed (and rejected) more than once, developers have been hitting against `_.isEmpty` for years now (from at least 2012 #432 to 2018 #2788). And well... 2019 if I count myself. 

I also know this our fault, after quickly scanning through the docs we assume it will behave nicely for every variable we throw at it. 
Still, I'm sure adding an explicit example of numbers (please do NOT use 0 as an example, as it would imply it actually checks for number emptyness), and perhaps a note in bold, would help a lot of people in the future. 

Something like this would probably do:

isEmpty `_.isEmpty(object)`
----

Returns true if an enumerable object contains no values (no enumerable own-properties). For strings and array-like objects _.isEmpty checks if the length property is 0.

> _.isEmpty([1, 2, 3]);
> => false
> _.isEmpty({});
> => true

**Note that primitive types are always considered empty:**
> _.isEmpty(42);
> => true
> _.isEmpty(true);
> => true

jgonggrijp:
@mcurrao Could you explain what is so surprising to you about the behaviour? Personally, I'd be very surprised if `_.isEmpty(42)` returned `false`.

mcurrao:
@jgonggrijp When using variables, the concept of emptiness on a primitive-containing variable would seem to indicate the variable content not being set/filled.

Therefore:

```
let var1, var2 = 42;
_.isEmpty(var1);  // sensibly true
_.isEmpty(var2);  // expectancy unsure
```

For arrays and objects, emptyness considering content makes sense. I understand that the concept of emptyness does not apply to numbers, for example. That is why my request was not to modify its behaviour including primitives, but to modify the documentation. 
In Javascript we are used to treating any variable as a boolean variable (`if (!!var1)`). An `isEmpty` function has a tempting name for a one-fits-all solution for checking value existance, as `!!obj` would return true for an empty object, or array.
This becomes especially useful when working with unknown content format. 
```
const formValues= [ { foo: 'lala' }, { }. [ ], 42, null, null, false, [ 8 ] ];
const cleanup = (contentArr) => contentArr.filter( (elem) => _.isEmpty(elem) );
[...]

PR #2840, opened 2020-04-10
Be even more explicit about the semantics of _.isEmpty

jgonggrijp:
This is meant to address #2804.

@mcurrao Please let me know whether this sufficiently addresses the problem for you.


index.html
```diff
@@ -1732,17 +1732,21 @@ <h2 id="objects">Object Functions</h2>
...
+        enumerable own-properties. Note that primitive numbers, booleans and
+        symbols are always empty by this definition.
       </p>
       <pre>
 _.isEmpty([1, 2, 3]);
 =&gt; false
 _.isEmpty({});
 =&gt; true
+_.isEmpty(42);
+=&gt; true
```
jashkenas, line 1749: I’d leave out this example, as it's not demonstrating useful behavior — it's an example of *how not to use Underscore*, not how to use it.

jgonggrijp:
Thanks @jashkenas. As I understand it, you are fine with the current change (given that I have removed the counter-example that you rightly pointed out). So if @mcurrao is fine with this, too, I'll just merge this.

mcurrao:
@jgonggrijp LGTM. Only missing elements would be functions, but as they can be considered non-enumerable objects, I guess there is no need to include them. Thanks!

jgonggrijp:
Sorry, requesting another review anyway because I just noticed a tautology in the code and removed it.

modules/index.js
```diff
@@ -1296,7 +1296,7 @@ export function isEqual(a, b) {
 // An "empty" object has no enumerable own-properties.
 export function isEmpty(obj) {
   if (obj == null) return true;
-  if (isArrayLike(obj) && (isArray(obj) || isString(obj) || isArguments(obj))) return obj.length === 0;
+  if (isArray(obj) || isString(obj) || isArguments(obj)) return obj.length === 0;
```
jashkenas, line 1299: This wasn’t a tautology, it was an optimization! We only wanted to proceed to the three `toString()` equality checks on the right if we were sure that we already had a numeric length property on the left.

That said, it may only have been an optimization for ancient versions of IE. You’d have to measure to determine if it matters at all anymore.

If you want to leave the optimization in, perhaps add a comment, eg: `// Only proceed to check types if a numeric .length property is present.`

jgonggrijp: Ah, good thing I asked for a review. So it's still a tautology but *also* an optimization. ;-)

I would do some profiling, but I can't test in IE11 or Edge. So I think I'll probably just leave it in and go for the comment option.

jgonggrijp: Testing anyway.

jgonggrijp: Alright. I'm quite sure I totally overdid this, but since I took the effort, here are the conclusions of my testing.

On Node 10, `isArrayLike` and `isArray` are about equally cheap operations.

On Firefox 75, `isArray` takes nearly three times as much time as `isArrayLike`.

On Safari 13, `isArray` takes about 18 times as much time as `isArrayLike`.

Overall, it definitely seems better to keep the optimization even if you don't take old IE into account.

I ensured maximum optimization by running the hot loop millions of times on each trial and used Latin squares and systems of linear equations in order to obtain accurate numbers that could be compared. Impressively, Safari needs only about 230 CPU cycles to make a single iteration in its best scenario (plain object with optimization). Firefox can do the array without optimization scenario in about 245 CPU cycles. Taking the `length` of a string is much more expensive than taking the `length` of an array or `arguments`, perhaps because a string has to be object-wrapped first.

I don't really expect anyone to care, but script and data [here](https://gist.github.com/jgonggrijp/06644e2667466beaa32324b7b066e6a2).

Tomorrow, I'll put back the optimization and add a comment as you suggested.

jashkenas: Very cool! Thanks for the benchmarks!

PR #2841, opened 2020-04-25
Add alias merge for extend function

archbloom:
While using the `underscorejs` I was missing the function `merge`.


jgonggrijp:
@archbloom Thank you very much for taking the effort to prepare and submit a pull request.

I presume that you were missing `merge` because you were previously using Lodash, or because you were using a library that depends on Lodash and you were replacing Lodash by Underscore in your build configuration. I am happy to hear that just aliasing `merge` to `extend` solves the issue for you.

Unfortunately, the original [`_.merge` from Lodash](https://lodash.com/docs/4.17.15#merge) is not exactly the same as `_.extend`. Rather, it is a *recursive* version of `_.extendOwn`. This means that your alias will not always work for other people if they are actually depending on the recursion. For this reason, we cannot accept your pull request; it would lead to too much confusion.

However, for your own project, you can still have an Underscore that has `merge` as an alias to `extend`. It starts with a thin wrapper of the `index.js`:

```js
// your-own-index.js
export * from 'underscore/modules/index.js';
export { default, extend as merge } from 'underscore/modules/index.js';
```

Based on that, you can then make an `index-default.js` that exports a version of `_` with all functions mixed in, including your `merge` alias:

```js
// your-own-index-default.js
import { mixin } from 'underscore/modules/index.js';
import * as allExports from './your-own-index.js';

var _ = mixin(allExports);
// next line supports some old emulations of ES modules
_._ = _;

export default _;
```

You can turn the above module into an UMD bundle with Rollup. It will be identical to the official UMD bundle except that it has an extra line aliasing `merge` to `extend`.

If you are using a build tool that looks for the `module` field in the `package.json`, such as Rollup or WebPack 2+, you can finish this off with a third module in order to support both named and default exports in native ES format:

```js
// your-own-index-all.js
export * from './your-own-index.js';
export { default } from './your-own-index-default.js';
```

Closing now, but please feel free to comment or reopen.

PR #2844, opened 2020-05-02
Typechecks and isEqual support for ArrayBuffer, DataView and typed arrays

jgonggrijp:
This is an updated and polished continuation of #2694. Merging this PR will (should?) also merge the other.

@jashkenas I'm requesting your review again because I ended up making quite substantial changes compared to the original PR that you already reviewed:

- Exports new `isArrayBuffer`, `isDataView` and `isTypedArray` functions. These are also tested.
- Created an internal `isBufferLike` function that follows the pattern of `isArrayLike` (with `byteLength` instead of `length`) and factored out the common logic.
- Moved around and reformulated @maartenbreddels's logic, although I kept the principles and left his tests for `isEqual` untouched.
- Added `BigInt64Array` and `BigUint64Array`, which probably didn't exist yet in 2018.

I could not entirely avoid listing the typed array constructor names, because typed arrays appeared in IE10 while `ArrayBuffer.isView` appeared only in IE11. However, I was able to compress the list into a regular expression.


jashkenas:
This looks really nice to me! 👏

Issue #2214, opened 2015-06-12
executeBound on var arg constructors


megawac:
As mentioned earlier (#2199), `exectueBound` doesn't support all of `Function.prototype.bind` cases on constructors which take variable arguments. As the main implementation of `bind` no longer uses the native implementation, it's important we fix it.

Lodash currently supports these cases.

```js
var D = _.bind(Date, null, 2015, 05)
new D(10);
```

jdalton:
> Lodash currently supports these cases.

FWIW lodash didn't support this until 3.9.0 (last month). Not sure if that affects the priority but it's something to consider.

megawac:
Any hints to get this working jdd?

jdalton:
See [createCtorWrapper](https://github.com/lodash/lodash/blob/539b82f63369bc9fc7d08ee889969d027dfcc869/lodash.src.js#L3260-L3283). Support is a side effect of patching it to work with ES6 class constructors.

jridgewell:
I have solutions of varying degrees:

- [Revert #2199](https://github.com/jashkenas/underscore/compare/master...jridgewell:bind-nativeBind-overridable)
  - Slow creation
  - Fast invocation
  - Handles any constructor
- [Switch on Function#bind](https://github.com/jashkenas/underscore/compare/master...jridgewell:bind-nativeBind)
  - Slow (but slightly faster) creation
  - Fast invocation
  - Handles any constructor
- [JDD's `new`](https://github.com/jashkenas/underscore/compare/master...jridgewell:bind-nativeBind-new)
  - Fast creation
  - Fast invocation
  - Handles any native JS constructor
  - Won't handle ES6 class constructors with 7+ parameters

jdalton:
I'm not a fan of hybrid (obviously I dig the JDD way :P) but it looks like the `Switch on Functin#bind` is a more preferable hybrid as it forks on definition instead of invocation.

jashkenas:
Switch on bind does look slightly preferable. Creating a new switch/case pyramid of doom would be distasteful, even more so if it puts an upper limit on the number of arguments.

megawac:
The issue is the fallback for old ie will be inconsistent

megawac:
The best way to handle this may be to do as jdd decided and drop explicit support for non es5 and defer this stuff to es5-shim

jdalton:
> Creating a new switch/case pyramid of doom would be distasteful, even more so if it puts an upper limit on the number of arguments.

[...]

PR #2845, opened 2020-05-06
_.bind nontrivial constructor fix (?)

jgonggrijp:
A quick summary of what led up to this diff:

- #2197 was a complaint about `_.bind` performing worse than native `Function.prototype.bind`, due to it basically being `slice.call(arguments)` + native bind (at least at the time).
- #2199 "fixed" this by completely removing native bind out of the equation (but didn't update the comment, which has continued to claim a fallback to native bind until this very day).
- #2214 reported that this was a breaking change: `_.bind` previously worked with nontrivial constructors such as `Date` in environments that supported native bind, but not anymore. This has been the status quo since June 2015.
- In the discussion that followed, [this](https://github.com/jashkenas/underscore/compare/master...jridgewell:bind-nativeBind) was considered the most favorable solution: check for availability of native bind only at the time of definition of `_.bind`.
- I took the above solution, updated it to the current situation and added a regression test.

I'm not actually sure this should be merged. Is using a bound function as a constructor important to support? On the other hand, #2199 was probably the wrong solution to the wrong problem in the first place, so maybe this detour should never have been made.


jashkenas:
> I'm not actually sure this should be merged. Is using a bound function as a constructor important to support?

It is not important to support. Within a constructor function, `this` should be the instance you’re constructing.

I feel like we should rely on native `bind` here if there are performance improvements it gives us, and not worry about it otherwise.

jgonggrijp:
Here's the thing though. `executeBound` already has a fallthrough case for being called as a constructor (i.e. with `new`), in which case it doesn't keep the `this` binding but does keep the argument bindings:

https://github.com/jashkenas/underscore/blob/4334c126b3e71e4aeb8912d05d72139d24f7d0f1/modules/index.js#L770-L778

And it works for simple constructors, but not for fancy stuff like `Date`. Native bind has this sophistication too, but it does work for fancy constructors as well. So where to take this:

- Support some constructors but not others (current situation)?
- Support all constructors (this PR)?
- Support no constructors at all (another breaking change)?

jashkenas:
> So where to take this:
> 
> * Support some constructors but not others (current situation)?
> * Support all constructors (this PR)?
> * Support no constructors at all (another breaking change)?

I don’t believe that it’s terribly important — these sorts of tickets are originally driven by test-case, nit-picked development, not real-world use cases ... But, if it doesn’t cost us much (in bound function call speed, and in file size) to support all constructors with this PR, I think it’s fine and good to merge it!

jgonggrijp:
Time for another benchmark, then.

Issue #2846, opened 2020-05-06
Add a note about JS comparison operators


jgonggrijp:
I would like to close all three of these tickets, because I believe they are based on a misunderstanding of the `<` operator, and I'd like to address this with a note in the documentation instead:

- #1768
- #1834
- #2042

The note would be somewhere at the bottom of the page, with a link to the MDN entry on the comparison operators. Reference entries on functions like `_.sortBy` and `_.sortedIndex` would reference to this note.

@jashkenas do you agree?

jashkenas:
I agree! I don’t think that sorting `NaN` or `undefined` are an important problem to "fix" — you should *never* be sorting them.

PR #2848, opened 2020-05-06
Document the use of operator <

### LINKED ISSUE #1768: Comparator edge-case handling

### LINKED ISSUE #1834: _.sortedIndex align with _.sortBy

jgonggrijp:
@jashkenas I inserted a new "Notes" section between the links and the changelog. For now it only contains the note on operator `<`. Maybe that note is too long. Maybe the references are too short.

Implements #2846
Closes #2042
Closes #1834
Closes #1768

### DIFF
--- a/index.html
+++ b/index.html
@@ -2435,7 +2435,7 @@ _.chain([1, 2, 3]).reverse().value();
       <h2 id="notes">Notes</h2>
 
       <p id="relational-operator-note">
-        <b class="header">On using the <tt>&lt;</tt> operator in JavaScript</b>
+        <b class="header">On the use of <tt>&lt;</tt> in Underscore</b>
         <br>
         Underscore functions that depend on ordering, such as
         <a href="#sortBy"><tt>_.sortBy</tt></a> and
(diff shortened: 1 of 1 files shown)


jashkenas:
I love it, but it is a bit much 😉

Instead of such an excellent treatise on `<` and `NaN` behavior, perhaps try trimming it down to a version that simply briefly mentions that non-comparable JS values have undefined sort order behavior, perhaps with a link to good external docs on the matter, and then move on quickly to a more practical discussion about how to fix your sorting problems: By defining a custom comparator that allows you to specify a sub-key, or compute a value, by which to sort.

jgonggrijp:
I do get carried away sometimes. 😅

Better now?

jashkenas:
Looks great. The one thing I might consider adding would be what I consider the most common case of difficulties sorting in JS: sorting an array of objects or class instances. Generally, there’s a single property on the object, or a method to call, to produce the desired value for the sort iterator — but that’s not obvious to beginners!

It also makes for a good example snippet.

jgonggrijp:
I agree there is a value in a general "sorting in Underscore for beginners" type of note. However, when writing this note I specifically set out to address issues like #1768, which I suspect people still run into after they have already figured out the property shorthand thing. After all, a property that is missing in some objects in a collection is a common way in which people may end up comparing `undefined`. The documentation of `_.sortBy` et al. already mentions the option to use an iteratee, anyway.

I think a general introduction to sorting for beginners would require a different approach. I'm not opposed to changing the note in order to suit this purpose, but I want to concentrate on other things right now. Given your approval, I'll just merge this for now. We can revise it at any time.

I have however added a link to the `_.iteratee` section to give a bit more of a hint of the available options.

Issue #2325, opened 2015-10-21
dropping IE8 support


jdalton:
Tie-in issue to Backbone issue [#3828](https://github.com/jashkenas/backbone/issues/3828).

On Jan 12, 2016 MS is [essentially dropping support for IE8](https://support.microsoft.com/en-us/gp/microsoft-internet-explorer). With that in mind jQuery and Lodash are dropping IE8 support in their 3.0 and 4.0 releases. So let's discuss dropping it in the next major bump of Underscore.

akre54:
:+1: What would that entail? `Object.create` is finally supported, as are most of the Array.prototype methods, but we dropped those already.

jdalton:
It would mean dropping [hasEnumBug fixes](https://github.com/jashkenas/underscore/blob/13fe120d16af3eee2b96c4c1446e0eca7d4df9c0/underscore.js#L932-L934) in `_.keys` and `_.allKeys`, the [shift and splice fixes](https://github.com/jashkenas/underscore/blob/13fe120d16af3eee2b96c4c1446e0eca7d4df9c0/underscore.js#L1576-L1576) in the array wrapper assignments, the [isArguments fallback](https://github.com/jashkenas/underscore/blob/13fe120d16af3eee2b96c4c1446e0eca7d4df9c0/underscore.js#L1290-L1294), and simplify [isArray](https://github.com/jashkenas/underscore/blob/13fe120d16af3eee2b96c4c1446e0eca7d4df9c0/underscore.js#L1271-L1273).

akre54:
In theory I'm in favor. You probably should be using es5shim/es5sham if you need the older browsers anyways.

jridgewell:
:+1:

jdalton:
> In theory I'm in favor. You probably should be using es5shim/es5sham if you need the older browsers anyways.

That's how I'm approaching it with Lodash. I make sure that my usage has a compat path with es5shim and es6shim.

jehrenzweig:
Is there a living document somewhere that states what browsers are supported by Underscore?

jgonggrijp:
Reminder to self: `_.delay` is essentially a ponyfill for `setTimeout` in IE < 10 (where `setTimeout` doesn't support passing additional arguments), so it can be changed into just an alias to `setTimeout` when we implement this.

@jehrenzweig 

> Is there a living document somewhere that states what browsers are supported by Underscore?

Since version 1.12, there is: https://underscorejs.org/#compatibility.

Issue #2847, opened 2020-05-06
Fix the Travis build


jgonggrijp:
Preferably using GitHub Actions. See discussion in #2740.

jgonggrijp:
@jashkenas We just got a patch by @dominykas that fixed an inconspicuous indentation mistake in the `.travis.yml`. I suspect this is the reason why Travis stopped working in 2018 (see https://github.com/jashkenas/underscore/pull/2856#issuecomment-644083601). However, it appears that fixing the indentation is not sufficient to re-activate the repository; if I try to trigger a custom build on travis-ci.org, it still tells me that the repository is not active.

Would you mind a quick try whether you can activate the repository on travis-ci.org? I cannot do this on your behalf. With some luck, re-activating is all that is left to be done to get CI working again.

jashkenas:
Sure thing. I’ve logged in and reactivated [the repository](https://travis-ci.org/github/jashkenas/underscore) on Travis.

jgonggrijp:
Thanks @jashkenas! I was able to trigger a [custom build](https://travis-ci.org/github/jashkenas/underscore/builds/698687023). It failed, but hopefully at least we now have any CI at all. After I'm done with #2849, I'll see whether I can fix the build.

Keeping this open until I'm sure the build triggers automatically.

jgonggrijp:
I pushed a bunch of new commits to #2849 and Travis automatically started a new build, so it appears we'll have CI again. The build is however consistently failing over something futile, so keeping this open and changing the title to remind myself that there is still something to be done.

PR #2849, opened 2020-05-08
Chopping Underscore into lots of tiny modules

jgonggrijp:
After #2826, I'm sure some people have been waiting for this. Here it is, finally!

## Enhancements

- `modules/index.js` is split into lots of tiny modules in order to promote code reuse and to facilitate people who want to create their own bundles with an even smaller Underscore footprint.
    - Constants are named exports from the internal `modules/_setup.js`.
    - Each function has its own module of which it is the default export. Internal functions have a module name starting with `_`, e.g., `modules/_baseCreate.js`.
    - `modules/index.js` still exists, it is now a module that pulls in all the exports from the individual public function modules and exposes them as named exports with aliases. Its interface has not changed compared to the semimonolithic version.
- Removed several circular import chains by changing the internal `collectNonEnumProps` so that it doesn't depend on `_.contains` anymore. See also compatibility notes below.
- Added a monolithic ESM build for people who want to use ES import syntax in the browser. It is named `underscore-esm.js` for the time being; see compatibility notes below.
- Added modular AMD and CommonJS builds so that people using (or transpiling to) Require.js, Browserify etcetera can benefit from fast development builds and treeshaking at the same time.
- In addition to the "single read" version of the annotated source, people now have the option to read the annotated source in a modular version as well, with hyperlinked module references.
- Like the monolithic UMD and ESM bundles, the "single read" annotated source is now in strict order of dependency, which makes it potentially easier to digest.
- I have made a full polishing pass over all the comments.

@jashkenas: both versions of the annotated source are slightly different from what I have shown you before, so you may want to re-read them.

## Tradeoffs

- The source in the monolithic bundles and the "single read" annotated source is not ordered by category (collection/array/function/object/utility) anymore.
- A potential pitfall is introduced when writing comments at module scope: one cannot safely assume that the reading context before the comment and after the function following it will be the same in a monolithic build, so each comment should be written such that it makes sense when viewing only it and the function that immediately follows it.
- Another slight pitfall is that comments may disappear in a monolithic build if they are placed before an elided `import` statement. This is not a problem as long as that comment only makes sense when you can also see the `import` statement itself, but it does mean that developers should avoid writing important comments about actual logic before an `import` statement.
- The minified and gzipped UMD bundle weight increased by another 110 bytes.

## Compatibility notes

I have opted out of all breaking changes that may seem natural to combine with modularization. I did this because I want there to be a backwards-compatible modular 1.x release before we move on to a breaking Underscore 2.0 release. Having 1.x and 2.x branches that are structured in the same way makes it much easier to share bugfixes between them, so that we can continue to support the 1.x series for a while. It also simplifies development work on other libraries that may want to have separate support branches for Underscore 1.x and Underscore 2.x, such as (a revived) Underscore-Contrib.

So I have **not** made the following changes and I propose to make them in Underscore 2.0 instead:

- Make the monolithic ESM bundle the default, i.e., rename the current `underscore.js` to `underscore-umd.js` and the new `underscore-esm.js` to `underscore.js`.
- Remove `collectNonEnumProps` entirely to shake weight again (see also #2325).
- Remove the default export from `modules/index.js`. There are three stages of “completeness” of the `_` object and users should make a conscious choice as to which one they want to import, instead of ambiguously importing the middle one from the index (which may not be what they expect):
    - the bare version from `modules/underscore.js`;
    - the one with unwrapping and array methods from `modules/underscore-oop.js`;
    - the one with unwrapping and array methods and all Underscore functions mixed in from the package entry point (or equivalently, `modules/index-default.js`).

## Open questions

- Currently, the modular AMD and CommonJS builds are in the .gitignore, so they are never committed (but they are uploaded in the NPM release). They are regenerated every time you invoke `npm run bundle`, including on every commit. This adds a delay of about a second. We may want to take a slightly more sophisticated approach to this. One of the options is to put them in a separate Rollup config so that we can separate building the monolithic bundles from building the modular AMD and CJS versions. I'm open to suggestions.
- The scripts in the `package.json` are starting to get a bit unwieldly. It may be worth investigating a dedicated automation tool such as Gulp, although I think this would belong in a new pull request.
- I have wrapped the `ctor` in `modules/create.js` in order to reduce side effects. However, I'm not sure this is actually necessary. We may be able to shave off a few bytes by undoing that.

## Next steps

- If (when) we release this, I think it would be worthwhile to publish a blog post about modular Underscore, with a discussion on the various ways in which one may import from Underscore and some recommendations (such as using monolithic from a CDN unless you have special needs).
- I'm planning on another big non-breaking change right after merging this to enhance the functional style of the implementation. I expect that this will shake a lot of weight and that it will help future enhancements.

**Edit to add:** You can support my work on Underscore and other open source projects [on Patreon](https://www.patreon.com/juliangonggrijp).


docs/linked-esm.jst
```diff
@@ -0,0 +1,107 @@
+<% /*
+  This is an extended version of the resources/parallel/docco.jst that is
+  bundled with Docco 0.8.0.
+  The license of the original file is available over here:
+  https://github.com/jashkenas/docco/blob/master/LICENSE.
+  The extension adds hyperlinking to javascript import statements.
+*/ %>
```
jashkenas, line 7: I’d be happy to merge a PR over in Docco to add hyperlinking to import statements...

jgonggrijp: More than happy to do so. This comes in two mutually exclusive flavors, though, which is the reason I haven't done this yet:

1. The quick and dirty way, which monkeypatches the hyperlinks into the page post-rendering using jQuery and which only works when the import syntax and path resolution are like in JavaScript. This is what I've done here.
2. A clean but much slower and more difficult way, which should potentially work for all programming languages that Docco supports. This would require not only patching Docco, but also patching highlight.js, since the output of highlight.js currently offers no portable way to detect module references.

Which would you prefer?

jgonggrijp:
> Okay, you have my blessing! 👑

Woooo! Thanks, this means a lot to me.

> My one request would be that we keep a single link in the website sidebar to the old, single-file full source version of 1.10.2, as "Underscore Classic". I’ve heard from a lot of folks over the years who have cut their teeth on JavaScript by reading and trying to understand that file, and I’d like to not have that experience lost in the sands of time...

In case you missed it, the proposal as it currently stands has a link in the website sidebar to (a docco rendered version of) [this monolithic build](https://github.com/jashkenas/underscore/blob/9e3117efcd9a58bcac39823ab65b9f4505239785/underscore-esm.js). The sidebar refers to this as the "single read" annotated source. It is not sorted by category (collection, array etcetera), but it is strictly ordered by dependency and it does have all the comments. Don't you think this can fullfill the same need?

> (If you need my help with anything specific, in terms of website stuff or cutting releases, just let me know.)

Thanks! I will get back to you about this.

jashkenas:
> In case you missed it, the proposal as it currently stands has a link in the website sidebar to (a docco rendered version of) [this monolithic build](https://github.com/jashkenas/underscore/blob/9e3117efcd9a58bcac39823ab65b9f4505239785/underscore-esm.js). The sidebar refers to this as the "single read" annotated source. It is not sorted by category (collection, array etcetera), but it is strictly ordered by dependency and it does have all the comments. Don't you think this can fullfill the same need?

Not entirely — I think it’s great, but that you lose quite a bit in what’s possible to do with explanatory prose, when you lose control of the ordering of the paragraphs. People also like to just view the `.js` source file directly in their favorite text editor. But maybe don’t worry about this request — I’d like to brush up and fix a lot of the overgrown comments in any case, and we can always add this at a later date.

(They used to be clearer, back at the beginning: https://cdn.rawgit.com/jashkenas/underscore/0.4.0/underscore.js)

docs/linked-esm.jst
jashkenas: Ah, I see — in that case, it seems best to leave it as you have it in this PR.

jgonggrijp:
> I think (...) that you lose quite a bit in what’s possible to do with explanatory prose, when you lose control of the ordering of the paragraphs.

Not to counter your point (I agree this is a drawback), but I should mention that we can still influence the order by reordering the imports. It's not full control, but still some.

> People also like to just view the `.js` source file directly in their favorite text editor.

This is possible with either monolithic bundle, no problem.

> But maybe don’t worry about this request —

I'm not worried. I value your opinion a lot and I don't mind delaying the merge if I can make the changes (even) more satisfying.

Although I should mention that I'd prefer fixing the comments in the new monolithic bundles over including a link to an outdated version of Underscore.

> I’d like to brush up and fix a lot of the overgrown comments in any case, and we can always add this at a later date.
> 
> (They used to be clearer, back at the beginning: https://cdn.rawgit.com/jashkenas/underscore/0.4.0/underscore.js)

I see what you mean. I could make another pass over the source in an attempt to approach that quality more, but maybe you want to do it yourself?

As an aside, version 0.4.0 might be nice as a superlative of "classic", i.e., "archeological".

jashkenas:
> Although I should mention that I'd prefer fixing the comments in the new monolithic bundles over including a link to an outdated version of Underscore.

You’re right — it would be much better if we improved the monolithic current comments, instead of linking to an old and dusty shelf copy.

The best way to do this would be to make a tool that allows us to order the monolithic builds in any arbitrary ordering that we choose ... but the second-best way to do that would be to re-write the comments in such a way that they flow naturally based on the actual ordering present in the monolithic build. That’s hard, given that they need to make sense in isolation, without reference to anything surrounding them ... and also make sense when read from top to bottom (this is the concern that you mention in your issue description above), but I guess that’s the challenge that presents itself in front of us.

modules/isEqual.js
```diff
@@ -0,0 +1,136 @@
+import _ from './underscore.js';
+import { toString, SymbolProto } from './_setup.js';
+import getByteLength from './_getByteLength';
+import isTypedArray from './isTypedArray';
```
cambecc, line 4: missing .js suffix for isTypedArray and getByteLength imports

modules/template.js
```diff
@@ -0,0 +1,86 @@
+import defaults from './defaults.js';
+import _ from './underscore.js';
+import './templateSettings';
```
cambecc, line 3: missing .js suffix here

jgonggrijp:
Quick update: I'm currently in the process of analyzing the dependency graph of the modules, in order to choose the most comfortable import order.

modules/index.js
```diff
@@ -3,1729 +3,142 @@
...
+export { default as findKey } from './findKey.js';
+export { default as findIndex } from './findIndex.js';
+export { default as findLastIndex } from './findLastIndex.js';
+export { default as sortedIndex } from './sortedIndex.js';
+export { default as indexOf } from './indexOf.js';
+export { default as lastIndexOf } from './lastIndexOf.js';
+export { default as find, default as detect } from './find.js';
+export { default as findWhere } from './findWhere.js';
+
+// Collection Functions (plus `omit`).
```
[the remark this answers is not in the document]
jashkenas: Perhaps: "Collection functions work on both arrays and objects" (and Maps and Sets if/when we choose to support them)

modules/index.js
```diff
@@ -3,1729 +3,142 @@
...
+export { default as throttle } from './throttle.js';
+export { default as debounce } from './debounce.js';
+export { default as wrap } from './wrap.js';
+export { default as negate } from './negate.js';
+export { default as compose } from './compose.js';
+export { default as after } from './after.js';
+export { default as before } from './before.js';
+export { default as once } from './once.js';
+
+// Find et al.
```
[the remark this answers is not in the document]
jashkenas: Perhaps: "Finders: find values and positions of elements in arrays"

modules/index.js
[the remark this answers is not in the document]
jgonggrijp: I should have specified what I was unhappy about. I'm fine with the "Collection Functions" part but not with the "(plus `omit`)" part. I felt compelled to append that because `omit` is out of place here, but it feels ugly.

I do like your suggestions to clarify the section titles a bit, though. I'll follow up on them.

jgonggrijp:
Glad to hear it passes the bar for you!

Good catch about those comments. I think I'll embrace Rollup's "feature" of copying only the first line of a module's opening comment; if I rewrite that line so that it makes sense in isolation, it may help clarify what is going on at the bottom of the monolithic bundle.

jashkenas:
Maybe give me a sense of what you’d like from me for another review? A read through and testing of the diff again? Or just a last minute thumbs up to recognize that we’re still on the same page?

jgonggrijp:
@jashkenas Just a thumbs-up, mostly for the `index.js` which has changed quite a bit since your last review. I'm especially wondering whether you're OK with the way I'm introducing the categories.

jgonggrijp:
Thanks a lot Jeremy!

jgonggrijp:
Boom, Underscore is completely modular from now on! I will work towards a new release next, but before that, I want to do a couple of things:

- [ ] Fix the Travis build (#2847/#2857) in order to not needlessly worry people who pay attention to that.
- [ ] Write an article/blog post about the best ways to take advantage of modular Underscore as well as its monolithic builds.

jashkenas:
Hip hip hooray! Congratulations, @jgonggrijp! The longevity of open source meaningful, positive contribution continues to surprise and impress.

PR #2856, opened 2020-06-15
fix: typo in .travis.yml

dominykas:
When I pass the current version through js-yaml, it tells me that the whitespace there is unexpected.

Could this be why the [CI no longer works](https://github.com/jashkenas/underscore/issues/2847)?

Then again, this might just need a migration to travis-ci.com, etc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,10 @@ node_js:
   - "6.14.1"
   - "8.11.1"
   - "9.11.1"
- matrix:
-   include:
-     - node_js: "4.0"
-       env: BROWSER=true
+matrix:
+  include:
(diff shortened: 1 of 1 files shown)


jgonggrijp:
Thanks for the fix, @dominykas!

Also thanks for the tip of migrating to travis-ci.com, I didn't think of that. From what I can find about it, however, it doesn't seem to be necessary; travis-ci.org is also still operational.

After some more research, I think your patch might very well be sufficient. The last commit for which Travis still worked was 1dfec11447b56aa0 (31. May 2018). The commit directly after that, 2e57316, introduced the misindentation by incorrectly uncommenting the lines in question.

Issue #2762, opened 2018-05-30
Debounce function documentation


carpben:
According to the documentation at http://underscorejs.org/docs/underscore.html#section-86 the debounce function: 
> Returns a function, that, as long as it continues to be invoked, will not be triggered. 

I think this phrasing is a bit misleading, as the function that won't be triggered is the first parameter it recevies.

jashkenas:
Want to send a PR that improves it, and get the credit for the change?

carpben:
Yes I would, thanks. I'll post a suggestion later this week.

jgonggrijp:
@carpben Friendly reminder that you are still welcome to submit a pull request and get the credit.

carpben:
@jashkenas @jgonggrijp , Thanks for the patience, and sorry for the delay on my part. 

I submitted https://github.com/jashkenas/underscore/pull/2858.

jgonggrijp:
Was merged, closing.

PR #2858, opened 2020-06-23
Updating debounce documentation

carpben:
This PR tries to solve https://github.com/jashkenas/underscore/issues/2762 and updates the definition/documentation of the debounce function. 

I'm not a native English speaker, so the wording might need to be adjusted. However, I think that talking in terms of sequence can make things easier to understand.


carpben:
Thanks @jgonggrijp ! 

I've noticed that #2849 was merged, so I'd like to continue with option 2. However, I don't understand why `npm install` is required in this context - I don't need to run any code, just update the comment. Am I missing anything?

jgonggrijp:
Welcome!

The devDependencies include `husky`, which will run pre- and post-commit hooks to ensure that monolithic builds and corresponding sourcemaps are created which accurately reflect the content of the modules. I should probably patch the `CONTRIBUTING` to reflect this.

carpben:
@jgonggrijp, By mistake the two commits above were made with my company's computer. Lol. Are we all set?

jgonggrijp:
As far as I'm concerned, we're all set. Are you sure you're comfortable with your company account taking the credit, though? I wouldn't object to a force-push so you can rectify the commit attribution. See https://stackoverflow.com/questions/3042437/how-to-change-the-commit-author-for-one-specific-commit if you don't know how to change the attribution.

carpben:
There are 3 commits on my private user. So if the difference is that there will be 3 commits on my user instead of 5 I really don't mind. 
I tried `git rebase -i` as suggested in the  SOf you sent, but it didn't work well on my device (windows, created many conflicts and didn't work as expected). I don't mind resetting current branch to the previous commit, and redoing those 2 commits.

jgonggrijp:
So which route do you prefer? Merge as-is, or redoing those last two commits?

carpben:
I prefer we merge as is.

jgonggrijp:
Thanks for your contribution @carpben!

PR #2859, opened 2020-06-30
fix: isarraylike except function type

jqoker:


jgonggrijp:
Duplicate of #2830. See discussion there.

PR #2862, opened 2020-07-07
find and replace usage of blacklist and whitelist with disallow and allow

UseWIthCaution:
This issue aims to bring the underscore codebase and documentation in line with the [Code of Conduct. ](https://github.com/jashkenas/underscore/blob/master/CODE_OF_CONDUCT.md)

Short: Provide terminology to package users that does not perpetuate white supremacy and racism against Black people

Long: There has been a much-needed shift in the tech industry to transition our naming conventions away from problematic racist terminology. For example, [Github is replacing the convention of having a master branch to remove any master/slave reference. ](https://www.cnet.com/news/microsofts-github-is-removing-coding-terms-like-master-and-slave/)

### DIFF
--- a/docs/underscore-esm.html
+++ b/docs/underscore-esm.html
@@ -2725,7 +2725,7 @@ The opposite of object.</p>
               <div class="pilwrap ">
                 <a class="pilcrow" href="#section-122">&#182;</a>
               </div>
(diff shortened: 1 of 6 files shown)


jashkenas:
Works for me. Thanks for sending the patch!

PR #2868, opened 2020-07-27
Enable flattening to a specified depth (continued)

jgonggrijp:
This is a continuation of #2460 by @paulfalgout. The former boolean `shallow` parameter of both the internal and the external `flatten` is replaced by a `depth` parameter which can be either boolean or numeric. Boolean values still have the same meaning. Numeric `1` has the same meaning as `true` while `0` or less results in a shallow copy without any flattening. The default is still infinite depth.

@jashkenas The original PR was approved by @akre54 and @michaelficarra in 2016. If this is sufficiently reassuring for you, you can stop reading here.

@paulfalgout I moved your checks against nonnumeric and nonpositive `depth` values back from the public to the internal `flatten`. At the time, you were concerned that this would be inconsistent with `strict === true`, but there is actually no conflict. When `depth === 1`, every *nested* element of every array is copied whether the nested element is an array or not, even when `strict === true`. Extrapolating this to `depth === 0`, "every array" is just the top-level array, so every element of it should be copied, regardless of whether it is an array and regardless of the value of `strict`.

Merging this should automatically also merge #2460.


jgonggrijp:
Thanks for the review @jashkenas!

PR #2870, opened 2020-08-01
Prepare 1.11 release

jgonggrijp:
Before I submit this for review, I want to insert a link to the article in two places. I'm holding off on this until all reviews of the article are in.

This is a much larger update than I realized. I thought it was mostly the modularization, and it is, but all the other enhancements and fixes still add up to a very substantial update.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2542,6 +2542,12 @@ _.chain([1, 2, 3]).reverse().value();
             bound object.
           </li>
           <li>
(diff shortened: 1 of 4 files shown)


jgonggrijp:
@jashkenas Ready for review!

In case you agree 100% with the changes: please don't release just yet, because I'm still waiting for a final bit of feedback on the article.

PR #2876, opened 2020-09-21
Improvements to _.isEqual comparisons for typed arrays and DataView

### LINKED ISSUE #2875: isEqual not producing correct result when comparing array of Buffers  in 1.11.0
- version underscore@1.11.0

```javascript
const _u = require("underscore");
const _l = require("lodash");

const a = [Buffer.from([1, 2, 3]), Buffer.from([1, 2, 3])];
const b = [Buffer.from([4, 5, 6]), Buffer.from([4, 5, 7])];

console.log(_u.isEqual(a, b)); // True => Wrong
console.log(_l.isEqual(a, b)); // False => Correct
```

- version  underscore@1.10.2 was oK

jgonggrijp:
- Fixes #2875 by taking the `.byteOffset` into account when comparing two typed arrays.
- Factors out the internal `toDataView` function as the single source of truth on how to correctly cast (or shallow copy) buffer-like objects to `DataView`.
- Adds an optimization to avoid byte-wise comparisons between `DataView`s that have identical buffer, offset and length.

@carpben @mafiosso @sktguha @yashshah1 I would welcome a review by one of you.

(@jashkenas you are of course also welcome to review. I'm however addressing other reviewers because I know your time is scarce and because the change doesn't seem dramatic enough to require your attention.)


sktguha:
Thanks for the mention @jgonggrijp 😊😊. Will check in the morning

sktguha:
@jgonggrijp two comments
1) I could be wrong, but it seems Buffer exposes a .equals method
so the issue given in #2875 could simply be fixed by using the .equals method given by default.
i.e 
const a = [Buffer.from([1, 2, 3]), Buffer.from([1, 2, 3])];
const b = [Buffer.from([4, 5, 6]), Buffer.from([4, 5, 7])];
const c = [Buffer.from([1, 2, 3]), Buffer.from([1, 2, 3])];

const isEqual = (a,b) => {
  if(a.length !== b.length) return false;
  return a.every((aElem,i)=> aElem.equals(b[i]));
}
console.log(isEqual(a,b)); //return false -> Correct
console.log(isEqual(a,c)); //returns true -> Correct

//of course this assumes both arrays have only Buffer, but you get the idea

//This approach also fixes your regression test that you have added in this PR, [https://github.com/jashkenas/underscore/pull/2876/commits/20a39e75c755d3d17142f7bea0b931edec0816d4](https://github.com/jashkenas/underscore/pull/2876/commits/20a39e75c755d3d17142f7bea0b931edec0816d4)

var shared = new Uint8Array([1, 2, 3, 4]);
      var view1 = new Uint8Array(shared.buffer, 0, 2);
      var view2 = new Uint8Array(shared.buffer, 2, 2);

console.log('reg test', Buffer.from(view1).equals(Buffer.from(view2))) // returns false, which is the correct result acc to your test

2) Also I found an interesting comment on stackoverflow regd performance of such tests
"You can also further optimise the test by comparing an aligned prefix (e.g., given 131 bytes, comparing the first 131 bytes as longer words then comparing the final three bytes one-byte at a time)."
Here is the link:  https://stackoverflow.com/questions/21553528/how-to-test-for-equality-in-arraybuffer-dataview-and-typedarray#comments-52181275:~:text=You%20can%20also%20further%20optimise%20the,three%20bytes%20one%2Dbyte%20at%20a%20time).

So if it requires any code changes, please let me know . As it would be Really Cool 😎😎😎 for me to push code to underscore itself.

jgonggrijp:
Thanks for the very interesting feedback @sktguha!

You are more than welcome to contribute code. If you agree that my code could be faster, but is still correct, I suggest that we merge this first and that you then submit another PR with optimizations. Would that order work for you?

sktguha:
Sure Julian, that does sound good
But I am wondering is that if we have the .equals readymade method available, do we really need to convert to ArrayView as it is being done currently
One case where we might still need the conversion is that the .equals method is not supported but the ArrayView support is there. Not sure if that will ever be the case as I think it's a recent feature and it's nodejs , not browser environment, so there kind of things maybe are not present.

Also regards the performance comparison between . equals method and conversion to ArrayView , let me make a jsperf. But not sure if jsperf and other tools support nodejs. Otherwise let me test on local machine.

sktguha:
@jgonggrijp But ok if its an important fix we can go ahead and merge. ya later I can make another PR for optimisation. 
looks good to me otherwise (LGTM)

jgonggrijp:
Performance testing is definitely a good idea.

I'll merge this now so you can use the `master` branch as your base.

sktguha:
so can we discuss about future improvements here or I should create a new issue for the discussion ?

jgonggrijp:
Here is fine, or you can already create a draft PR for your submission.

sktguha:
sure. let me think of the things to do in the new PR ,and come back

PR #2877, opened 2020-09-21
Path customization

jgonggrijp:
This is the first follow-up on https://github.com/documentcloud/underscore-contrib/pull/231. The next (and final) follow-up will be a PR to Contrib.

I should emphasize that *the behavior of paths in Underscore stays the same*. That is, full-blown array paths (`['a', 0, 'b']`) remain supported while shorthand string paths (`'a.0.b'`) remain **un**supported.

The purpose of this PR is to add a customization point, `_.toPath`, which users can override, similar to `_.iteratee`. The most obvious application of such an override would be to enable string paths, though in principle, users can do something completely different. `_.toPath` is used internally throughout Underscore.

The follow-up PR to Contrib will add a function to its public interface, which users can optionally use to override `_.toPath`. Using that optional override will cause all functions in Underscore *and* Contrib to support string paths in a uniform way. As long as users *don't* do this, all functions in Underscore *and* Contrib will only support array paths. This will be a slightly breaking change for Contrib, but it will be a major step up in consistency between the libraries as well as within Contrib. The future Fusion library will use this mechanism, too, in order to support legacy code that depends on Lodash. Assuming the PR gets approved, of course.

This PR also adds one other function, `_.get`, which could be regarded a dual to `_.has` as well as a simpler counterpart to `_.result` that doesn't invoke methods. This function makes it easier to inherit Underscore's deep property extraction logic in Contrib and Fusion.

Some more functional relations:

```js
_.get = _.compose(deepGet, _.toPath); // if we ignore _.get's optional third argument
_.propertyOf(obj) = _.partial(_.get, obj);
_.property(path) = _.partial(_.get, _, path);
```

@jashkenas Please let me know if anything is unclear.

@joshuacc FYI.


jashkenas (approved):

jgonggrijp:
Thanks for the quick review and approval, @jashkenas!

joshuacc (commented):
Looks great!

PR #2885, opened 2020-10-21
Modified Some Functions

stephenas:


captbaritone:
I don't see any value in this change. I'll close for now. If there's something valuable in this change that I'm missing please provide some description of the benefit and we can reconsider.

jgonggrijp:
Thanks for stepping in, @captbaritone. Nice surprise to hear from you!

PR #2888, opened 2020-11-15
Test: Switch to built-in QUnit CLI

Krinkle:
- Update from QUnit 2.6 to QUnit 2.10 and remove qunit-cli package.
  This drastically reduces the number of (indirect) dev dependencies for this project. 🙂 

- Use the built-in `qunit` command directly.
  It supports the same glob argument as qunit-cli, although I've simplified t to pass the directory instead for simplicity.

- Reference the qunit.js file from the node_modules directly in `index.html` so that the same version is used by both Nod and the browser test.

- Remove reference to qunit-extras.js file. This file has not existed for a while. Karma has been tolerating its absence:
  `WARN [watcher]: Pattern "/home/travis/build/jashkenas/underscore/test/vendor/qunit-extras.js" does not match any file.`


jgonggrijp (approved):
Looks like a clear improvement, thanks for contributing!

PR #2891, opened 2020-12-05
Fix a bug in _collectNonEnumProps.js

anfengchen:
As hash inherit Object.prototype, when we call "contains", property keys of Object.prototype will return true. This isn't our expectation. Using "hash[key]===true" can exclude these keys.


jgonggrijp:
Thanks for going through the trouble to fix this, @anfengchen. Why did you close the PR?

PR #2892, opened 2020-12-05
Reimplementing timer optimization  #1269

kritollm:
Restore the optimization #1269


modules/debounce.js
```diff
@@ -1,30 +1,30 @@
...
-
-  var debounced = restArguments(function(args) {
-    if (timeout) clearTimeout(timeout);
-    if (immediate) {
-      var callNow = !timeout;
-      timeout = setTimeout(later, wait);
-      if (callNow) result = func.apply(this, args);
+  var timeout, timestamp, args, result, context;
+  var later = function() {
+    var last = now() - timestamp;
```
jgonggrijp, line 10: I think the name `last` suggests a point in time instead of a duration. Perhaps call it `passed` instead.

modules/debounce.js
```diff
@@ -1,30 +1,30 @@
...
     } else {
-      timeout = delay(later, wait, this, args);
+      timeout = null;
+      if (!immediate) result = func.apply(context, args);
     }
-
+  };
+// Remove timer for immediate, samme as for throttle
+  var debounced = function() {
+    var callNow = immediate && !timeout;
```
jgonggrijp, line 20: This variable can be avoided, see later comment.

modules/debounce.js
```diff
@@ -1,30 +1,30 @@
...
+    context = this;
+    args = [].slice.call(arguments, 0);
+    timestamp = now();
+    if (!timeout) timeout = setTimeout(later, wait);
+    if (callNow) result = func.apply(context, args);
```
jgonggrijp, line 25: The function body evaluates `!timeout` twice. This can be avoided, together with the `callNow` variable, by nesting the conditionals as follows:

```suggestion
    if (!timeout) {
      timeout = setTimeout(later, wait);
      if (immediate) result = func.apply(context, args);
    }
```

This also strengthens the visual similarity to the `later` function body.

modules/debounce.js
```diff
@@ -1,30 +1,30 @@
...
+      timeout = null;
+      if (!immediate) result = func.apply(context, args);
     }
-
+  };
+// Remove timer for immediate, samme as for throttle
+  var debounced = function() {
+    var callNow = immediate && !timeout;
+    context = this;
+    args = [].slice.call(arguments, 0);
```
jgonggrijp, line 22: This expression can't be minified at all. Please put back the `restArguments` wrapper instead.

jgonggrijp (changes requested):
Thanks for this nice optimization!

Please remove the changes to `_.throttle` and follow up on my comments below. You're welcome to submit a separate PR for `_.throttle` as discussed in #2832.

modules/debounce.js
kritollm: Totally agree, I change to passed

modules/debounce.js
kritollm: Done.

modules/debounce.js
kritollm: Done.

modules/debounce.js
kritollm: I have.

modules/debounce.js
```diff
@@ -1,35 +1,41 @@
...
+  var timeout, previous, args, result, context;
+
   var later = function() {
-    var last = now() - timestamp;
-    if (wait > last) {
-      timeout = setTimeout(later, wait - last);
+    var passed = now() - previous;
+    if (wait >= passed) {
+      timeout = setTimeout(later, wait - passed);
+
```
kritollm, line 15: This blank line should be removed.

modules/debounce.js
```diff
@@ -1,35 +1,41 @@
...
-  var timeout, result;
+  var timeout, previous, args, result, context;
 
-  var later = function(context, args) {
-    timeout = null;
-    if (args) result = func.apply(context, args);
-  };
+  var later = function() {
+    var passed = now() - previous;
+    if (wait >= passed) {
```
jgonggrijp, line 13: Shouldn't this be `>`? If `wait == passed`, I'd think you want to invoke immediately instead of setting another timeout.

modules/debounce.js
jgonggrijp: Sure, go ahead.

modules/debounce.js
```diff
@@ -1,35 +1,41 @@
...
 
   debounced.cancel = function() {
     clearTimeout(timeout);
-    timeout = null;
+    timeout = args = context = null;
   };
 
   return debounced;
-}
+}
```
jgonggrijp, line 41: Please put back the newline at the end of the file. I know editors are nowadays often configured to strip them, but it is terminal-friendlier and concatenation-friendlier to always have a newline at the very end of a text file.

modules/throttle.js
```diff
@@ -44,4 +44,4 @@ export default function throttle(func, wait, options) {
   };
 
   return throttled;
-}
+}
```
jgonggrijp, line 47: Same here.

jgonggrijp (changes requested):
Thanks for adressing my comments. I think we're nearly ready to merge, just a few more details.

modules/debounce.js
kritollm: Sorry, I thought it was like that original. I prefer newline at the end of the file.

jgonggrijp: No problem!

modules/debounce.js
kritollm: You're absolutely right.

modules/throttle.js
kritollm: yes

modules/debounce.js
kritollm: I have made a version in parallel with this that does not use timeout when immediate == true, so it confused me, since I there had the opposite order in the if else block. Good thing you catched it!

modules/debounce.js
```diff
@@ -1,34 +1,39 @@
...
-    timeout = null;
-    if (args) result = func.apply(context, args);
+  var later = function() {
+    var passed = now() - previous;
+    if (wait > passed) {
+      timeout = setTimeout(later, wait - passed);
+    } else {
+      timeout = null;
+      if (!immediate) result = func.apply(context, args);
+      // This check is needed because the func can recursively invoke debounced
```
jgonggrijp, line 18:
```suggestion
      // This check is needed because `func` can recursively invoke `debounced`.
```

jgonggrijp (approved):

Issue #2448, opened 2016-02-27
iterator support for _.toArray


jdalton:
I think support for converting iterators to arrays would be a nice to have for `_.toArray`.

anthony-redFox:
@jdalton Do you mean [Array.from](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/from) ?

jdalton:
I mean something like [iteratorToArray](https://github.com/lodash/lodash/blob/4.5.1-npm/_iteratorToArray.js) which can be used as part of [toArray](https://github.com/lodash/lodash/blob/4.5.1-npm/toArray.js#L41-L55). Browsers like IE11 have support for `Map` and `Set` without support for `Array.from`.

akre54:
Rather than reimplementing `toArray` why don't we just call `Array.from` if it exists? You probably wouldn't / shouldn't be using iterators if you haven't polyfilled `Array.from`.

jdalton:
> You probably wouldn't / shouldn't be using iterators if you haven't polyfilled `Array.from`.

Because `Array.from` isn't a requirement for iterator conversion and there are environments without it were maps and sets and other iterables work fine. So it would be inconsistent support.

jdalton:
The lines savings aren't worth the inconsistent support. We're talking about 4 or 5 LOC. Nothing major.

jdalton:
Something like less than 100 bytes.

jdalton:
IE11's market share for reference:
<a href="https://www.netmarketshare.com/browser-market-share.aspx?qprid=2&qpcustomd=0"><img alt="marke tshare" width=80% src="https://cloud.githubusercontent.com/assets/4303/13399664/55b23264-deb9-11e5-8e2a-3504e3ebf986.png"></a>

jgonggrijp:
Will only make sense if combined with #2147.

PR #2894, opened 2020-12-17
iteratorToArray Support

matinFT:


modules/isIterable.js
```diff
@@ -0,0 +1,7 @@
+function isIterable(obj) {
```
jgonggrijp, line 1: If you intended `isIterable` to be a public function (i.e., mixed into the `_` object), then you should list it in `modules/index.js`. There should also be a comment before the function to explain its purpose. Take a look at a few other public modules to get a taste of how we write these comments.

If you intended `isIterable` to be only an internal function instead, then you should start the name of the module with a leading underscore, i.e., `modules/_isIterable.js`. In this case it would still be desirable to have a comment that explains the purpose of the function, but I'll readily admit that there are currently more internal functions where this is missing.

jgonggrijp, line 1: Also, you forgot `export default`.

modules/isIterable.js
```diff
@@ -0,0 +1,7 @@
...
+    // checks for null and undefined
```
jgonggrijp, line 2: This is an unnecessary comment.

modules/isIterable.js
```diff
@@ -0,0 +1,7 @@
...
+    if (obj == null) {
+      return false;
+    }
+    return typeof obj[Symbol.iterator] === 'function';
```
jgonggrijp, line 6: We can't safely assume that `Symbol` or `Symbol.iterator` is defined, because that will (needlessly!) break compatibility with older environments that only know ES3. Instead, we would use feature detection and then create an internal `getIterator` function that simply returns `undefined` in environments where these objects don't exist. You can see an example of this practice in `modules/keys.js`, which uses `Object.keys` but only if it exists.

A similar, but more subtle, consideration applies to `typeof x === 'function'`. In some environments, this expression doesn't work as intended. This is why you should use `_.isFunction` instead whenever possible. `isFunction(x)` can also be minified much more compactly than `typeof x === 'function'`.

So this line should have read like this:

```suggestion
    return isFunction(getIterator(obj));
```

jgonggrijp, line 6: Given my previous comment, the entire body of this function would actually fit into a single expression:

```suggestion
    return obj != null && isFunction(getIterator(obj));
```

Though `getIterator` should probably check against `null` and `undefined` anyway, so you could actually leave out the null check entirely.

modules/iteratorToArray.js
```diff
@@ -0,0 +1,9 @@
+export default function iteratorToArray(iterator) {
```
jgonggrijp, line 1: Like with `isIterable`, you should either add a descriptive comment and list this function in `modules/index.js` if you intended this as a public function, or start the module name with `_` otherwise. However, since this function seems to be purely an implementation detail of `_.toArray`, it should probably be internal, and maybe it should even be defined inside `modules/toArray.js`.

modules/toArray.js
```diff
@@ -5,7 +5,8 @@ import isArrayLike from './_isArrayLike.js';
 import map from './map.js';
 import identity from './identity.js';
 import values from './values.js';
-
```
jgonggrijp, line 8: This blank line should stay.

modules/toArray.js
```diff
@@ -16,5 +17,6 @@ export default function toArray(obj) {
     return obj.match(reStrSymbol);
   }
   if (isArrayLike(obj)) return map(obj, identity);
+  else if (isIterable(obj)) return iteratorToArray(obj[Symbol.iterator])
```
jgonggrijp, line 20: When adding a feature to Underscore, you always also have to add [tests](https://github.com/jashkenas/underscore/tree/master/test) to verify that the feature works as intended. This is especially important because this enables us to notice when a later change breaks a pre-existing feature.

I'm mentioning this here because `_.toArray` is already a public function. If `isIterable` or `iteratorToArray` was public as well, though, you'd have to add tests for that, too.

jgonggrijp (commented):
Thanks @matinFT for taking the effort to submit a PR!

Unfortunately, I have to reject this particular PR for reasons I'll explain below. However, I'd like to encourage you to submit PRs again in the future. For this reason, I've decided to review your changes in detail, so you can already learn about some of the things that you need to take into account when contributing to Underscore. You'll find my comments after this post.

The reason that I'm rejecting this PR, is that Underscore needs to be consistent. If we're going to support iterables, then we should support them throughout the library rather than in just one function; in fact, iterables should become one of the main collection types next to array, object, map and set. This means that adding iterable support is a big breaking change. However, we recognize that it is very desirable to have, so it is planned for Underscore version 2.0. I dedicated #2147 as the central ticket for this plan. I previously mentioned the need for wholesale iterable support in https://github.com/jashkenas/underscore/issues/2448#issuecomment-626240272.

A similar PR that was rejected with similar reasoning is #2808.

Issue #2688, opened 2017-07-09
_.max and _.min may behave incorrectly if iteratee returns infinite number


jpli:
The iteratee "swaps" the contents of the array, so `_.max` should return the smaller member:

    _.max([2, 5], function (value) { return value === 5 ? 2 : 5; });  // 2, correct

When the smaller member is `-Infinity`:

    _.max([-Infinity, 5], function (value) { return value === 5 ? -Infinity : 5; });  // 5, wrong

Similar for `_.min`:

    _.min([10, 5], function (value) { return value === 5 ? 10 : 5; });  // 10, correct
    _.min([Infinity, 5], function (value) { return value === 5 ? Infinity : 5; });  // 5, wrong

I know which part of the code causes the bug, for example in `_.min`, it's the conditional expression after `||`:

    if (computed < lastComputed || computed === Infinity && result === Infinity) {
        ...

But I don't know how to fix it because I can't figure out the exact intent of that expression.

anvyne:
I think that the conditional expression after `||` is used to favor some element of the collection over the default `Infinity` result. 

For `_.min`, `Infinity` is the default result. It is preferred that some element from the collection is returned instead of the default value, even if its computed value is also `Infinity`.

jgonggrijp:
I would like to fix this by generalizing `min` and `max` so they also work for strings and they return `undefined` for empty collections. Below is just a draft to illustrate the principle, as a proper implementation should invoke the `iteratee` only once for each element and it should also pass the key and the collection when doing that:

```js
function max(collection, iteratee, context) {
    if (isEmpty(collection)) return;
    iteratee = cb(iteratee, context);
    return reduce(collection, function(memo, value) {
        return iteratee(value) < iteratee(memo) ? value : memo;
    });
}
```

This would be a breaking change.

jgonggrijp:
Update: I have a "compromise" fix on a private work-in-progress branch, which implements the generalization I mentioned above and then jumps through a few additional hoops in order to keep it a non-breaking change. I'll probably publish that branch somewhere in the next month or so.

PR #2561, opened 2016-07-09
Note in documentation that non-numerical values don't work for max/min

jdelbello:
Documentation addition for #2552!

PR #2896, opened 2020-12-18
_.max()  should compare between numbers

anderlaw:
As mentioned in the underscore document, _.max function should compare items by numbers only

### DIFF
--- a/index.html
+++ b/index.html
@@ -821,7 +821,7 @@ _.pluck(stooges, 'name');
         function is provided, it will be used on each value to generate the
         criterion by which the value is ranked. <i>-Infinity</i> is returned
         if <b>list</b> is empty, so an <a href="#isEmpty">isEmpty</a> guard
-        may be required. Non-numerical values in <b>list</b> will be ignored.
+        may be required. This function can currently only compare numbers reliably.
         This function uses operator <tt>&lt;</tt>
         (<a href="#relational-operator-note">note</a>).
       </p>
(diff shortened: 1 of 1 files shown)


jgonggrijp (changes requested):
Thanks @anderlaw for taking the effor to submit a PR. You are right to point out that there is a mismatch between the documentation and the actual logic. Good catch!

However, I think you need to take a different approach. The documentation is broken, not the logic. I did some research, and found that the implementation has never ignored non-numeric values; rather, the false claim that it does so was added to the documentation in #2561 in a failed attempt to make the documentation more accurate.

Of course, you *could* decide to update the logic so that it reflects the documentation, but I think updating the documentation to reflect the logic is a better choice here, for the following reasons:

- In this case, changing the logic is a breaking change.
- The current logic is more general; it could work for strings, if `result` weren't hardcoded to start with `-Infinity`. I'm planning to fix that in version 2.0; see #2688.

So I suggest that you undo the code changes and instead update the `index.html` to replace the sentence "Non-numerical values in list will be ignored" by "This function can currently only compare numbers reliably". If you do so, please make sure to also update the documentation for `_.min`. You'll then still get the credits.

Does that sound like a plan?

anderlaw:
Thanks for the response, I'd like to do that

jgonggrijp (approved):

jgonggrijp:
Congratulations with your first contribution to Underscore, @anderlaw!

PR #2897, opened 2020-12-18
transition function implemented #2850

### LINKED ISSUE #2850: Add _.toggle to underscore
Hi,
I want to propose to add a new function called toggle to underscore.
At simple level it would toggle the value given to it. i.e if it is passed true it will return false and if it is passed false it will return true.
ex-
flag = _.toggle(flag)
// if flag was true then it will get set to false

Also additionally we can have another function that  if passed custom arguments then it will return a new function which can be used to toggle with those arguments
For ex - 
bitToggler = getTogglingfunction(0,1)
Now if we do 
bit = bitToggler (bit)
// If bit had value of 0 it will be set as 1

mohammadali-seifkashani:


jgonggrijp:
Hey @mohammadali-seifkashani, thanks for this very interesting contribution! There will be some more work to do and I might decide to put this in Underscore-contrib instead, but that's fine, we can make this ride together.

As a first step, could you add a few more commits to do the following? Thanks in advance.

- [ ] Rename the module to `modules/transition.js` (so without the leading `_` in the name).
- [ ] List your function with the others in `modules/index.js`. I think it belongs in the utility section.
- [ ] Add tests for your function in `test/utility.js`.

I'll do a more detailed code review after this.

PR #2899, opened 2020-12-21
Correct "Non-numerical values in list will be ignored"

anderlaw:


jgonggrijp:
@anderlaw Thanks for your submission. Why did you close this?

jgonggrijp:
Ah I see now, because of #2896.

Issue #1856, opened 2014-09-18
Feature request: _.deleteWhere


morgoth:
Currently having array of objects, to delete one of them I need to write:

``` javascript
var collection = [{id: 1}, {id: 3}];
var objectToRemove = _.findWhere(collection, function (obj) { return obj.id === 3 });
var objectIndex = collection.indexOf(objectToRemove);
collection.splice(objectIndex, 1);
```

It would be nice to have in underscore something like:

``` javascript
_.deleteWhere(collection, {id: 2})
```

If you find this useful I can try to implement it and send a PR.

PR #2906, opened 2021-02-03
added remove method (issue #1856)

matinFT:


test/arrays.js
```diff
@@ -583,4 +583,11 @@
     assert.deepEqual(_.chunk([10, 20, 30, 40, 50, 60, 70], 2), [[10, 20], [30, 40], [50, 60], [70]], 'chunk into parts of less then current array length elements');
     assert.deepEqual(_.chunk([10, 20, 30, 40, 50, 60, 70], 3), [[10, 20, 30], [40, 50, 60], [70]], 'chunk into parts of less then current array length elements');
   });
+
+  QUnit.test('remove', function(assert) {
+    assert.deepEqual(_.remove([1, 2, 3, 4], [1, 3, 4]), function(obj) {return(obj == 2)});
```
jgonggrijp, line 588: This line seems to have been written with the following intuition: if you have the elements 1, 2, 3, 4 and you remove the elements 1, 3, 4, you'll be left with only element 2. While this intuition is right, this is not what you should be testing, and the way you're trying to test it also doesn't work.

You shouldn't be testing this, because the intent of `_.remove` is to take as its second argument a *criterion* (`predicate`) by which to decide whether an element should be removed, not a *list of values* to be removed. A function that accepts a list instead of a criterion as its second argument would be the modifying version of `_.difference` instead of the modifying version of `_.reject`. As-is (with your current implementation) the argument `[1, 3, 4]` will in fact be interpreted as a predicate, i.e., it will be passed to `_.iteratee` and be converted to a function equivalent to `function(x) { return x[1][3][4]; }`. I'm sure this is not at all what you meant; I think you meant `function(x) { return x == 1 || x == 3 || x == 4; }`.

The test also doesn't work, because `assert.deepEqual()` takes two arguments and then recursively compares them by value. You're passing it two arguments: the result of `_.remove([1, 2, 3, 4], [1, 3, 4])` and a function. Unless you expect `_.remove` to return a function, this assertion is always going to fail. You should pass the expected value as the second argument to `assert.deepEqual`, not a function that performs a comparison.

There is also something missing. You should not only compare the result of `_.remove` to an expected value, but also verify that this result is the same array as the one that you passed as the first argument. After all, the intent of `_.remove` is that it modifies the array in place.

test/arrays.js
```diff
@@ -583,4 +583,11 @@
...
+    assert.deepEqual(_.remove([{a: 1}, {a:3, b: 2}, 3], [{a: 1}, 3], function(obj) {return(obj.a == 3)}));
+    assert.deepEqual(_.remove([{a: {a: 1}}, 2, 3], [2, 3], function(obj) {return(obj.a.a == 1)}));
+    assert.deepEqual(_.remove([{a: 1, b: 2, c: 3}, {a:3, b: 2}, {c: 3}], [{a:3, b: 2}, {c: 3}], function(obj) {return(obj.b == 2 && obj.c == 3)}));
```
jgonggrijp, line 591: The next three tests have the same problems, but in addition, you mixed up the parentheses. You're now passing three arguments to `_.remove` and only one to `assert.deepEqual`.

underscore.js
```diff
@@ -1,3 +1,5 @@
+const { default: remove } = require("./modules/remove");
```
jgonggrijp, line 1: This is not what you're supposed to be doing; please add your function to `modules/index.js` instead. You should only edit the sources in `modules/`, not the `underscore.js` or `underscore-esm.js` bundles.

From the fact that you were able to commit this, I can tell that you didn't run `npm install`, because otherwise the commit hook would have prevented it. Please run `npm install` before committing again.

jgonggrijp (changes requested):
Thanks for your submission. Please run `npm install` before your next commit and please see my comments below.

test/arrays.js
matinFT: yeah sorry my bad. don't know how i came up with this! i'll fix it and try with a new pr.

PR #2907, opened 2021-02-04
added remove method (issue #1856)

matinFT:


modules/remove.js
```diff
@@ -0,0 +1,8 @@
+// removes first element having the condition
```
jgonggrijp, line 1:
```suggestion
import cb from './_cb.js';

// removes first element having the condition
```

This is required for my next suggestion...

modules/remove.js
```diff
@@ -0,0 +1,8 @@
...
+export default function remove(collection, predicate) {
+  var index = collection.length;
```
jgonggrijp, line 3: ... which is to pass the predicate through our internal `cb` function. This enables the iteratee shorthands, so users can for example do `_.remove(anArray, { a: 1 })` (which would be equivalent to your testcase where you're passing `function(obj) { return obj.a == 1; }` as the second argument).

```suggestion
  var index = collection.length;
  predicate = cb(predicate);
```

modules/remove.js
```diff
@@ -0,0 +1,8 @@
...
+  while(index--){
+    if(predicate(collection[index])) collection.splice(index, 1)
```
jgonggrijp, line 5: Following the conventions of other Underscore functions, it would be nice to also pass the index and the whole collection to the predicate.

```suggestion
    if (predicate(collection[index], index, collection)) collection.splice(index, 1);
```

modules/remove.js
```diff
@@ -0,0 +1,8 @@
...
+  }
+  return collection;
+}
```
jgonggrijp, line 8: Please add a linebreak at the end of the file.

test/arrays.js
```diff
@@ -583,4 +583,19 @@
     assert.deepEqual(_.chunk([10, 20, 30, 40, 50, 60, 70], 2), [[10, 20], [30, 40], [50, 60], [70]], 'chunk into parts of less then current array length elements');
     assert.deepEqual(_.chunk([10, 20, 30, 40, 50, 60, 70], 3), [[10, 20, 30], [40, 50, 60], [70]], 'chunk into parts of less then current array length elements');
   });
+
+  QUnit.test('remove', function(assert) {
+    var result = [1, 2, 3, 4]
+    _.remove(result, function(obj) {return(obj == 2)})
+    assert.deepEqual(result, [1, 3, 4]);
```
jgonggrijp, line 590: This is good. I now understand that your tests in the previous PR where intended differently than I thought.

There is just one thing missing: these tests aren't verifying yet that `result` is also returned from `_.remove`. You can fix this as follows:

```suggestion
    var result = [1, 2, 3, 4];
    assert.strictEqual(_.remove(result, function(obj) {return(obj == 2)}), result);
    assert.deepEqual(result, [1, 3, 4]);
```

jgonggrijp (changes requested):
Very good @matinFT. Thanks for moving forward! We're nearly there.

I'm not seeing an update in `underscore.js` and `underscore-esm.js`. It's good that you didn't edit those files manually, but I expected the commit hook to update them automatically given your addition in `modules/index.js`. Did you not run `npm install`, or is the commit hook not working for you for some reason?

I see three remaining steps. You can continue adding and pushing new commits to your `remove_method` branch, they will be included in the current PR automatically (so you don't have to open a new PR).

- [ ] Get the commit hook to work so that `underscore.js` is automatically updated.
- [ ] Process my comments below.
- [ ] Add documentation for `_.remove` in the `index.html`.

Please let me know if you need any help. Good luck!

jgonggrijp:
Follow-up on #2906.

Issue #2883, opened 2020-10-11
Review _.throttle, _.debounce tests for time tampering


jgonggrijp:
`_.throttle` and `_.debounce` both have a test to confirm that the limiting time interval continues unaltered when `_.now` is overridden. These tests can be found by searching the `test/` directory for the phrase `after system time is set backwards`.

Since the modularization, these tests don't meet the intended purpose anymore; most Underscore functions now use direct references to other Underscore functions instead of indirect references through the `_` object. In a way, this has made them more tamper-proof. This also applies to the use of `_.now` in `_.throttle` and `_.debounce`.

How to resolve this issue depends on whether there is any other way to tamper with the time as it is reported by `_.now`. If there is such a way, then the tests in question should be updated to use it. Otherwise, the tests can be removed.

ognjenjevremovic:
Hey @jgonggrijp 👋 

Giving a glimpse to the implementation of `throttle` and `debounce` methods I can see they use the `now` method directly, rather than indirectly referencing them through the globally available `_` object.
Does this mean that the related tests ([this one](https://github.com/jashkenas/underscore/blob/master/test/functions.js#L387) and [this one](https://github.com/jashkenas/underscore/blob/master/test/functions.js#L541)) should be removed as obsolete, as the test cases don't provide overall value (since overriding the `_` methods have zero impact on the dependent functions)?

I would be willing to submit PR, but would perhaps need a little guidance on it.
If I understood well, this issue relates only to test cases? Are there any other test cases that prove to be obsolete after the modularization of the package?

Cheers!

jgonggrijp:
Hey @ognjenjevremovic 👋

Thanks for taking a peek at this, and for adding the links. I guess I should have done that immediately.

> Does this mean that the related tests ([this one](https://github.com/jashkenas/underscore/blob/master/test/functions.js#L387) and [this one](https://github.com/jashkenas/underscore/blob/master/test/functions.js#L541)) should be removed as obsolete, as the test cases don't provide overall value (since overriding the `_` methods have zero impact on the dependent functions)?

Well, that's basically the core question of this issue. You can't influence `throttle` and `debounce` by overriding `_.now`, but maybe you can _influence the output_ of `now` by overriding `window.Date`/`self.Date`/`global.Date` (or maybe just `Date.prototype.getTime`)?  Would be awesome if you could spend some time trying to hack `now` by overriding `Date`.

[...]

PR #2913, opened 2021-03-06
[Tests]: System time tampering effects on `_.throttle` and `_.debounce` methods

ognjenjevremovic:
✅ Closes: #2883

### Changes
Tests only:
* Monkey patch `Date.prototype.getTime` and `Date.prototype.valueOf` methods and `Date.now` method in `_.debounce` and `_.throttle` test cases,
* Provide additional test case asserting the outputs before and after monkey patching,
* Remove the reference to `_.now` method in test cases as the underlying implementation of library methods don't reference the methods from the `_` object but rather call dependent methods directly.

### Checklist
After the changes I ensured that:
 * no stylistic or unwanted errors are present, by running `npm run lint`,
 * no tests are failing, by running `npm run test`,
 * build passes with no issues, by running `npm run bundle`.

### Done
- [x] Added/updated unit tests for this change
- [x] Included links to related issues/PRs

### Additional notes
The underlying changes could've been done by monkey patching the native `Date` and `Date.prototype` methods in `before` (or `beforeEach`) and `after` (or `afterEach`) hooks in the *Functions* QUnit module. However, that would require a minor changes in two other test cases from the same module (as the `valueOf` method of the `Date.prototype` would be changed for all test cases and always return the same value which would cause certain test cases to hang to the operations `new Date - date < amount` as that will always result in the `false` value in the `while` loop).

I can provide the change if necessary using `before` and `after` hooks instead.
I just wanted to assure that the implementation of `_.throttle` and `_.debounce` methods will **not** break, if the system time would be tempered (monkey patched).
I guess the test cases can be safely omitted as they're redundant @jgonggrijp?
If so, I can provide another commit to the same PR removing the obsolete test cases.

### DIFF
--- a/test/functions.js
+++ b/test/functions.js
@@ -390,20 +390,64 @@
     var counter = 0;
     var incr = function(){ counter++; };
     var throttledIncr = _.throttle(incr, 100);
-    var origNowFunc = _.now;
+    var originalNowFunc = Date.now;
+    var originalGetTimeFunc = Date.prototype.getTime;
 
     throttledIncr();
-    assert.strictEqual(counter, 1);
-    _.now = function() {
-      return new Date(2013, 0, 1, 1, 1, 1);
-    };
+    assert.strictEqual(counter, 1, '_.throttle: incr was called immediately');
+
+    Date.prototype.getTime = function() {
+      return +(new Date(2013, 0, 1, 1, 1, 1));
+    }
+    Date.now = function() {
+      return +(new Date(2013, 0, 1, 1, 1, 1));
+    }
 
     _.delay(function() {
       throttledIncr();
-      assert.strictEqual(counter, 2);
+      assert.strictEqual(counter, 2, '_.throttle: incr was called successfully, with tampered system time');
       done();
-      _.now = origNowFunc;
+      Date.now = originalNowFunc;
+      Date.prototype.getTime = originalGetTimeFunc;
     }, 200);
+  });
+
+  QUnit.test('throttle continues to function after system time is not accessible (or in invalid format)', function(assert) {
+    assert.expect(3);
+    var done = assert.async();
+    var counter = 0;
+    var incr = function(){ counter++; };
+    var throttledIncr = _.throttle(incr, 100);
+    var originalNowFunc = Date.now;
+    var originalGetTimeFunc = Date.prototype.getTime;
+    var originalValueOfFunc = Date.prototype.valueOf;
+
+    throttledIncr();
+    assert.strictEqual(counter, 1, '_.throttle: incr was called immediately');
+
+    Date.prototype.valueOf = function() {
+      return null;
+    }
+    Date.prototype.getTime = function() {
+      return null;
+    }
+    Date.now = function() {
+      return null;
+    }
+
+    _.delay(function() {
+      throttledIncr();
+      assert.strictEqual(counter, 2, '_.throttle: incr was debounced successfully, with tampered system time');
+      Date.now = originalNowFunc;
+      Date.prototype.getTime = originalGetTimeFunc;
+      Date.prototype.valueOf = originalValueOfFunc;
+    }, 200);
+
+    _.delay(function() {
+      throttledIncr();
+      assert.strictEqual(counter, 3, '_.throttle: incr was debounced successfully, after system time method restoration');
+      done();
+    }, 400);
   });
 
   QUnit.test('throttle re-entrant', function(assert) {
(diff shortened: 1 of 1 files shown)


test/functions.js
```diff
@@ -390,22 +390,66 @@
...
-    var origNowFunc = _.now;
+    var originalNowFunc = Date.now;
+    var originalGetTimeFunc = Date.prototype.getTime;
 
     throttledIncr();
-    assert.strictEqual(counter, 1);
-    _.now = function() {
-      return new Date(2013, 0, 1, 1, 1, 1);
-    };
+    assert.strictEqual(counter, 1, '_.throttle: incr was called immediately');
```
jgonggrijp, line 397: Yes, good to add a comment to the assertion. Much clearer if it fails. 👍

test/functions.js
```diff
@@ -390,22 +390,66 @@
...
-      assert.strictEqual(counter, 2);
+      assert.strictEqual(counter, 2, '_.throttle: incr was called successfully, with tampered system time');
       done();
-      _.now = origNowFunc;
+      Date.now = originalNowFunc;
+      Date.prototype.getTime = originalGetTimeFunc;
     }, 200);
   });
 
+  QUnit.test('throttle continues to function after system time is not accessible (or in invalid format)', function(assert) {
```
jgonggrijp, line 415: Wow!

test/functions.js
```diff
@@ -390,22 +390,66 @@
...
+    Date.prototype.getTime = function() {
+      return null;
+    }
+    Date.now = function() {
+      return null;
+    }
+
+    _.delay(function() {
+      throttledIncr();
+      assert.strictEqual(counter, 2, '_.throttle: incr was debounced successfully, with tampered system time');
```
jgonggrijp, line 440: Throttling and debouncing are subtly different things, so to prevent confusion, I'd rather avoid the word "debounce" here. There is no convention to prefix all assertion comments with a qualified function name (at least not as far as I'm aware) and I don't think such a convention is necessary, either, so I suggest just changing it to this:

```suggestion
      assert.strictEqual(counter, 2, 'incr was throttled successfully, with tampered system time');
```

test/functions.js
```diff
@@ -390,22 +390,66 @@
...
+      Date.now = originalNowFunc;
+      Date.prototype.getTime = originalGetTimeFunc;
+      Date.prototype.valueOf = originalValueOfFunc;
+    }, 200);
+
+    _.delay(function() {
+      throttledIncr();
+      assert.strictEqual(counter, 3, '_.throttle: incr was debounced successfully, after system time method restoration');
```
jgonggrijp, line 448: likewise

```suggestion
      assert.strictEqual(counter, 3, 'incr was throttled successfully, after system time method restoration');
```

test/functions.js
```diff
@@ -542,24 +586,68 @@
...
+      return +(new Date(2013, 0, 1, 1, 1, 1));
+    }
+    Date.now = function() {
+      return +(new Date(2013, 0, 1, 1, 1, 1));
+    }
 
     _.delay(function() {
       debouncedIncr();
-      assert.strictEqual(counter, 2, 'incr was debounced successfully');
+      assert.strictEqual(counter, 2, '_.debounce: incr was debounced successfully, with tampered system time');
```
jgonggrijp, line 607: for brevity

```suggestion
      assert.strictEqual(counter, 2, 'incr was debounced successfully, with tampered system time');
```

test/functions.js
```diff
@@ -542,24 +586,68 @@
...
+    Date.prototype.getTime = function() {
+      return null;
+    };
+    Date.now = function() {
+      return null;
+    };
+
+    _.delay(function() {
+      debouncedIncr();
+      assert.strictEqual(counter, 2, '_.debounce: incr was debounced successfully, with tampered system time');
```
jgonggrijp, line 640: again for brevity

```suggestion
      assert.strictEqual(counter, 2, 'incr was debounced successfully, with tampered system time');
```

test/functions.js
```diff
@@ -542,24 +586,68 @@
...
+      Date.now = originalNowFunc;
+      Date.prototype.getTime = originalGetTimeFunc;
+      Date.prototype.valueOf = originalValueOfFunc;
     }, 200);
+
+    _.delay(function() {
+      debouncedIncr();
+      assert.strictEqual(counter, 3, '_.debounce: incr was debounced successfully, after system time method restoration');
```
jgonggrijp, line 648: and one last time.

```suggestion
      assert.strictEqual(counter, 3, 'incr was debounced successfully, after system time method restoration');
```

jgonggrijp (changes requested):
This is a great patch! Not only did you update those time-tampering tests to make them meaningful again, you also added new tests proving that `_.throttle` and `_.debounce` provide a *stronger guarantee*, i.e., they even continue working in the face of... let's call it "time vandalism". And you were very thorough in the PR description, I like that. Thank you so much, @ognjenjevremovic!

I have a few very minor change requests about the assertion comments, otherwise everything looks excellent.

Out of curiosity: why is there a ring emoji in the commit title? Is it some kind of cultural reference?

To answer your notes:

> The underlying changes could've been done (...) in `before` (or `beforeEach`) and `after` (or `afterEach`) hooks in the _Functions_ QUnit module. However, that would require a minor changes in two other test cases from the same module (...).
> 
> I can provide the change if necessary using `before` and `after` hooks instead.

I'm fine with the current solution, but if you want to refactor things a little bit, I'm fine with that, too. You can nest QUnit modules so that the hooks only affect the tests that should be affected, see docs here: https://api.qunitjs.com/QUnit/module/.

> I just wanted to assure that the implementation of `_.throttle` and `_.debounce` methods will **not** break, if the system time would be tempered (monkey patched).
> I guess the test cases can be safely omitted as they're redundant @jgonggrijp?
> If so, I can provide another commit to the same PR removing the obsolete test cases.

Which test cases are obsolete?

ognjenjevremovic:
Hey @jgonggrijp, thank you so much for such an in depth review.
I really appreciate it!

> I have a few very minor change requests about the assertion comments, otherwise everything looks excellent.

You're right; the messages should definitely be more concise.
To be honest, I wasn't sure how to structure the assertion comments. I wanted to somehow diversify the messages between `_.debounce` and `_.throttle` but I'm not that good with the wording; I thought prefixing the assertion comments with the method name could help.
I'll be sure to address these with the suggested changes.


> Out of curiosity: why is there a ring emoji in the commit title? Is it some kind of cultural reference?

I apologize for this 😄 .
I usually like to structure the commit messages in a descriptive manner, using the format: *intention:* + *summary* of the change (as a commit message head) followed by a *long description*. I use [`git-cz`](https://www.npmjs.com/package/git-cz) to make such formated commit messages.
I think the latest `major` version of the tool included the emojis and automatically prepends them to every commit message (made using the tool).
I can change the commit message and remove the emoji if you'd like?
\* *(Just to assure you, the emoji does not represents a symbol of any sort and it was unintentional as it does not really provide any meaningful value to the actual commit message).*

Thank you for noticing such a detail 🙂 .


> Which test cases are obsolete?

Just wanted to check if adding these 2 new test cases provided value to the overall tests and didn't include redundancy.
Also, I wanted to check if it is ok to keep the altered tests now that they use `Date.prototype` methods instead of `_.now`.
I think you already answered my question in your review; we want to keep all the tests.


> I'm fine with the current solution, but if you want to refactor things a little bit, I'm fine with that, too.

If code duplication in tests is not too big of a trouble and doesn't introduce the confusion in the test cases by any means, I'm fine with the current implementation as well.
Using `beforeEach` and `beforeAll` would require certain changes and might be different from all other scenarios in the tests.

jgonggrijp (approved):
> To be honest, I wasn't sure how to structure the assertion comments. I wanted to somehow diversify the messages between `_.debounce` and `_.throttle` but I'm not that good with the wording; I thought prefixing the assertion comments with the method name could help.
> I'll be sure to address these with the suggested changes.

For clarity, I'm not entirely opposed to doing this. I just suggested removing the prefix in a few places because it was redundant in those particular cases.

> I usually like to structure the commit messages in a descriptive manner, using the format: _intention:_ + _summary_ of the change (as a commit message head) followed by a _long description_. I use [`git-cz`](https://www.npmjs.com/package/git-cz) to make such formated commit messages.
> I think the latest `major` version of the tool included the emojis and automatically prepends them to every commit message (made using the tool).
> I can change the commit message and remove the emoji if you'd like?

Ah, that explains! I think I have seen similar commit messages elsewhere, now I understand where they're coming from. No need to apologize and you can leave the emoji in there. I think it's amusing.

> If code duplication in tests is not too big of a trouble and doesn't introduce the confusion in the test cases by any means, I'm fine with the current implementation as well.
> Using `beforeEach` and `beforeAll` would require certain changes and might be different from all other scenarios in the tests.

Yeah, if going that route, it's probably more consistent to restructure the tests a bit in general. That would more likely belong in a new issue or PR.

jgonggrijp:
Congratulations with your first contribution! Please feel welcome to make more contributions.

(A suggestion, if you feel up to it: I have a big PR waiting (#2908) which requires multiple reviews by different people. If you could take a look at the diff and leave some comments, or if you know other, seasoned Underscore users who might be willing to do a code review (or contribute a real-world benchmark), that would be great.)

PR #2928, opened 2021-05-16
Fix bug in function emulatedSet

anfengchen:
The old code for contains: when we check 'toString' inherited from Object.prototype, the contains return Object.prototype.toString;In this case, it prevent us to add obj's own 'toString' to its keys.


jgonggrijp:
Looks good to me. Thanks for the fix!

PR #2930, opened 2021-05-30
Enabled faster algorithm for _.uniq with non-function iteratee

afanasy:
The faster algorithm on sorted array will not work for `_.uniq` with an iteratee, if the iteratee is not a one-to-one function. Because of that it was disabled for all iteratees, even for those producing one-to-one functions, such as `String` and `Array` iteratees (handled by `property()` function). This PR introduces the code that checks the iteratee type and enables faster algorithm if possible.

### DIFF
--- a/modules/uniq.js
+++ b/modules/uniq.js
@@ -1,4 +1,6 @@
 import isBoolean from './isBoolean.js';
+import isString from './isString.js';
+import isArray from './isArray.js';
 import cb from './_cb.js';
 import getLength from './_getLength.js';
 import contains from './contains.js';
@@ -6,7 +8,7 @@ import contains from './contains.js';
 // Produce a duplicate-free version of the array. If the array has already
 // been sorted, you have the option of using a faster algorithm.
 // The faster algorithm will not work with an iteratee if the iteratee
-// is not a one-to-one function, so providing an iteratee will disable
+// is not a one-to-one function, so providing an iteratee function will disable
 // the faster algorithm.
 export default function uniq(array, isSorted, iteratee, context) {
   if (!isBoolean(isSorted)) {
@@ -14,13 +16,14 @@ export default function uniq(array, isSorted, iteratee, context) {
     iteratee = isSorted;
     isSorted = false;
   }
+  var faster = isSorted && (!iteratee || isString(iteratee) || isArray(iteratee));
   if (iteratee != null) iteratee = cb(iteratee, context);
   var result = [];
   var seen = [];
   for (var i = 0, length = getLength(array); i < length; i++) {
     var value = array[i],
         computed = iteratee ? iteratee(value, i, array) : value;
-    if (isSorted && !iteratee) {
+    if (faster) {
       if (!i || seen !== computed) result.push(value);
       seen = computed;
     } else if (iteratee) {


jgonggrijp (commented):
Thank you for going out of your way to contribute, @afanasy!

Unfortunately, I have to reject your proposal. I believe it would make the semantics of `_.uniq` less consistent, as I've explained in a comment below. It would also be a breaking change.

At this time, I can think of three other ways that `_.uniq` might change, which would also improve performance and which I could agree with. Two of them would however also be breaking changes, so they would have to wait until version 2.0.

1. Allow the fast algorithm for `_.identity` as iteratee function. This would be non-breaking, since the result for `_.identity` is the same regardless of whether you take the fast or the slow algorithm.
2. Use ES6 `Set` for the slow algorithm. This will dramatically reduce the performance difference between the fast and the slow algorithms. It is however also a breaking change, since it requires (a polyfill for) `Set` and it treats `NaN` differently. I'm seriously considering this change for Underscore 2.0 (see also #2147, #2453, #1862).
3. Change the precondition for the fast algorithm from requiring that the iteratee is injective to requiring that the array `isSorted` *with respect to the iteratee*. This basically amounts to taking the user's word that they will only pass `isSorted === true` when the array has the same order as if `_.sortBy(array, iteratee)` had been applied just before and that all values within each equivalence class compare equal. Since this changes the precondition as well as the behavior, this is a less subtly breaking change.

I will close this pull request now, but please feel welcome to discuss.

afanasy:
@jgonggrijp Thank you for the detailed and prompt response. I agree that 
```js
var faster = isSorted && !isFunction(iteratee);
```
looks much better. However, I'm failing to understand the main point here - why is it a breaking change and how the `_.property` iteratees are able to produce not "one-to-one" values, unsuitable for the fast algorithm? Could you provide an example? And just as a side note, introducing `Set` will "dramatically reduce the performance difference between the fast and the slow algorithms" only for smaller arrays. For larger arrays the fast algorithm `O(n)` will still be much faster than the slow one `O(n^2)`.

jgonggrijp:
Thank you for sharing your doubts.

> (...) I'm failing to understand the main point here - why is it a breaking change and how the `_.property` iteratees are able to produce not "one-to-one" values, unsuitable for the fast algorithm? Could you provide an example?

I think this part of the unittests illustrates the one-to-one issue quite well:

https://github.com/jashkenas/underscore/blob/3d2f17a62b2987c2b7f93880edb0d4e48ce61f10/test/arrays.js#L169-L171

The discrepancy arises because the criterion the array was originally sorted by (magnitude) is not the same criterion as the array is being uniqued by (magnitude of the square). You can get into a similar situation with `_.property` iteratees as follows. Let's say the array was sorted by the `rank` property and is now being uniqued by the `flavor` property:

```js
var theArray = [{rank: 1, flavor: 'vanilla'}, {rank: 2, flavor: 'strawberry'}, {rank: 3, flavor: 'vanilla'}];
var result = _.uniq(theArray, true, 'flavor');
```

My "acceptable change" number 3 above rests on the semantic change of expecting users to only pass `isSorted === true` when the sort criterion and the unique criterion are the same. Your confusion may stem from the assumption that this is already the case, but it isn't.

> And just as a side note, introducing `Set` will "dramatically reduce the performance difference between the fast and the slow algorithms" only for smaller arrays. For larger arrays the fast algorithm `O(n)` will still be much faster than the slow one `O(n^2)`.

The standard dictates that `Set` must provide a faster-than-linear lookup mechanism. Generally, that's either O(k) with small k for hash tables or O(log n) for binary search trees. The slow algorithm then becomes O(nk) or O(n log n) overall, which is not quite  as fast as the fast algorithm yet, but *way* faster than the current slow algorithm, *especially* for large arrays.

afanasy:
@jgonggrijp This all makes sense, yes, I assumed users know what they are doing, when passing the `isSorted === true`, but that may not always be the case. And you are right about `Set`, it should provide the boost from using internal uniqueness check. Thank you for taking the time to clarify my misunderstanding.

Issue #2780, opened 2018-10-08
Documentation Site has very small font.


GajananMore:
Hi,

Documentation is in small font, have to zoom in every time needs to see document. can you increase font?

Thanks,
Gajanan.

HZooly:
@GajananMore Font size seems correct. I'm using Chrome 70 on macOS

jgonggrijp:
@GajananMore Sorry to hear that you found the documentation difficult to read. Do you still have this problem?

If you still have this problem, please tell us some details about your operating system, hardware and browser version. Perhaps also show us a screenshot. Thanks in advance.

bnjmnt4n:
This is probably related to the website being not very mobile-friendly.

PR #2935, opened 2021-07-13
Mobile responsive and more readability

itsdeekay:
Tried to resolved issue #2780 to make documentation site mobile responsive and change font little bit for better readability.

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,15 +4,14 @@
   <meta http-equiv="content-type" content="text/html;charset=UTF-8" />
   <meta http-equiv="X-UA-Compatible" content="chrome=1" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
-  <meta name="viewport" content="target-densitydpi=device-dpi" />
-  <meta name="HandheldFriendly" content="true"/>
+  
   <link rel="canonical" href="https://underscorejs.org/" />
   <link rel="shortcut icon" href="favicon.ico" type="image/x-icon" />
   <title>Underscore.js</title>
   <style>
     body {
       font-size: 14px;
-      line-height: 22px;
+      line-height: 20px;
       background: #f4f4f4 url(docs/images/background.png);
       color: #000;
       font-family: Helvetica Neue, Helvetica, Arial;
@@ -24,7 +23,7 @@
       background: #fff;
       position: fixed;
       top: 0; left: 0; bottom: 0;
-      width: 200px;
+      width: 18vw;
       overflow-y: auto;
       overflow-x: hidden;
       -webkit-overflow-scrolling: touch;
@@ -46,12 +45,12 @@
         font-weight: normal;
       }
     ul.toc_section {
-      font-size: 11px;
-      line-height: 14px;
+      font-size: 12px;
+      line-height: 15px;
       margin: 5px 0 0 0;
       padding-left: 0px;
       list-style-type: none;
-      font-family: Lucida Grande;
+      font-family: Helvetica Neue, Helvetica, Arial;;
     }
       .toc_section li {
         cursor: pointer;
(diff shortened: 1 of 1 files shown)


jgonggrijp (changes requested):
Thanks for your hard work, @itsdeekay! It's definitely a large improvement on mobile already. I'm also impressed by the clever, well-implemented CSS-only menu toggle.

For anyone reading along, you can compare the website before and after the changes I'm reviewing here, "live" in any browser on desktop or mobile by visiting these two links: [before](https://cdn.statically.io/gh/jashkenas/underscore/1.13.1/index.html), [after](https://cdn.statically.io/gh/itsdeekay/underscore/aa37bed/index.html).

On desktop, I'd prefer to keep the old font and the absolute (rather than relative) sidebar width. I'm however fine with making the font slightly larger, like you have done, and optionally also with making the sidebar slightly wider.

On mobile, I'm very happy with the changes you have made, but I'd like you to take them a bit further still if possible. Trying out your changes in Safari on my mobile phone, I identified a few additional changes that I think could make things even better:

- [x] Since the menu can be closed again, there is no need to keep the main content in sight while it is open. So as far as I'm concerned, you can stretch the menu to fill the entire viewport. This will enable mobile users to see more of the menu at once and will also make it easier to scroll through the menu.
- [x] Likewise, as far as I'm concerned you can give the text inside the menu the same size as the main content.
- [x] After navigating to a content section down on the page, the menu is hard to reach because you have to scroll back all the way to the top. Giving the menu toggle and the sidebar a `position: fixed` and layering them on top of the main content with a higher `z-index` might solve this.
- [x] The search box has `autofocus`. On desktop, that's great because you can start typing immediately. On mobile with your changes, it's not so great, because the autofocus only takes effect when the menu becomes visible (at least on mobile Safari) and then it causes the zoom level and scrolling position to mess up (because mobile Safari is trying to make the `<input>` element well visible and well positioned above the on-screen keyboard). Mobile users will not always want to type something when opening the menu, anyway, so one possible solution might be to just remove the `autofocus` attribute on mobile browsers after page load using a little bit of JavaScript.
- [ ] When following a link from the menu to a section within the main content, the menu stays open. It would be better to close it immediately, especially if you adopt my earlier suggestion to make the menu float on top of the main content. I think all it takes is a little bit of JavaScript that re-checks the `#menu` checkbox on the `hashchange` event.

A final question: on which browsers have you tested your changes? Have you tested it with a screen reader as well, by any chance?

In any case, I'm quite sure I will accept your changes after some more polishing. My compliments for your work so far!

itsdeekay (commented):
Hi @jgonggrijp Sorry for this late reply, caught up in something. I have worked on the issues you mentioned and improved it. Pull request is modified with new code. Please review it and provide your feedback.

jgonggrijp:
Hey @itsdeekay thanks for getting back here!

I'm going to review this soon, but right now, I can't easily preview the code on a mobile device because Statically.io has changed its policies. You now need to opt in to the HTML service. Would you mind doing that? If you fork this gist: https://gist.github.com/jashkenas/e425f9c8e676ab175534e61c43318569, I can take it from there (the second and final step is a tweet to the maintainer of Statically.io). The whole "procedure" is explained over here: https://statically.io/docs/using-staticzap/.

jgonggrijp:
Just realized I could do a workaround by copying the branch to my own fork. This comment will be updated.

[before](https://cdn.statically.io/gh/jashkenas/underscore/1.13.1/index.html), [after](https://cdn.statically.io/gh/jgonggrijp/underscore/eb921fb/index.html)

pointert:
您的邮件我已收到，谢谢！

jgonggrijp (approved):

PR #2941, opened 2021-09-23
Update groupBy.js

Felichz:


jgonggrijp:
Hey @Felichz, thank you for going out of your way to contribute to Underscore.

Unfortunately I have to reject this proposal, for a couple of reasons:

- The change does not solve a problem or add new functionality (modern syntax is nice, but it is never a goal in itself).
- The change is not consistent (you updated only one module, while similar changes could apply to most other modules).
- The change introduces a dependency on new syntax that will break the library for many existing users.

Successful pull requests generally include the following elements:

- a very descriptive title;
- a short explanation of the goal of the PR;
- a short explanation of the overall approach;
- changes to both logic and tests if fixing a problem or adding functionality;
- changes to documentation if adding functionality or improving documentation;
- a small increase in code size for a large increase in functionality, if adding functionality;
- no changes that would break the library for existing users.

If you have any questions or comments, please feel free to reply.

PR #2951, opened 2022-04-03
Test: Use same QUnit version in CI as on website

Krinkle:
Follows-up c9b4b63fd08847281260205b995ae644f6f2f4d2 and 13f71bc9711b3617dd0fd5f1f085655df8505cf8.

Contrary to the commit message, it actually restored the older QUnit 2.6, and not the newer QUnit 2.10 version from package.json.

The oldest Node.js LTS is currently Node 12.
The latest QUnit 2.18 supports Node 10+.
This repo currently has a Node 8 CI job.

While at it, update both to 2.10.1, the last version to support Node 8 ([changelog](https://github.com/qunitjs/qunit/blob/2.18.1/History.md#2110--2020-08-16)), and pin it to avoid "npm install" from expanding it and saving it to the lock file which Node 8 would see. Hopefully also avoiding e.g. using a newer method which would then fail on the website version later.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "underscore",
-  "version": "1.13.1",
+  "version": "1.13.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


jgonggrijp (approved):
I was initially a bit confused by your description, but after reading back the commits you referred to, I understood what happened. It looks correct to me. Thanks for the fix!

Issue #2949, opened 2022-03-06
_.max gets incorrect results in extendScript


RaymondClr:
```
//In Underscore.js 1.13.2
var stooges = [{name: 'moe', age: 40}, {name: 'larry', age: 50}, {name: 'curly', age: 60}];
var result = _.max(stooges, function(stooge){ return stooge.age; });
result.age
=> Result: 40
```
After my troubleshooting, I found that the problem first appeared in version 1.7.0

```
//In Underscore.js 1.6.0
var stooges = [{name: 'moe', age: 40}, {name: 'larry', age: 50}, {name: 'curly', age: 60}];
var result = _.max(stooges, function(stooge){ return stooge.age; });
result.age
=> Result: 60
```

jgonggrijp:
@RaymondClr Thanks for reporting.

I don't have the means to reproduce the problem. Your code example returns `60`, as it should, with Underscore 1.13.2 in Safari. `_.max` is also passing the unittests in Node.js and several browsers, which include examples like yours.

I looked at [the diff between 1.6.0 and 1.7.0](https://github.com/jashkenas/underscore/compare/1.6.0...1.7.0#diff-246afe7b9e4859c8a04b7747287c195ec066153c711eba01e552b2cbdd2da227L251-R283), but I see no obvious change that might have broken `_.max` for ExtendScript. However, that is likely due to my limited knowledge of ExtendScript.

(The diff link opens on the wrong part of the page. You have to click on the "Files changed" tab, scroll all the way to the bottom and click "Load diff" on `underscore.js`. The scrolling position will then jump to `_.max`, which I've highlighted.)

Concluding, I believe you, but I have no means to address the problem myself. Does ExtendScript have a debugger and if so, could you step through your example code to find out where things are going wrong?

RaymondClr:
@jgonggrijp 
Thank you for your reply. First of all, I am sorry that my English is not very good. The following content is the result of Google translation. I hope it can express my meaning correctly.

extendScript has a very old debugging tool (its official download page is not even accessible anymore), called Adobe ExtendScript Toolkit CC, its debugging function is very poor, so I can only troubleshoot problems by printing intermediate variables in the program, but In the end, I found the cause of the problem. The logic of extendScript (based on es3) in parsing the ternary operator seems to be different from the conventional implementation. I inserted the following print content in the _.max source code of Underscore version 1.7.0:

```javascript
  _.max = function(obj, iteratee, context) {
[...]

PR #2953, opened 2022-04-04
Parenthesize mixed expressions of || and &&

jgonggrijp:
ExtendScript apparently gives equal precedence to the || and && operators, so when the || appears before the &&, we need to anchor the order of evaluation with an extra pair of parentheses. See #2949.

@RaymondClr Could you test whether all of your code works correctly in ExtendScript with https://github.com/jgonggrijp/underscore/raw/extendscript-precedence/underscore-umd.js?


jgonggrijp:
Excellent comments, thanks @captbaritone! In one short post, you mentioned three things I didn't think of:

- [ ] Try whether parentheses can be added in a build step.
- [ ] Failing that, try to enforce them through an eslint rule.
- [ ] Parentheses must not be stripped during minification (though if we can't prevent this, I guess ExtendScript users can fall back to the unminified code).

I plan to go through the above list within the current PR.

jgonggrijp:
@captbaritone Outcome of our discussion:

- The obvious way to add the parentheses in a build step would be by applying the `fix` option of the `eslint` plugin for the Rollup bundler. That basically means that `eslint` comes into play in any case and that the question remains whether the rule can be automatically fixed.
- The rule in question is `no-mixed-operators`, which unfortunately `eslint` cannot fix automatically. Nevertheless, I added the rule and configured it such that only mixes of `&&` and `||` are disallowed. CI will fail if people submit PRs with mixed expressions.
- It turns out that `terser` was already adding parentheses to mixed expressions, so nothing needs to change with respect to minification.

jgonggrijp:
@RaymondClr I'm just going to merge this now on the assumption that this solves #2949. However, please still double-check that the issue is solved on your end when you find the time. Thanks in advance!

captbaritone:
Awesome. Thanks for the write up! Sounds like a great solution.

captbaritone:
It might be worth adding a comment to the ESLint config explaining why the rule is enabled. Otherwise we might opt to remove in the future assuming it's purely stylistic without realizing it's load bearing.

jgonggrijp:
Oops, good point. I assumed this wouldn't be allowed because the config is in JSON, but according to the eslint docs, it is allowed. I'll add it it in a separate commit on `master`.

dogbubu:
gsdg54sdg5s4d5gドっぱくラバニラサダウのブはとてもおいしそうなりんかマカョ♂

PR #2955, opened 2022-04-23
Prepare 1.13.3

jgonggrijp:
As usual, I'm submitting this PR just to leave a trace and to trigger CI.


jgonggrijp:
The Travis build reports "failed", but the situation is not as bleak as it seems. The first failure was a fluke, one of those `throttle` tests that fails every 1000 runs or so. It passed in the second run, but that one failed because Coveralls cannot stand it when you attempt to run it twice on the same commit. Which reminds me once again not to do PRs from the main repository.

Long story short: everything is fine.

Issue #2958, opened 2022-05-23
webpack - module federation "No version" warning as a shared module with version


petschki:
we use this package with webpack module federation which complains, that there's no version secified in `modules/package.json`

```
WARNING in shared module underscore -> /workspace/buildout.coredev-6.0/src/mockup/node_modules/underscore/modules/index-all.js
No version specified and unable to automatically determine one. No version in description file (usually package.json). Add version to description file, or manually specify version in shared config. ```

PR #2959, opened 2022-05-23
Fix for webpack module federation "No version" error

petschki:
fixes https://github.com/jashkenas/underscore/issues/2958


jgonggrijp (approved):
I raised an eyebrow initially, but I see the sense of it now. Thanks for contributing!

Issue #2948, opened 2022-02-28
Possibility to update a nested object's property in a non-mutable way


StudioSpindle:
Currently I am using something like:

```typescript
const foo = bar.map((baz) => {
  return {
    ...baz,
    qux: baz.qux.map((quux) => {
       return {
         ...quux,
         aNewProperty: 'yay!'
       }
    }
  }
});
```
To make the point clear, it would be ideal to have something similar to the [set method in lodash](https://lodash.com/docs/4.17.15#set):

```typescript
const foo = bar;
set(foo, 'bar.baz.qux.quux', { bar.baz.qux.quux, aNewPorperty: 'yay!' });
```

Note: I am not aware of how lodash and underscore are related but the current project I'm working on is using underscore. And at this time it's not possible to convert to lodash just for this single use-case.

dogbubu:
gsdg54sdg5s4d5gドっぱくラバニラサダウのブはとてもおいしそうなりんかマカョ♂

jgonggrijp:
Yes, I agree it would make sense to have a `set` function in Underscore as well. In the meanwhile, you can define your own `set` function like this so it will be maintainable and reusable and integrate well with Underscore:

```js
import _, { extend, isObject } from 'underscore';

var arrayIndex = /^\d+/;

function keyValue(key, value) {
    var result = {};
    result[key] = value;
    return result;
}

function innerSet(obj, path, value) {
    if (!path.length) return value;
    var key = path[0];
    // Important: the next line prevents prototype pollution.
    if (key === '__proto__') throw new Error('Prototype assignment attempted');
    obj = obj || (arrayIndex.test(key) ? [] : {}); 
    value = innerSet(obj[key], path.slice(1), value);
    return extend(obj, keyValue(key, value));
}

function set(collection, path, value) {
    if (!isObject(collection)) return collection;
    path = _.toPath(path);
    return innerSet(collection, path, value);
}
```

Use it like this:

```js
set(foo, ['bar', 'baz', 'qux', 'quux', 'aNewProperty'], 'yay!');
```

You can add it to Underscore so you can also use it in chaining:

```js
import { mixin } from 'underscore';
import { set } from './your/own/module.js';

_.mixin({set});

_.chain({a: 1})
.extend({b: 2})
.set(['c', 0, 'd'], 3)
.value();
// {a: 1, b: 2, c: [{d: 3}]}
```

[...]

PR #2961, opened 2022-07-19
Feature/set

AzatCoder:
Hi there!
Let's add function `set` to underscore.
 #2948


jgonggrijp (changes requested):
Thank you for contributing!

I want to merge your branch, but it needs a bit more work. I suggest you work in the following order:

- [ ] I requested a change in the tests (last comment). Please make that change before changing the logic and verify that it fails initially.
- [ ] I pointed out three logical mistakes in my comments below: `isArray`/`isObject` in _set.js:14, `isNumber` in _set.js:16 and `_.toPath` in set.js:8. Before addressing them, please add tests that expose them (i.e., that fail initially).
- [ ] Your code is currently vulnerable to [prototype pollution](https://portswigger.net/daily-swig/prototype-pollution-the-dangerous-and-underrated-vulnerability-impacting-javascript-applications). Add a test that demonstrates this.
- [ ] Make the edited tests pass and address the remainder of my comments, possibly taking inspiration from the code I suggested in #2948.
- [ ] Add an entry to the `index.html` to document the new `set` function. (I know the `CONTRIBUTING.md` states you shouldn't update the documentation. That is meant to refer only to the annotated sources and the changelog, both of which are updated on release. You are welcome to clarify that part of the `CONTRIBUTING.md` if you feel like it.)

AzatCoder:
Thank you for quick reviewing, @jgonggrijp

modules/_set.js
```diff
@@ -0,0 +1,20 @@
...
+
+export default function set (obj, path, value) {
+  var key = String(path[0]);
+
+  if (path.length === 1) {
+    obj[key] = value;
+    return;
+  }
+
+  if (!isArray(obj[key]) || !isObject(obj[key])) {
```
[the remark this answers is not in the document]
AzatCoder: yep, my bad. It should be `if (!isObject(obj[key])) {...}`

modules/set.js
```diff
@@ -0,0 +1,14 @@
+import isArray from './isArray.js';
+import isObject from './isObject.js';
+import _set from './_set.js';
+
+
+// set the value in given path
+export default function set (obj, path, value) {
+  if (!isObject(obj) || !isArray(path)) return obj;
+  if (path.length === 0) return obj;
```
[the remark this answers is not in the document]
AzatCoder: Sorry, I don't see other option better than this. I like this implementation. Please let me know if you have an idea how to improve it.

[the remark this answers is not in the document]
jgonggrijp: Please refer to the approach I suggested in #2948. If you do postorder assignment instead of preorder assignment, you can check for `path.length === 0` only in the internal function and immediately return `value` if true. You can avoid lookahead beyond `path[0]` in that case by replacing `obj` instead of `obj[key]`.

Referring again to #2948, I also realize that your code is still vulnerable to prototype pollution. You need to fix that as well.

modules/set.js
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// Set the value on `path` of `object`.
+// If any property in `path` does not exist it will be created.
+// Returns mutated object (`obj`).
+export default function set(obj, path, value) {
+  path = toPath(path);
+
+  if (!isObject(obj) || !path.length) return obj;
+  if (contains(path, '__proto__')) throw new Error('Prototype assignment attempted');
```
jgonggrijp, line 32: Two things. Firstly, I'm in favor of reusing our own function in principle. However, in this case the use of `_.contains` means that `_.get` will come out a bit "fatter" when people try to treeshake it. For this reason, I suggest checking the current `key` inside `deepGet` instead.

Secondly, I just realized we need to check not only for `__proto__` but also for `constructor` and `prototype`. Otherwise, attackers can (for example) still pollute `String` methods.

modules/set.js
```diff
@@ -0,0 +1,37 @@
...
+import toPath from './_toPath.js';
+import contains from './contains.js';
+
+
+var arrayIndex = /^\d+$/;
+
+// Internal function of `set`.
+function deepSet(obj, path, value) {
+  var key = path[0];
+
```
jgonggrijp, line 11: I think this is the right place to check that `key` is not one of `'__proto__'`, `'constructor'` or `'prototype'`.

jgonggrijp (changes requested):
I'm quite happy! I just think we should avoid using `_.contains` for this particular function, to keep the treeshakers happy, and add two more safeguards against prototype pollution.

Nearly there!

modules/set.js
AzatCoder: I'm afraid it will mutate before throwing an error.
```javascript
const obj = { value: 50 };

try {
  _.set(obj, [ 'key', '__proto__' ], {});
} catch(e) {
  console.log(e.message);  //  Prototype assignment attempted
  console.log(obj);        //  { value: 50, key: {} }  -  `obj` was mutated :(
}
```

jgonggrijp: Yeah, I realized that as well. I can think of three options (from least to most preferred, in my opinion):

1. Stick with `_.contains`, accept the double iteration and the greater weight after treeshaking.
2. Accept that `obj` gets somewhat corrupted. At least the prototype pollution attempt is detected and unsuccessful.
3. Switch to postorder assignment as I suggested before, which has neither of these problems.

I didn't mention it because I was ready to accept option 2, but I would prefer option 3.

jgonggrijp:
@AzatCoder you are still welcome to finish this, as far as I'm concerned.

PR #2962, opened 2022-08-29
Added more entities to _escapeMap.js

ChdML:
Adding more entities to `_escapeMap.js` doesn't hurt


jgonggrijp:
Yes it does.

Next time you make a contribution, make sure that the change actually solves a problem, respect existing tests and add new tests.

Issue #2853, opened 2020-06-02
Double chaining leads to a stack overflow


jgonggrijp:
The following expression is useless, since you'd expect the second `chain` call to just return a new wrapper of the same value:

```js
_.chain([1]).chain()
```

but rather than behaving like a no-op, it results in a stack overflow. Credits to @reubenrybnik for pointing this out in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45201#issuecomment-637696652. This minor pitfall might have been introduced with #1691.

I'm hesitant to give this a "bug" label as it is rather non-urgent, but technically it's a bug. Documenting in an issue ticket for reference.

PR #2963, opened 2022-08-31
Fix double chaining stack overflow

j0pgrm:
#2853

### DIFF
--- a/modules/_chainResult.js
+++ b/modules/_chainResult.js
@@ -2,5 +2,5 @@ import _ from './underscore.js';
 
 // Helper function to continue chaining intermediate results.
 export default function chainResult(instance, obj) {
-  return instance._chain ? _(obj).chain() : obj;
+  return instance._chain ? _(obj) : obj;
 }


jgonggrijp:
Thank you @j0pgrm for making a contribution! The approach you took is not quite right, as it breaks functionality. I'll help you find a better solution.

I suggest you take the following steps:

- [ ] Inside the root directory of your local clone of the Underscore project, on your own `double-chaining` branch, run `npm install`. This will install all the tools we use during Underscore development.
- [ ] Run `npm run test`. You will see that six tests fail, due to your changes.
- [ ] Run `git revert HEAD`. This will create a new commit that restores the situation before your change.
- [ ] Run `npm run test` again. Now, all the tests pass (although we know there is a bug, which is #2853 and which you are trying to fix).
- [ ] Add a *new* test which exposes the bug. It should do something like `var testValue = _.chain(1).chain()` and then assert that `testValue.value()` is equal to `1`. Since the bug is not solved yet, we expect this test to fail initially. Commit this change.
- [ ] Run `npm run test` again and verify that your new test fails, while all other tests pass.
- [ ] Open `test/index.html` in your browser. This will run the tests in the browser and should give you the same result.
- [ ] Use the developer tools of your browser to set a breakpoint at the start of your new test. If you have not done this before, [here](https://dzone.com/articles/how-to-debug-javascript-in-firefox-in-7-easy-steps) is a tutorial for Firefox and [here](https://developer.chrome.com/docs/devtools/javascript/) is one for Chrome. Step through the code to see what is causing the bug. Do this repeatedly until you fully understand what is going on.
- [ ] Post a comment here to explain what you observed and what you think is causing the bug.

After the above steps, we'll discuss your findings and possible solutions.

PR #2965, opened 2022-09-15
Upgrade testing-related dependencies

jgonggrijp:
I was warned by Sauce Labs that they are deprecating some older versions of their API, which made me realize I probably needed to upgrade `karma-sauce-launcher`. I decided to grasp the opportunity to upgrade more devDependencies. I'll be pushing the commits one by one to check that everything still works after every change.


jgonggrijp:
Browser tests through Sauce Labs intermittently fail, for three reasons:

- Full page reloads in Firefox 11 for unknown reasons (probably a test framework compatibility issue).
- Temporary failure to establish a connection with Sauce Labs.
- The well-known, highly sporadic failure of the throttle/debounce reentrant tests, usually in old Edge or old Firefox.

The first reason is by far the most common, and I'll be documenting it in a new issue. The upgrades otherwise seem unproblematic and are quite necessary, so I'm still going to merge this.

Issue #2971, opened 2022-09-23
underscore@1.13.5 - patch-package not found


vfidelix:
I'm founding problems when trying to compile my project with the 1.13.5 version.

`npm ERR! sh: 1: patch-package: not found`

![image](https://user-images.githubusercontent.com/11541680/192056035-4958e88b-9b2c-44b1-8d92-204896d650aa.png)

dpavancini:
Same here
Exit code: 127
Command: patch-package

joeycozza:
Duplicate of https://github.com/jashkenas/underscore/issues/2967

jgonggrijp:
I just released 1.13.6, hopefully that fixes it. Please close if that is the case.

Issue #2970, opened 2022-09-23
Underscore v1.13.5 install error


ODGodinho:
I'm getting the following error with the new version v1.13.5

> yarn install

Exit code: 127
Command: patch-package
Arguments: 
Directory: /home/dev123/Documentos/projetos/MyProject/node_modules/underscore
Output:
/bin/sh: 1: patch-package: not found
    at ProcessTermError.ExtendableBuiltin (/home/dev123/.nvm/versions/node/v16.17.0/lib/node_modules/yarn/lib/cli.js:721:66)
    at ProcessTermError.MessageError (/home/dev123/.nvm/versions/node/v16.17.0/lib/node_modules/yarn/lib/cli.js:750:123)
    at new ProcessTermError (/home/dev123/.nvm/versions/node/v16.17.0/lib/node_modules/yarn/lib/cli.js:790:113)
    at ChildProcess.<anonymous> (/home/dev123/.nvm/versions/node/v16.17.0/lib/node_modules/yarn/lib/cli.js:25787:17)
    at ChildProcess.emit (node:events:513:28)
    at maybeClose (node:internal/child_process:1093:16)
    at Socket.<anonymous> (node:internal/child_process:451:11)
    at Socket.emit (node:events:513:28)
    at Pipe.<anonymous> (node:net:757:14)
error /home/dev123/Documentos/projetos/MyProject/node_modules/underscore: Command failed.
Exit code: 127
Command: patch-package
Arguments:

iglosiggio:
You can add patch-package as a devDependency in your project (as a hacky workaround, i don't really like the idea of a dependency patching other dependencies).
This was triggered by https://github.com/jashkenas/underscore/commit/08cb1404eb27b80bcadb87a32fcd515f23f7c9a6

mykola-mokhnach:
I can confirm, we started observing the same error after the recent release:
https://github.com/appium/appium-uiautomator2-driver/actions/runs/3115559702/jobs/5052578876

boneskull:
I think that should be `prepare`, not `postinstall`.

boneskull:
cc @jgonggrijp

ODGodinho:
> You can add patch-package as a devDependency in your project (as a hacky workaround, i don't really like the idea of a dependency patching other dependencies). This was triggered by [08cb140](https://github.com/jashkenas/underscore/commit/08cb1404eb27b80bcadb87a32fcd515f23f7c9a6)

The command did not solve
[...]

Issue #2969, opened 2022-09-23
Patch is a dev dependency in 1.13.5. 


markfoo1:
Consumers who use underscore as a dependency rather than as a dev dependency fail in postinstall because patch package is not available.

mjaggard:
This has broken my build :(

joeycozza:
duplicate of https://github.com/jashkenas/underscore/issues/2967

jgonggrijp:
I just released 1.13.6, hopefully that fixes it. Please close if that is the case.

Issue #2968, opened 2022-09-23
Npm install fails due to post install script


Rofer11607:
hello,
I cannot deploy my firebase project because this is a dependency to alot of packages I have installed. The remote build bails with code 127 and says 'sh: patch-package: command not found'

I tried to install the package locally and the same thing happens when I run npm install underscore --force. How can I fix this?

vincent-ogury:
duplicate this: https://github.com/jashkenas/underscore/issues/2967

iglosiggio:
You can add patch-package as a devDependency in your project (as a hacky workaround, i don't really like the idea of a dependency patching other dependencies).

jgonggrijp:
I just released 1.13.6, hopefully that fixes it. Please close if that is the case.

PR #2972, opened 2022-09-23
Run patch-package only on development (test+publish)

iglosiggio:
I'm not sure this is the correct fix (don't have the setup to do js development here) but it looks like a viable direction.

This should help with #2967 #2968 #2969 #2970 #2971

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "test-browser": "npm i karma-phantomjs-launcher && karma start",
     "bundle": "rollup -c && eslint underscore-umd.js && rollup -c rollup.config2.js",
     "bundle-treeshake": "cd test-treeshake && rollup --config",
-    "prepare-tests": "npm run bundle && npm run bundle-treeshake",
+    "prepare-tests": "patch-package && npm run bundle && npm run bundle-treeshake",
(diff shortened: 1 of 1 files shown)


ODGodinho (approved):

sald19 (approved):

jgonggrijp:
Thanks for the patch and sorry for the trouble I have caused. I'm going to hotfix this first by simply throwing the postinstall script out of the package.json; I'm not sure what is the right place to run patch-package yet, but it is not required for end users in any case.

jgonggrijp:
I chose a slightly different, but similar approach: run it before building the annotated sources, rather than before running the tests. That's because the patched package is docco, which we use for the annotated sources. Thanks again for the fast action!

jgonggrijp:
(Oh, and also #2974 was merged first by @jashkenas, while I was doing the same thing locally.)

iglosiggio:
> I chose a slightly different, but similar approach: run it before building the annotated sources, rather than before running the tests. That's because the patched package is docco, which we use for the annotated sources. Thanks again for the fast action!

Thank you for your swift resolution of the issue! <3

PR #2978, opened 2022-11-28
fix(range): step documentation

stevenh:
Fix the documentation for range step which changed behaviour in 1.10 to automatically set step to -1 if start is after stop.


jgonggrijp:
That change was introduced in a595fe10, which was already included in the 1.9.0 release. That said, you are right to correct the documentation. Thanks for contributing!

jgonggrijp (approved):

jgonggrijp:
Now online and will be included in the next release.

PR #2982, opened 2023-08-02
Change Typeof Equality

0o001:
* Change typeof equaltiy (==) with strict equality (===)

Why;
https://www.measurethat.net/Benchmarks/Show/19/0/which-equals-operator-vs-is-faster
https://www.measurethat.net/Benchmarks/Show/26565/0/typeof-eqeq-vs-typeof-eqeqeq

### DIFF
--- a/modules/_createIndexFinder.js
+++ b/modules/_createIndexFinder.js
@@ -6,7 +6,7 @@ import isNaN from './isNaN.js';
 export default function createIndexFinder(dir, predicateFind, sortedIndex) {
   return function(array, item, idx) {
     var i = 0, length = getLength(array);
-    if (typeof idx == 'number') {
+    if (typeof idx === 'number') {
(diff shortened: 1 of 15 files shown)


jgonggrijp (changes requested):
Thank you for going through the trouble to contribute.

I'm not necessarily against this change, but a microbenchmark of the type you referenced is meaningless. It does not represent real world situations and provides no justification for a change to the code, even if you get a big difference (which is not even the case in these benchmarks; the difference is less than the uncertainty).

Please try the [Offen benchmark](https://github.com/offen/underscore-benchmark) with d12221366eacbb63e37cff328bed475f34ebe083 as the **baseline** ref and 6c6bcb13f00a475c0fed513c9998c5dbda30a530 as the **comparison** ref, and report back here with your results. If we see a significant improvement, we will continue the review.

Issue #2985, opened 2023-08-17
"supportsDataView" checks DataView is defined but not if it has a constructor


colingm:
We have noticed a few cases where a site has DataView still defined on the page but for some reason they have decided to modify it such that it is no longer a constructor (maybe by doing something that assigns to `window.DataView`). It would be nice if we could make the `supportsDataView` check more than just whether or not DataView is defined.

![image](https://github.com/jashkenas/underscore/assets/11670266/41b9e5a6-2f36-456f-b381-a1abc7b64bdc)

PR #2986, opened 2023-08-23
Add nativeDataView check to isDataView

colingm:
The is DataView check relies on DataView being a constructor like it is natively. However, if an application overrides DataView from the window this ends up throwing an error when underscore is included. This here will check if DataView is the native implementation before trying to use the constructor on it. If it is not native it will fallback on the same heuristic used on older IE versions for checking if something is a DataView or not.

I wasn't really sure what to do for tests here since we don't seem to be testing any of the supports or native methods. We still have the tests for isDataView and those seem to pass in a standard browser as well as in the case of you overriding the native data view.

This is to resolve https://github.com/jashkenas/underscore/issues/2985


jgonggrijp (commented):
Thanks for taking the extra step to address this yourself!

You can test this as follows:

1. Define new tests for the hypothetical scenario that `DataView` has been overridden, which only run when this is the case.
2. Run the test suite to verify that your new tests don't run.
3. Adjust the guards on the existing tests related to `DataView`, so they don't run if `DataView` has been overridden.
4. Run the test suite to verify that the latter tests still run.
5. Write a new module inside the `test/` directory that actually overrides `DataView`.
6. Add the overriding module to `test/index.html` and the `files` field in `karma.conf.js` and `karma.conf-sauce.js`.
7. Run the tests; verify that your new tests are included and that the tests that should be excluded are skipped. Try `npm run test-node` and `npm run test-browser` as well as opening the `test/index.html` in all browsers you have installed locally.
8. Fiddle with your code as necessary until your new tests pass everywhere.
9. Remove the overriding module from the `test/index.html` again.
10. Adjust the `karma.conf.js` and the `karma.conf-sauce.js` so that the overriding module is included randomly, once every three runs on average.
11. ...?
12. profit!

modules/_setup.js
```diff
@@ -21,7 +21,7 @@ export var push = ArrayProto.push,
 
 // Modern feature detection.
 export var supportsArrayBuffer = typeof ArrayBuffer !== 'undefined',
-    supportsDataView = typeof DataView !== 'undefined';
+    supportsDataView = typeof DataView !== 'undefined' && /\[native code\]/.test(String(DataView));
```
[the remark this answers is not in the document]
colingm: That is a great question and consideration. So here is how far back I have tested based on browser support:
- IE9 & older: no DataView support
- IE10 & IE11: passes
- Safari 8 (couldn't go back to Safari 5.1 where DataView was added): passes
- Chrome 26 (couldn't go back to Chrome 9 where DataView was added): passes
- Edge 13: passes
- Firefox 14 & older: no DataView support
- Firefox 15: passes
- Opera: now uses Chromium engine and passes, didn't test pre-chromium Opera

As far as I am aware this should show it is the consistent way to do a toString for native code, it always looks like this:
```
function FUNCTION_NAME(FUNCTION_PARAMS) { [native code] }
```

I could have used a function type check like you suggested but I was still a little uneasy with this approach. DataView is a fairly common name used by third party tools and I have seen it be overriden a few times. I would be concerned about the case where DataView is overriden where it still has a constructor but it doesn't behave the same way so it doesn't actually know what to do with an ArrayBuffer. That might be a bit overly cautious so I am fine going that route if it makes everyone feel more comfortable with this change though.

colingm:
> Thanks for taking the extra step to address this yourself!
> 
> You can test this as follows:
> 
>     1. Define new tests for the hypothetical scenario that `DataView` has been overridden, which only run when this is the case.
> 
>     2. Run the test suite to verify that your new tests don't run.
> 
>     3. Adjust the guards on the existing tests related to `DataView`, so they don't run if `DataView` has been overridden.
> 
>     4. Run the test suite to verify that the latter tests still run.
> 
>     5. Write a new module inside the `test/` directory that actually overrides `DataView`.
> 
>     6. Add the overriding module to `test/index.html` and the `files` field in `karma.conf.js` and `karma.conf-sauce.js`.
> 
>     7. Run the tests; verify that your new tests are included and that the tests that should be excluded are skipped. Try `npm run test-node` and `npm run test-browser` as well as opening the `test/index.html` in all browsers you have installed locally.
> 
>     8. Fiddle with your code as necessary until your new tests pass everywhere.
> 
>     9. Remove the overriding module from the `test/index.html` again.
> 
>     10. Adjust the `karma.conf.js` and the `karma.conf-sauce.js` so that the overriding module is included randomly, once every three runs on average.
> 
>     11. ...?
> 
>     12. profit!

Awesome, thanks for the testing suggestion here. I had investigated 1-5 but wasn't sure how much value was in those types of tests in comparison to the new module suggestion you had. I will go ahead and look at adding both types of tests.

modules/_setup.js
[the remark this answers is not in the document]
jgonggrijp: You do make a convincing case that `[native code]` should be at least somewhat portable. Have you also tried it with Node or Deno?

Regardless of the approach, I'm afraid a watertight detection is not possible, anyway. `[native code]` is easy to fool with `window.DataView = window.ArrayBuffer`, for example.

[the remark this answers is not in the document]
colingm: Yep I tested them both and they handle it the same. You are correct that it is never going to be a watertight detection, to me it is just what is the most likely to occur. At the very least Sizzle.js uses a variation of the same check quite regularly so I am confident in it being a reasonable enough solution: https://github.com/jquery/sizzle/blob/4194dc464e7381139220e39ceb46c0bdfe12dd59/src/sizzle.js#L142

colingm:
@jgonggrijp I have gone ahead and added a module that will override dataview in the browser that makes the tests fails without my corresponding change to supportsDataView. Let me know what you think, sorry it took a bit longer to get back to this than I intended.

test/objects.js
```diff
@@ -609,8 +609,12 @@
       var view2 = new Uint8Array(shared.buffer, 2, 2);
       assert.notOk(_.isEqual(view1, view2), 'same buffer with different offset is not equal');
 
-      // Some older browsers support typed arrays but not DataView.
-      if (typeof DataView !== 'undefined') {
+      if (typeof NativeDataView !== 'undefined') { // Still test overriden DataView
+        assert.ok(_.isEqual(new NativeDataView(u8.buffer), new NativeDataView(u8b.buffer)), 'Identical NativeDataView are equal');
+        assert.ok(_.isEqual(new NativeDataView(u8.buffer), new NativeDataView(i8.buffer)), 'Identical NativeDataView of different typed arrays are equal');
+        assert.notOk(_.isEqual(new NativeDataView(u8.buffer), new NativeDataView(u16.buffer)), 'Different NativeDataView with different length are not equal');
+        assert.notOk(_.isEqual(new NativeDataView(u8.buffer), new NativeDataView(u16one.buffer)), 'Different NativeDataView with different byte data are not equal');
```
jgonggrijp, line 616: These seem to be copies of the lines that come after the `else`, in which you only replaced `DataView` by `NativeDataView`. It would be elegant if you could avoid the duplicate code. Something like this:

```js
var DataViewCtor = (
  typeof NativeDataView !== 'undefined' ? NativeDataView :
  typeof DataView !== 'undefined' ? DataView :
  undefined
);
if (DataViewCtor) {
  assert.ok(_.isEqual(new DataViewCtor(u8.buffer), ...));
  // etcetera
}
```

test/index.html
```diff
@@ -11,6 +11,7 @@
   <script src="vendor/qunit.js"></script>
   <script src="../underscore-umd.js"></script>
   <script src="qunit-setup.js"></script>
+  <script src="overrides.js"></script>
```
jgonggrijp, line 14: The overrides should go before Underscore itself, to ensure that `supportsDataView` is set correctly.

jgonggrijp (changes requested):
Thanks! Your overall approach looks good to me.

In the `test/index.html`, you placed the overrides after the Underscore UMD bundle instead of before it. If your tests work as intended, then this should cause them to fail in IE10 through Edge 13. Do you have the means to test that?

test/overrides.js
```diff
@@ -0,0 +1,11 @@
+(function() {
+    function overrideDataView() {
```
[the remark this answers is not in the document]
colingm: oh whoops I didn't even think about this, it was just my editor's default 😅 , fixed

test/overrides.js
```diff
@@ -0,0 +1,11 @@
...
+        NativeDataView = DataView;
+        DataView = {};
+    }
+
+    var runOverrides = Math.floor(Math.random() * 3) === 0;
+    if (runOverrides) { // Only override browser functions roughly 1/3rd of the time
+        overrideDataView();
+    }
+})();
```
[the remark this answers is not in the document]
colingm: good catch, fixed

test/objects.js
colingm: I had actually thought about this at one point and then must have gotten distracted and forgotten to clean it up, thanks for the reminder, fixed

jgonggrijp (approved):

test/objects.js
```diff
@@ -609,12 +609,12 @@
       var view2 = new Uint8Array(shared.buffer, 2, 2);
       assert.notOk(_.isEqual(view1, view2), 'same buffer with different offset is not equal');
 
-      // Some older browsers support typed arrays but not DataView.
```
jgonggrijp, line 612: I think this comment is still relevant, please keep it.

test/objects.js
```diff
@@ -609,12 +609,12 @@
...
-      if (typeof DataView !== 'undefined') {
-        assert.ok(_.isEqual(new DataView(u8.buffer), new DataView(u8b.buffer)), 'Identical DataViews are equal');
-        assert.ok(_.isEqual(new DataView(u8.buffer), new DataView(i8.buffer)), 'Identical DataViews of different typed arrays are equal');
-        assert.notOk(_.isEqual(new DataView(u8.buffer), new DataView(u16.buffer)), 'Different DataViews with different length are not equal');
-        assert.notOk(_.isEqual(new DataView(u8.buffer), new DataView(u16one.buffer)), 'Different DataViews with different byte data are not equal');
+      var DataViewImpl = typeof NativeDataView !== 'undefined' ? NativeDataView : typeof DataView !== 'undefined' ? DataView : undefined; // Browser could override DataView
```
jgonggrijp, line 612: Please turn the new comment into a full sentence and put it on a separate line above the code it annotates. That keeps it consistent with the literate style of the library.

test/objects.js
```diff
@@ -930,9 +930,9 @@
         'an ArrayBuffer': buffer,
         'a TypedArray': new Uint8Array(buffer)
       };
-      // Some older browsers support typed arrays but not DataView.
-      if (typeof DataView !== 'undefined') {
-        checkValues['a DataView'] = new DataView(buffer);
+      var DataViewImpl = typeof NativeDataView !== 'undefined' ? NativeDataView : typeof DataView !== 'undefined' ? DataView : undefined; // Browser could override DataView
```
jgonggrijp, line 933: This is a repetition of line 612. You can lift it to the scope of the IIFE wrapper (but first change the comment as I suggested there).

test/overrides.js
```diff
@@ -0,0 +1,11 @@
+(function() {
+  function overrideDataView() {
+    NativeDataView = DataView;
+    DataView = {};
+  }
+
+  var runOverrides = Math.floor(Math.random() * 3) === 0;
+  if (runOverrides) { // Only override browser functions roughly 1/3rd of the time
```
jgonggrijp, line 8: Please turn this into a standalone comment as well.

jgonggrijp (changes requested):
Sorry, a few more details I didn't spot before.

By the way, do you have the means to test the wrong order of the `test/index.html` in IE or Edge 13?

colingm:
Yeah I will make it work to test those with both the wrong order and the right order. Also I will take care of those style things too.

colingm:
@jgonggrijp I was going through and making those changes and testing on IE when I realized I have created an odd combination sometimes where you would normally have the stringTagBug and you can't rely on the string tag to know if an object is a DataView and you also have overriden the native DataView to be something else. I think I am going to try to rework the check, basically instead of it just being "hasStringTagBug" I would like to change it to more of a check of whether or not you need to use an alternate isDataView check. This will result in the renaming of a few things so after my next commit there might be a little bit more to review.

colingm:
@jgonggrijp okay I went ahead and pushed the changes to those test files as well as the rework for the DataView check that isolates a bit more the problem. The biggest thing for me right now is the naming structure. It made sense to not leave it as "hasStringTagBug" to me but I am not sure if you agree or agree with the new name for that function so please review and let me know what you think.

Also I went ahead and tested my old version before this rework and then my new version after the rework against IE10 and the new version actually passed all of the tests. The old version meant it would rely on `tagTester('DataView')` which didn't work in IE10 where as this new version will just use the alternate data view check for IE10 and for overriden DataView

jgonggrijp (changes requested):
I think your name changes are justified, although I'm not entirely happy with them. I have made a suggestion below for the variable name.

In any case, please remove the docs from your pull request, those should be updated only when we cut a new release. (Are you running the `prepublishOnly` script as a commit hook?)

Otherwise, I'm happy with your changes!

colingm:
@jgonggrijp Thank you for that last comment, I was also in a place where I wanted to change the name but naming is hard and I wasn't perfectly happy with what I had. I think I have everything taken care of now (no more docs changes). I think I had the docs changes because I had used build so that I could get the minified changes deployed to a site for testing on older browsers and then accidentally committed them.

jgonggrijp (approved):

jgonggrijp:
Well done and thanks again!

colingm:
no problem thanks for your guidance along the way!

jgonggrijp:
Released in 1.13.7. Sorry for the delay!

colingm:
@jgonggrijp that is alright, thank you for helping get this out there!

PR #2996, opened 2024-07-23
feature/theme-toggle fixes #2909

### LINKED ISSUE #2909: Add dark mode support for homepage
Tried adding dark mode support for the docs on the underscorejs.org homepage.

Here's how it looks before and after:

Screen Shot 2021-02-07 at 15 35 44
Screen Shot 2021-02-07 at 15 24 45

The code is available in a very short gist here:

[doccy-dark.css](https://gist.github.com/gprasanth/aae06b6e437d00f37af3bddc78776f57)

To submit a pull request I am not sure where to add this... in index.html in source root? or in docco.css?

In any case, this can be tweaked as required and added. Also, the logo lost the signature blue color, but can be avoided if an alternative logo image with white text is available.

elkcityhazard:
a crack at adding theme toggle for underscores documentation website


elkcityhazard:
@jgonggrijp I am open to feedback for improvements as well.

static/css/theme-toggle.css
```diff
@@ -0,0 +1,264 @@
...
+          text-decoration: underline;
+      }
+
+      .toc_section li a {
+          text-decoration: none;
+          color: #f4f4f4;
+      }
+
+      img#logo {
+          filter: brightness(0) invert(1);
```
jgonggrijp, line 167: This transform doesn't work. The `brightness(0)` is applied first and turns all opaque regions of the image black (because *zero brightness*). Then, the `invert` turns them all white. This explains why the logo is all white in dark mode.

I personally tend to do something more like this instead:

```suggestion
          filter: invert(1) hue-rotate(180deg) brightness(0.9);
```

You might as well do this for *all* images rather than only the one with the `logo` id.

static/css/theme-toggle.css
```diff
@@ -0,0 +1,264 @@
...
+
+  @media screen and (prefer-color-scheme: light) {
+      img#logo {
+          filter: unset;
+      }
+
+  }
+
+
+  @media screen and (prefers-color-scheme: dark) {
```
jgonggrijp, line 206: In these media queries, you are repeating all the styling that was also in the `body[data-theme]` selectors. It would be nice if you could avoid that repetition.

static/js/themeToggle.js
jgonggrijp, line 1: The code in this file is very messy. Please fix the indentation and use a semicolon after every statement.

*Or* omit the file entirely, as I have suggested in the main comment.

static/css/theme-toggle.css
```diff
@@ -0,0 +1,264 @@
...
+          background: #444;
+          border: 1px solid #777;
+      }
+
+      pre {
+          border-left: 5px solid #aaa;
+      }
+  }
+
+  @media screen and (prefer-color-scheme: light) {
```
jgonggrijp, line 198: typo (though I don't think you need this query)

```suggestion
  @media screen and (prefers-color-scheme: light) {
```

jgonggrijp (changes requested):
Thank you for jumping onto this so eagerly, @elkcityhazard!

I wonder whether a manual toggle is really needed. With the `(prefers-color-scheme: dark)` media query, doesn't everyone automatically get what they want? That being said, I like the way it looks and the fact that it remembers the setting on reload.

Of all the styling being applied in dark mode, only the filter on the logo image seems to be taking effect, and the wrong effect at that. This is what I see:

*Computer set to prefer light, toggle set to light*

Schermafbeelding 2024-07-23 om 23 49 43

*Computer set to prefer light, toggle set to dark*

Schermafbeelding 2024-07-23 om 23 49 58

*Computer set to prefer dark, toggle set to dark*

Schermafbeelding 2024-07-23 om 23 50 15

*Computer set to prefer dark, toggle set to light*

Schermafbeelding 2024-07-23 om 23 50 28

These screenshots were taken with Firefox. With Safari, I see the same thing, except that the toggle isn't visible.

I would suggest saving yourself some complexity. Remove the toggle and just rely on the media queries. If you really want to hold on to the toggle, however, I think it still needs some refinements:

- [ ] Make sure the toggle is set to the actual mode initially. I opened the page while the computer was set to dark mode. The logo image was all white (i.e., "dark", not as intended but as expected from the above screenshots). So the page was assuming "dark", but the toggle was switched to the left side with the sun icon.
- [ ] Keep the toggle in sight when users scroll down, for example with `position: fixed`. There are direct on-page links to each function, so a user landing on the page by following such a link will not see the toggle otherwise.

elkcityhazard:
@jgonggrijp are you loading just the index.html?  I moved the static assets into a static dir and am running locally using python http.server to serve the files.  

loading index.html with firefox isn't working out for me either.  

If it's easier I can refactor out the toggle and just use `prefers-color-scheme:dark`

here is what I am seeing for dark mode on firefox at least:

image

elkcityhazard:
@jgonggrijp removed unnecessary Javascript, removed unnecessary static dir, moved stylesheet into head of html.

awaiting feedback.

index.html
```diff
@@ -259,6 +259,66 @@
         width: 100%;
       }
     }
+
+@media screen and (prefers-color-scheme: dark) {
+      body {
+          background: #333 !important;
```
jgonggrijp, line 265: Your editor has a curious way of setting the indentation level. How does it work?

Regardless of the answer, please indent by steps of 2 spaces, like the rest of the code. I personally prefer deeper indentation like you have done here, but consistency is important.

**Edit:** I now realize what is going on. You were using regular 4-space indentation, but line 263 is missing the two leading spaces that would match it with its closing brace.

jgonggrijp (changes requested):
Much better!

I didn't think of running a http server. It is good that you removed this requirement. @jashkenas designed this project (and related ones from the same era such as Backbone) with a low-tech attitude: people should need as few steps as possible to view the documentation on their own computer, run the tests, try the bleeding-edge version etcetera. I would not make that choice if I were to invent the project myself, but I respect this approach and would like to keep it all working.

Three requests:

- [x] Please preserve the coloring of the logo. I think this should work with the CSS filter that I suggested in my previous review.
- [ ] _(optional)_ As far as I'm concerned, the dark mode background could be darker. But if this offends your sense of beauty, I am willing to let go of this request.
- [ ] Align the indentation with the rest of the code.

index.html
elkcityhazard: I changed my tabstop to 2 to better conform.  

applied the suggest filter (thanks)

and indeed, a little more dark mode to the dark mode

jgonggrijp: Thanks. I will look at it again tomorrow.

jgonggrijp (approved):
Perfect! Thanks a lot!

jgonggrijp:
Released in 1.13.7.

Issue #2992, opened 2024-03-14
Mention _.VERSION in the documentation


clanceyp:
Add an alias, or method, for VERSION const e.g. '_.UNDERSCORE_VERSION' or _.version() ==> "Underscore v.X.XX.X" to allow users to determine if Underscore is being used and if so which version.

jgonggrijp:
@clanceyp It is already accessible as `_.VERSION`. I now realize it isn't mentioned in the documentation, though. Will label and retitle accordingly. Thanks for bringing it up!

clanceyp:
What I want to know is whether it's the Underscore library or a different library using the "_" character alias.

jgonggrijp:
Are you aware of other libraries than Underscore or Lodash that use `_`?

If you only need to distinguish between those two, you can use feature detection. Lodash has `_.partialRight`, Underscore does not. Underscore has `_.restArguments`, Lodash does not (at least not in version 4+). There are many other differences as well.

Also, if the major version number is 1, you are much more likely dealing with Underscore than Lodash, since Lodash is currently at 4.

PR #2998, opened 2024-10-15
Add _.VERSION to the documentation 

Aurimas-stack:
Added `_.VERSION` as per https://github.com/jashkenas/underscore/issues/2992 issue.

In my opinion placing it under `Notes` should be more than enough as `utility` section seems to include more complex methods than this one.

version

### DIFF
--- a/index.html
+++ b/index.html
@@ -2979,6 +2979,15 @@ _.chain([1, 2, 3]).reverse().value();
         </ul>
       </p>
 
+      <p id="relational-operator-note">
+        <b class="header">Underscore version</b>
+        <br>
+        It's possible to get current Underscore version via
+        <tt>_.VERSION</tt> 
+        .
+        <pre>_.VERSION => 1.13.7</pre>
+      </p>
+
       <h2 id="changelog">Change Log</h2>
 
       <p id="1.13.7">
--- a/index.html
+++ b/index.html
@@ -499,6 +499,7 @@
           <li data-name="result">- <a href="#result">result</a></li>
           <li data-name="now">- <a href="#now">now</a></li>
           <li data-name="template">- <a href="#template">template</a></li>
+          <li data-name="version">- <a href="#version">version</a></li>
         </ul>
       </div>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2979,6 +2979,15 @@ <h2 id="notes">Notes</h2>
         </ul>
       </p>
 
+      <p id="relational-operator-note">
```
jgonggrijp, line 2982: It is OK to copy-paste another section and then edit, but you need to give the new section its own unique `id`. :-)

index.html
```diff
@@ -2979,6 +2979,15 @@ <h2 id="notes">Notes</h2>
...
+        <b class="header">Underscore version</b>
```
jgonggrijp, line 2983: For consistency with the other API sections, I suggest naming this after the actual symbol.

```suggestion
        <b class="header">VERSION</b>
```

index.html
```diff
@@ -2979,6 +2979,15 @@ <h2 id="notes">Notes</h2>
...
+        <br>
+        It's possible to get current Underscore version via
```
jgonggrijp, line 2985:
```suggestion
        It is possible to get the current Underscore version via
```

index.html
```diff
@@ -2979,6 +2979,15 @@ <h2 id="notes">Notes</h2>
...
+        <tt>_.VERSION</tt> 
+        .
+        <pre>_.VERSION => 1.13.7</pre>
```
jgonggrijp, line 2988: The characters `<`, `>`, `&`, `"` and `'` need to be escaped even within `<pre>`.

```suggestion
        <pre>_.VERSION =&gt; 1.13.7</pre>
```

jgonggrijp (changes requested):
Thank you for making a contribution. @Aurimas-stack!

Please put this in the Utility section and also add it to the table of contents so that people can find it by typing. The Notes section is for comments about the semantics of functions that apply to more than one function.

I have a few nitpicks as well, see below. Otherthan that, I am content.

Aurimas-stack:
Hey @jgonggrijp ,

I've made the requested changes, please review them, cheers.

jgonggrijp (approved):
Thanks! And congratulations with your first contribution!

jgonggrijp:
Now published on the website. Sorry for the delay.

PR #3003, opened 2025-02-20
Add documentation for the startIndex argument of restArguments

jgonggrijp:
Motivation: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/71151#issuecomment-2641787764

Preview of the added paragraph: https://cdn.statically.io/gh/jgonggrijp/underscore/document-restarguments-startindex/index.html#restArguments-startIndex

If anyone has suggestions for ways to simplify the explanation, I'm interested.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1820,6 +1820,28 @@ var raceResults = _.restArguments(function(gold, silver, bronze, everyoneElse) {
 
 raceResults("Dopey", "Grumpy", "Happy", "Sneezy", "Bashful", "Sleepy", "Doc");
 </pre>
+      <p id="restArguments-startIndex">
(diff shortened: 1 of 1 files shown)


jgonggrijp:
@GammaGames @Yahasana would one of you be willing to give this a review? Mentioning you because you expressed willingness to be mentioned over at Backbone.

colingm:
@jgonggrijp one of the challenges I see with the documentation for this function in general (the original example too, not just your additions), is the lack of showing the output. In various other function docs we see logs to a mock console but in this one we are just shown that the rest get called to "sendConsolations". I think it would go a long way to show with some mock console output a log statement of like "Sorry Sneezy!", "Sorry Bashful!", etc. and then to show that with the start index it is actually "Sorry Happy!" etc.

jgonggrijp:
@colingm great suggestion, thanks!

colingm (approved):
This is great thank you!

PR #3004, opened 2025-04-03
title changed by me

skjha80195:
i changed the title of underscore.js please see that and reflect me back.


jgonggrijp:
Do you realize April Fool's was two days ago?

PR #3013, opened 2026-02-25
Clarify _.template security

jgonggrijp:
Once a year or so (on average), someone reports a vulnerability because they noticed that `_.template` creates a string from user input and passes it to the `Function` constructor.  These reports are well-intended and understandable, but they are nevertheless false reports. Such logic might be considered a vulnerability in another context, but `_.template` is supposed to do this by design.

This pull request adds two pieces of documentation, with the aim to clarify `_.template`'s contract to security researchers. The aim is **not** to discourage reports; the misunderstanding does not happen often enough for this to be necessary.

I would like to ask reviewers three questions:

1. If you imagine being a security researcher (or if you are one), do you think that the additions would help you assess whether you found a real vulnerability in `_.template`? Do you think I could further improve on this?
2. Would you feel discouraged to report a vulnerability after reading the addition to the `SECURITY.md`? If yes, what should I remove or change in order to stop it from being discouraging?
3. Do you think the amount of text is proportional to the aims? What could be removed or expanded on to make it more proportional?

I welcome reviews from anyone. I will mention the following people because I think/hope they might be interested: @colingm @ByamB4 @GammaGames @Yahasana

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -24,6 +24,26 @@ disclose the issue.
 After receiving your email, we will respond as soon as possible and indicate
 what we plan to do.
 
+### A note on `_.template`
+
+[`template`][template] allows the user to inject arbitrary JavaScript
+code in the template string. This is allowed by design. In fact, it is
+the main feature of `template`. Without this feature, templates would
+not be able to have conditional or repeated sections.
+
+Because of this feature, it is the responsibility of the user not to
+pass any untrusted input to `template`. The contract is similar to
+that of the `Function` constructor or even `eval`: this function is so
+powerful that it can be dangerous, so use it with care.
+
+If this does not sound exactly like what you were considering to
+report, or in case of doubt, please do send us a report. Of course, we
+would rather be safe than sorry. You would not be the first to find a
+[vulnerability in `template`][cve-2021-23358].
+
+[template]: https://underscorejs.org/#template
+[cve-2021-23358]: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-23358
+
 ## Disclosure policy
 
 After confirming a vulnerability, we will generally release a security update
--- a/index.html
+++ b/index.html
@@ -2692,6 +2692,12 @@ _.now();
         should be a hash containing any <tt>_.templateSettings</tt> that should be overridden.
       </p>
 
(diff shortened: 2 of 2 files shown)


jgonggrijp:
@GammaGames 

> I think it's a good note, it at least gives an place to point when someone tries to submit an issue.

Thanks for the feedback!

> If people _really_ care about it they can disable templating with `_.templateSettings = null` anyway 🤷

Huh, I don't think that works? Did you mean `_.template = null`?

colingm (commented):
1. Yes I think it is helpful to have this sort of text. I think the problem is just a lot of scans find the use of a function constructor and instantly become reports. Additional context like this helps declare that the function can be a vulnerability in an application if used unsafely is a welcome note. 

2. It is very clear in your note to err on the side of caution and report something if you aren’t sure.

3.  I think the only addition/change I would consider would be to be explicit that when used on untrusted or unsanitized user input, the usage of template can become a XSS vulnerability in your application.


We have often had reports against us simply because static scans of our code found the use of a function constructor. For that reason alone we actually patch our included lib version of underscore to remove it entirely (since we didn’t need to use it anyways). I imagine that there are many reports that happen for similar reasons and it makes me wonder if we should/could investigate a built version of underscore that doesn’t even include template so that others that don’t use it can avoid frivolous reports against their own app. Here is how we patch it out (so that us and our customers are alerted if someone tries to use it):

```
const standards = createReplacePatches([
    {
        include: 'libs/underscore.js',
        patterns: [{
            // underscoreFunctionConstructor
            test: /(\s*).*new\s+Function\([^;]*;/g,
            replace: '$1throw new Error("Function constructor not supported");'
        }]
    }
]);
```

colingm (approved):

jgonggrijp:
Well what do you know. We merged and published this just a bit over a week ago, and I just got this same issue reported to me again. The reporter even acknowledged reading the new note in the security policy.

Apparently, the note does not really help reporters to assess whether they found something that needs reporting. On the bright side, I can rest assured that it does not discourage reports.

colingm:
I guess the hard part is we won't really know how many individuals will read the new note and then not report where they would have before. I guess we will just have to view it in a large timespan and see if the number of similar reports is trending down.

PR #3018, opened 2026-06-11
Fix TypeError in _.max/_.min on nullish collection with numeric iteratee

dkempner:
`_.max(null)` and `_.max(null, _.identity)` both return `-Infinity`, but `_.max(null, 0)` throws:

```js
_.max(null, 0);
// TypeError: Cannot read properties of null (reading '0')
```

The fast-path guard in `max`/`min` reads `typeof obj[0] != 'object'` before the `obj != null` check, so a nullish collection combined with a numeric iteratee dereferences `null` before the null check can short-circuit. This PR reorders the conditions so the null check comes first; `_.max(null, 0)` and `_.min(null, 0)` now fall through to the iteratee branch and return `-Infinity`/`Infinity`, consistent with every other nullish-collection case.

No behavior change for non-nullish collections — the numeric fast path and the falsy-iteratee lookup cases (`_.max([[1], [2, 3]], 0)`) are unaffected, and regression tests for both functions are included. Full test suite passes (208/208).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jgonggrijp:
Thanks for disclosing that you used a genAI tool, I highly appreciate that.

The proposed change looks sane and correct to me, in fact it is currently a mystery to me why the conditions are in the wrong order. I will be diving in the commit history to figure that out.

In the meanwhile, could you tell me a bit more about how you ran into this issue? What does the line in your own code look like, where this exception was being triggered by `min` or `max`?

dkempner:
> Thanks for disclosing that you used a genAI tool, I highly appreciate that.
> 
> 
> 
> The proposed change looks sane and correct to me, in fact it is currently a mystery to me why the conditions are in the wrong order. I will be diving in the commit history to figure that out.
> 
> 
> 
> In the meanwhile, could you tell me a bit more about how you ran into this issue? What does the line in your own code look like, where this exception was being triggered by `min` or `max`?

No production code to share. I was just using underscore as a way to test Fable's code spelunking capability.

Feel free to close if you think this isn't worth the merge.

jgonggrijp:
I will not close it (at least not yet), but I will take more time to review this critically.

modules/max.js
```diff
@@ -7,7 +7,7 @@ import each from './each.js';
 export default function max(obj, iteratee, context) {
   var result = -Infinity, lastComputed = -Infinity,
       value, computed;
-  if (iteratee == null || (typeof iteratee == 'number' && typeof obj[0] != 'object' && obj != null)) {
+  if (iteratee == null || (typeof iteratee == 'number' && obj != null && typeof obj[0] != 'object')) {
```
jgonggrijp, line 10:
```suggestion
  if (obj == null) return result;
  if (iteratee == null || (typeof iteratee == 'number' && typeof obj[0] != 'object')) {
```

modules/min.js
```diff
@@ -7,7 +7,7 @@ import each from './each.js';
 export default function min(obj, iteratee, context) {
   var result = Infinity, lastComputed = Infinity,
       value, computed;
-  if (iteratee == null || (typeof iteratee == 'number' && typeof obj[0] != 'object' && obj != null)) {
+  if (iteratee == null || (typeof iteratee == 'number' && obj != null && typeof obj[0] != 'object')) {
```
jgonggrijp, line 10:
```suggestion
  if (obj == null) return result;
  if (iteratee == null || (typeof iteratee == 'number' && typeof obj[0] != 'object')) {
```

jgonggrijp (changes requested):
Thanks for your patience. I dug deep, and found the following history of the affected line. Before July 2015, the condition was simpler:

```js
    if (iteratee == null && obj != null) {
```

Then, in #2211, Tom Martin proposed to change it to this:

```js
    if (iteratee == null || (typeof iteratee == 'number' && typeof obj[0] != 'object') && obj != null) {
```

He probably actually intended this, but apparently didn't realize that the order of evaluation and the precedence of operators was important:

```js
    if (obj != null && (iteratee == null || (typeof iteratee == 'number' && typeof obj[0] != 'object'))) {
```

However, this mistake was not intercepted by the reviewers JDD and Graeme Yeates. The latter merged the pull request.

As a result, the `iteratee == null` branch no longer included the `obj != null` check. We got away with this because `values(obj)` on the next line converts nullish objects to empty arrays. This is just a slight performance loss rather than a bug.

Later in 2018, Jeremy Ashkenas was going over the code to clean up notational mistakes and removed the superfluous pair of parentheses:

```js
    if (iteratee == null || typeof iteratee == 'number' && typeof obj[0] != 'object' && obj != null) {
```

More recently, I added parentheses around the whole group of conjunctions, because Extendscript erronously gives equal precedence to the `||` and `&&` operators. This finally brought us to the version before your change:

```js
  if (iteratee == null || (typeof iteratee == 'number' && typeof obj[0] != 'object' && obj != null)) {
```

---

Now for what to do with your pull request. In any case, I think the line in question should change, because it is based on historical mistakes.

Realistically, the `typeof iteratee == 'number' && typeof obj[0] != 'object'` part of the condition, that Tom Martin introduced, will only be triggered when `_.min` or `_.max` is operating on a container of arrays. It was also meant as a guard for that scenario. `obj != null` is always `true` in that case. This explains why the logical inconsistency was never noticed before. Hence, I somewhat disagree with Fable that this is a real bug. The `_.max(null, 0)` testcases are unrealistic.

That being said, the `obj != null` check has been there from the beginning, for a sane reason: there is no need to perform any iteration or function calls when we already know that `obj` is nullish. We don't even need to go into the `else` branch in that case.

For this reason, I would like to lift the `obj != null` check to a more prominent position than it has had before. I propose to split the affected line in two, with the first being an early return specifically for `obj == null`:

```js
  if (obj == null) return result;
  if (iteratee == null || (typeof iteratee == 'number' && typeof obj[0] != 'object')) {
```

You are welcome to change this in your PR and take credit for it (or share the credit with me, if you use the "apply suggestion" feature in my comments below). As for the tests, the ones you already introduced can be retained, but I suggest adding more variations on having a nullish `obj` with other arguments being present. Regardless of other arguments, these functions should always return the same extreme value when `obj` is nullish.

jgonggrijp:
@dkempner is this still on your to-do list?

dkempner:
Thanks for the suggestions @jgonggrijp !

jgonggrijp (approved):

jgonggrijp:
Thanks for contributing.

PR #3019, opened 2026-06-21
Fix _.uniq isSorted fast path failing to deduplicate NaN values

JSap0914:
## Problem

The `isSorted` fast path in `_.uniq` uses `seen !== computed` to detect whether consecutive elements differ. Since `NaN !== NaN` always evaluates to `true` per IEEE 754, consecutive NaN values in a sorted array are never recognized as duplicates:

```js
_.uniq([NaN, NaN], true);        // → [NaN, NaN]  (wrong, should be [NaN])
_.uniq([1, NaN, NaN, 2], true);  // → [1, NaN, NaN, 2]  (wrong, should be [1, NaN, 2])
```

The unsorted path (which delegates to `_.contains` / `_.indexOf`) already handles NaN correctly via the NaN branch in `createIndexFinder`. This bug only affects the `isSorted && !iteratee` fast path.

## Fix

Replace the bare inequality check with a SameValueZero-style comparison that treats NaN as equal to itself:

```js
// before
if (!i || seen !== computed) result.push(value);

// after
if (!i || !(seen === computed || (seen !== seen && computed !== computed))) result.push(value);
```

The sub-expression `(seen !== seen && computed !== computed)` is `true` iff both values are NaN, at which point they are treated as equal and no push occurs.

## Tests

Two regression assertions added to the existing `uniq` test in `test/arrays.js`. All 208 tests pass.

> AI-assisted contribution (GitHub Copilot / Claude).


jgonggrijp:
@JSap0914 Thank you for going out of your way to prepare a pull request. Did you actually run into a situation where you needed to deduplicate `NaN` values, or did you just find this "issue" from AI-assisted code analysis?

I'm quoting "issue" because the function already behaves [as documented](https://underscorejs.org/#uniq):

> Produces a duplicate-free version of the **array**, _using === to test object equality_.

(Italics mine.)

Also, `NaN` values are unsortable, so you technically cannot take the sorted path when an array contains `NaN`.

For this reason, I will be rejecting the pull request, although I appreciate your willingness to contribute. You are still welcome to reply after I close the pull request.

Side note: the `else if (iteratee)` path might be `isSorted && iteratee`, in which case `contains` defers to `sortedIndex`, which does not perform a `NaN` check. So in any case, your proposed solution would not completely cover all situations where duplicate `NaN`s might be present.

If you do have a use case for deduplicating a sorted array that contains `NaN` values, I suggest one of the following approaches. Option 1 completely removes all `NaN`s before doing a pass with `uniq`:

```js
var nonNaN = _.reject(yourArray, _.isNaN);
// optionally sort now if desired
var deduplicated = _.uniq(nonNaN);
```

You could of course collapse this into a single expression as `_.uniq(_.reject(yourArray, _.isNaN))`, or use a chain as in `_.chain(yourArray).reject(_.isNaN).uniq().value()`.

If you do actually want to retain a single copy of `NaN` (why?), there are two more options. Option 2 is for when you don't care about the original position of the `NaN` and you just want it to be at the start or the end of the array after deduplicating. It is like option 1, but you check beforehand whether there is a `NaN` in `yourArray` and if that is the case, add it back to your final result.

```js
var hasNaN = _.some(yourArray, _.isNaN);
// continuing as in option 1
var nonNaN = _.reject(yourArray, _.isNaN);
// optionally sort now if desired
var deduplicated = _.uniq(nonNaN);
// end of option 1
// add back NaN if it was present before
if (hasNaN) deduplicated.push(NaN);
// could also add it at the beginning with .unshift
```

Option 3 is for when the position where the first `NaN` appeared is important (but keep in mind that this position is meaningless if it came out of a sorting operation). In this case, you can use the `iteratee` to map `NaN` to a sentinel value.

```js
var NaNsentinel = {};

// the iteratee
function fromNaN(value) {
    return _.isNaN(value) ? NaNsentinel : value;
}

// this time, no _.reject needed, but this is slower
var deduplicated = _.uniq(yourArray, fromNaN);
```

PR #3020, opened 2026-06-21
Update repository metadata URL

123xiaode456-boop:
Summary:
- update the package repository URL from git:// to git+https://
- keep the existing GitHub repository target unchanged

Verification:
- parsed package.json with Node and confirmed repository.url
- ran git diff --check

This is a metadata-only change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "homepage": "https://underscorejs.org",
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


jgonggrijp:
Thank you for the fix and congratulations with your first contribution. If you don't mind, please share how you ran into this issue.

PR #3021, opened 2026-06-22
Fix stack overflow when chaining an already chained wrapper

### LINKED ISSUE #2853: Double chaining leads to a stack overflow

chatman-media:
## Summary

Fixes #2853.

`_.chain(value).chain()` — or any redundant `.chain()` call within a chained sequence — overflows the stack with `RangeError: Maximum call stack size exceeded` instead of behaving like a no-op and returning a wrapper of the same value.

```js
_.chain([1]).chain();             // RangeError: Maximum call stack size exceeded
_([1]).chain().chain().value();   // RangeError: Maximum call stack size exceeded
```

## Root cause

`chainResult` continues chaining intermediate results via `_(obj).chain()`. When `obj` is the value returned by the standalone `chain` function, it is *already* a wrapped `_` instance. `_(obj)` therefore returns that same wrapper (per `if (obj instanceof _) return obj;`), and invoking the prototype `chain` method on it re-enters `chainResult`, which re-chains again — recursing without a base case until the stack overflows.

The recursion loop is: prototype `chain` → standalone `chain` (produces a wrapper) → `chainResult` re-chains that wrapper via prototype `chain` → repeat.

As noted in the issue, this was introduced together with #1691.

## Fix

Call the standalone `chain` helper from `chainResult` instead of `_(obj).chain()`. `chain` sets the `_chain` flag on the (possibly already wrapped) value directly and returns, so chaining an already chained wrapper is idempotent and terminates:

```js
return instance._chain ? chain(obj) : obj;
```

This is safe for every `chainResult` call site: the array-method wrappers pass a plain (unwrapped) value, and `tap` returns the unwrapped value too — only `_.chain` returns an `_` instance, which is exactly the case that previously recursed.

## Tests

Added a regression test in `test/chaining.js` (`#2853`). It verifies the redundant `.chain()` no longer throws, preserves the wrapped value, stays chained, and that further chaining keeps working.

- Without the fix the new test fails with `Maximum call stack size exceeded` (verified).
- With the fix the full suite passes (`209` assertions) and `eslint` is clean.

No public API change; `_.chain([1]).chain()` now returns a chained wrapper of `[1]` as expected.


jgonggrijp (changes requested):
Kudos for the accurate finding and thank you for the high-quality pull request!

I want to accept your changes, but I think we first need to address a nasty subtlety. If someone creates a custom Underscore with a "pimped" variant of `chain`, then your fix will ignore the pimped `chain` and use Underscore's built-in `chain` instead. This could be a breaking change for some users, although we are currently not testing for it.

Hence, the first thing I would like you to adjust, is to add another test to verify that a redundant `chain` call still uses the *same* `chain` even when it is an overridden `chain`. I expect such a test to fail with your currently proposed fix; please verify this before continuing to make other changes.

I imagine such a test to go something like this (you may need/want to tinker a bit with it):

```js
// keep a copy of the original chain
var rawChain = _.chain;

function pimpedChain(obj) {
  var wrapped = rawChain(obj):
  wrapped._pimped = true;
  return wrapped;
}

// override chain for the duration of the test
_.mixin({chain: pimpedChain});

// testing (you may want to add other assertions)
assert.ok(_.chain(1).chain()._pimped === true);

// restore original chain for other tests
_.mixin({chain: rawChain});
```

Once you have such a test in place and you have confirmed that it fails, I *think* you can make it pass by reverting `chainResult` to how it was before and then inlining your fix in `mixin` (this would replace line 14):

```js
      var result = func.apply(_, args);
      if (name === 'chain') return this._chain ? func(result) : result;
      return chainResult(this, result);
```

jgonggrijp:
@chatman-media is this still on your to-do list?

PR #3024, opened 2026-09-13
Fix isEqual symmetry for NaN and its number wrapper

Hugohong258:
`_.isEqual(NaN, new Number(NaN))` returns `false`, while swapping the arguments returns `true`. The same inconsistency occurs inside arrays and objects.

Let comparisons against object values reach the existing class-specific comparison instead of returning early from the primitive NaN check. Add regression assertions for both nested cases and a finite number wrapper that must remain unequal. The non-minified bundles and source maps are regenerated as required by the project's build hooks; minified release files are untouched.

Validation: the new assertions fail against the original implementation after rebuilding its bundles. With the fix, `npm test` passes all 208 QUnit tests, ESLint and the build/tree-shaking checks.


lbesecker195 (approved):
Ran the built `underscore-node.cjs` before and after: `_.isEqual(NaN, new Number(NaN))`, `[NaN]` vs `[new Number(NaN)]`, and `{v: NaN}` vs `{v: new Number(NaN)}` go from false to true, matching the reversed argument order, while NaN against `new Number(79)`, `{}`, `[]`, `new Date(NaN)`, `'NaN'`, `new String('NaN')`, `null` and a function all stay false. `qunit test/objects.js` passes on the branch. LGTM.

jgonggrijp (approved):
I agree this is a consistency fix. Thank you for contributing.

Out of curiosity: did you ask an AI assistant to look for bugs in Underscore, or did you actually run into this inconsistency in your own code?

@lbesecker195 To what extent did you defer your review to an AI assistant?

Hugohong258:
Yes, I asked an AI assistant to look for potential issues in Underscore. This inconsistency came from that search; I didn't encounter it in my own code. I gave the proposed change a quick personal review before opening the PR, but I should have looked at it more carefully. It's striking how quickly these tools have improved. Thanks for reviewing and merging the fix.

PR #3026, opened 2026-09-18
Fix toArray for empty string objects

jakezwang:
`_.toArray(new String(''))` returns `null`, although `_.toArray('')` returns `[]`.

An empty string object is truthy, so it reaches `String#match`, which returns `null` when there are no characters. Restore the empty-array fallback removed in #2349. Add regression coverage for an empty string object and a string object containing astral characters.

Validation: `npm test` passes all 208 tests, including lint and bundle checks. The repository's commit hook regenerated the unminified bundles and source maps.


jgonggrijp:
I'm sorry but this seems very much like a non-issue to me. Did you ask an AI assistant to look for "bugs" in Underscore, or has this actually hit you in a real application? In the latter case, please share how it affected you.

jakezwang:
I have not encountered this in a real application. This is an independently found edge case involving an empty String object, so I cannot provide production impact evidence. I resolved the source-map conflicts against current master and regenerated the unminified bundles; `npm test` passes all 208 tests. That resolves the mechanical conflict but does not establish that the behavior change is worth adopting.

jgonggrijp:
By "independent", do you mean "without AI assistance"?

